PR #23, opened 2023-10-17
Update README.md

bllchmbrs:
Small correction to the html of the readme.


vivi:
Thanks for spotting this!

PR #48, opened 2023-10-20
chore: Add Twitter Preview Card

devXprite:
I added meta tags for [`Open Graph & Twitter`](https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started).

### Before:
sc_b

### After: 
sc_a


vivi (approved):
LGTM -- thanks for doing this!

PR #54, opened 2023-10-20
Added Dockerfile for running main.py from docker

rahulvramesh:
Hi, i have added a basic `Dockerfile` for running the cli from docker. it uses `python:3.10-slim` as base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,12 +4,15 @@ FROM python:3.10-slim
 # Set the working directory in the container
 WORKDIR /usr/src/app
 
-# Copy the current directory contents into the container at the working directory
-COPY . .
-
+# Set default environment variables
 ENV OPENAI_API_KEY=""
 
-# Install any dependencies specified in requirements.txt
+# This ensures that the pip install layer is recreated only when there are changes in the requirements.txt
+COPY requirements.txt ./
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+# Set the working directory in the container
+WORKDIR /usr/src/app
+
+# Copy the current directory contents into the container at the working directory
+COPY . .
+
+ENV OPENAI_API_KEY=""
+
+# Install any dependencies specified in requirements.txt
+RUN pip install --no-cache-dir -r requirements.txt
```
drnic, line 13: Is there a way to move this above `COPY . .` so that this "download many pip packages and install them" image layer is only run if `requirements.txt` is modified?

rahulvramesh: thanks @drnic for the review , updated Dockerfile

vivi:
Thank you for your contribution and the effort you've put into this PR. While Docker is useful for managing complex dependencies, for the current phase of our project, we're focusing on managing dependencies through Python packages to maintain simplicity, and we expect that the people who are running Docker also have python3 installed and can make a virtual environment if they wish to do so.

We hope you understand our perspective -- your enthusiasm and initiative are genuinely appreciated!

PR #73, opened 2023-10-21
added puntuation

bhargavshirin:


vivi:
Thanks for submitting a PR!

Closing this because we have been consistent with not having punctuation in bullet points. Moving the 🚀 around is also not fixing punctuation.

PR #79, opened 2023-10-21
azure support

### LINKED ISSUE #63: Azure OpenAI support
It would be wonderful if you could add the Azure OpenAI as option.

We are on the path of developing the elderly care soft  robots using  Azure AI services like Prompt Flow, and found the MemGPT could evolve the care robot as emotion support level with long term  human  interaction !!!!

Https://hyodol.com

cpacker:
Closes #63 

Does not break existing OpenAI support, but needs testing with an Azure OpenAI account.


vivi (approved):
LGTM. Needs someone w/ Azure credentials to test though

PR #83, opened 2023-10-21
autosave on /exit

cpacker:
Question: should we autoload (run `/load`, loads most recent checkpoint) by default?

if so, we need a flag to disable

if not, user will need to use `/load`

Also, should we catch ctrl-c/d and pass into /exit to get the autosave?

---

edit: upon further thought, maybe we should autoload, but only once we move to questionary?

for example, when you start main.py, you select whether or not you want to load from your last saved session

### DIFF
--- a/main.py
+++ b/main.py
@@ -43,6 +43,62 @@ def clear_line():
         sys.stdout.flush()
 
 
+def save(memgpt_agent):
+    filename = utils.get_local_time().replace(' ', '_').replace(':', '_')
+    filename = f"{filename}.json"
+    filename = os.path.join('saved_state', filename)
+    try:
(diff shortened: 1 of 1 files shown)


vivi (approved):
LGTM.

vivi:
> Question: should we autoload (run `/load`, loads most recent checkpoint) by default?
> 
> if so, we need a flag to disable
> 
> if not, user will need to use `/load`
> 
> Also, should we catch ctrl-c/d and pass into /exit to get the autosave?
> 
> edit: upon further thought, maybe we should autoload, but only once we move to questionary?
> 
> for example, when you start main.py, you select whether or not you want to load from your last saved session

Yeah, we should autoload once we move to questionary.

Issue #44, opened 2023-10-19
Cost estimates for gpt4 api calls


vivi:

cpacker:
Taking off roadmap for now to clear the issue stack for now, can reopen later if we reprioritize

Issue #18, opened 2023-10-17
[Feature Request] Support for local LLMs like Ollama


mmmeff:
title

tjweir:
The ability to use local LLMs would be great.

cpacker:
added to the roadmap!

Aaronminer1:
using it withe LM studios local api server would be great spent half the day trying to get it to connect but alas no good results it should be possible since the server is suppose to be a drop in replacement for openai api https://lmstudio.ai/

molander:
LM Studio is interesting but in keeping with the spirit of open-source, a better solution would be https://github.com/go-skynet/LocalAI, a fully open drop-in OpenAI API replacement that includes support for functions. I am cloning memGPT now and have a localAI installation so perhaps I can see this weekend what would be required.

giuliogatto:
Support for local LLMs would be a game changer, in particular being able to use Mistral 7B

jackfood:
> using it withe LM studios local api server would be great spent half the day trying to get it to connect but alas no good results it should be possible since the server is suppose to be a drop in replacement for openai api https://lmstudio.ai/

Any luck on running with LM Stuido?

cronobjs:
> > using it withe LM studios local api server would be great spent half the day trying to get it to connect but alas no good results it should be possible since the server is suppose to be a drop in replacement for openai api https://lmstudio.ai/
> 
> Any luck on running with LM Stuido?

I have LM Studio And Im trying to figure this out but its so confusing. If anyone out there has able to get any Llama versions to run with MemGPT that would be Helpful.

dataswifty:
> added to the roadmap!

Thank you!

cpacker:
We are actively working on this (allowing pointing MemGPT at your own hosted LLM backend that supports function calling), more updates to come soon.

d0rc:
sorry, for asking obvious questions - isn't it possible to just start local OpenAI API, following llama.cpp-python bindings documentation, or bringing it up with LMStudio and override OPENAI_API_ENDPOINT environment variable or something like that?

nbollman:
> We are actively working on this (allowing pointing MemGPT at your own hosted LLM backend that supports function calling), more updates to come soon.

[...]

Issue #12, opened 2023-10-17
New LLM Integration


aniketmoha9:
Have you integrated this with other llm api provider like amazon bedrock or watsonx? If you have can refr me some documentation.

ishaan-jaff:
made a PR for this: https://github.com/cpacker/MemGPT/pull/86

PR #86, opened 2023-10-22
Add support for Bedrock, Ollama, Palm, Claude-2, Cohere, Replicate Llama2, CodeLlama, Hugging Face (100+LLMs) - using LiteLLM

ishaan-jaff:
This PR adds support for the above mentioned LLMs using LiteLLM https://github.com/BerriAI/litellm/ 
LiteLLM is a lightweight package to simplify LLM API calls - use any llm as a drop in replacement for gpt-3.5-turbo. 

Example
```python
from litellm import completion

## set ENV variables
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion(model="command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/memgpt/openai_tools.py
+++ b/memgpt/openai_tools.py
@@ -4,6 +4,7 @@ import os
 import time
 
 import openai
+import litellm
 
 
 def retry_with_exponential_backoff(
@@ -52,7 +53,7 @@ def retry_with_exponential_backoff(
 
 @retry_with_exponential_backoff
 def completions_with_backoff(**kwargs):
-    return openai.ChatCompletion.create(**kwargs)
+    return litellm.completion(**kwargs)
 
 
 def aretry_with_exponential_backoff(
@@ -105,7 +106,7 @@ async def acompletions_with_backoff(**kwargs):
     azure_openai_deployment = os.getenv('AZURE_OPENAI_DEPLOYMENT')
     if azure_openai_deployment is not None:
         kwargs['deployment_id'] = azure_openai_deployment
-    return await openai.ChatCompletion.acreate(**kwargs)
+    return await litellm.acompletion(**kwargs)
 
 
 @aretry_with_exponential_backoff
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ rich
 tiktoken
 timezonefinder
 tqdm
+litellm


ishaan-jaff:
addressing: 
https://github.com/cpacker/MemGPT/issues/18 
https://github.com/cpacker/MemGPT/issues/12 
https://github.com/cpacker/MemGPT/issues/44(we also allow you to track cost for llm requests)

ishaan-jaff:
@cpacker can you take a look at this PR when possible ?

angar4ik:
when I was trying to work this with LM Studio, it was complaining about missing function role

theCuratorCM:
Voting this as one of the highest leverage PRs of all time.

vivi:
Thank you for taking the time to propose the integration of the `litellm` package into our project. We recognize the effort and dedication that has gone into developing `litellm`, and we genuinely appreciate your interest in contributing to our platform.

We've decided that we won't be integrating `litellm` directly into our codebase for the following reasons:
1. Applicability to MemGPT: We are committed to preserving the flexibility of our current codebase and ensuring that our users are not locked into a specific solution. It is also misleading to users currently because the “100+LLMs” support does not directly apply to MemGPT. Almost all of these models—including the last two in your example—do not support function calling for the purposes of MemGPT.
2. Complexity: While `litellm` offers various features, we believe that our specific needs can be addressed in a more streamlined manner without the need for adding another package dependency. See https://github.com/cpacker/MemGPT/issues/18#issuecomment-1773912818. We are also rolling our own support for function calling on local LLMs (#97).

We hope you understand our decision and wish you all the best with the continued development of `litellm`.

krrishdholakia:
hey @vivi , thanks for the feedback: 

- Re: function calling for unsupported models via litellm - For unsupported models, you can choose to add it to the prompt as part of the system message via `litellm.add_function_to_prompt = True`. 

- Re: streamlined manner - we work **extremely hard** to make sure litellm is as light as possible. Is there anything specific bloat we can help address here?

cpacker:
We believe our users are better off using OpenAI's officially supported `openai` python API with our function calling parsers if they want to use non-OpenAI LLMs (which have to be written regardless to get good MemGPT performance with non-OpenAI models). This is not even touching on the privacy implications of unnecessarily routing every MemGPT user's personal traffic through a startup's proxy servers. If they really want to use litellm for some reason, they can install it on their own and swap the imports.

Please stop spamming github repositories with pull requests, thank you.

image

krrishdholakia:
wait @cpacker litellm isn't a proxy server. we let users spin up an openai-compatible server if they'd like. 

It's just a python package for translating llm api calls. I agree with you, routing things through a proxy would be a bit weird.

Issue #11, opened 2023-10-17
Improve the CLI


vivi:
- [x] probably move to questionary
- [x] ask if want to update embeddings / `--refresh-embeddings` flag
- [x] support #8 
- [x] currently, CLI only supports a limited amount of characters for input. 

~- [ ] questionary-style CLI UI for (1) loading+saving (2) viewing bot memory~

drnic:
I started having a conversation about some code. But when I pasted a chunk of code into the CLI, the cli only took the first line -- which led to a nonsensical conversation. Perhaps could we support toggling between single line + multiline inputs (like chatgpt UI)

vivi:
@drnic Did you try typing `//` to enter multi-line mode?

drnic:
@vivi I hadn't and that's embarrassing :) I've now found https://github.com/cpacker/MemGPT#interactive-cli-commands section of bonus commands :)

PR #96, opened 2023-10-23
CLI overhaul

vivi:
Addresses #11.

* runs black formatter on `main.py` (should be another PR 🤫)
* uses questionary in lieu of flags (still supports flags as a legacy)
* introduces configuration files that are saved to disk and loaded in
* asks user if they want to autoload saved agent file (if exists)
* asks user if they want to regenerate embeddings
* no longer relies on readline

cli-overhaul

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,5 +23,6 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
 
-    - name: Run main.py
-      run: python main.py
+    - name: Run main.py with input
+      run: |
+        echo -e "\n\n\nn" | python main.py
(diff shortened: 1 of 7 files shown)


cpacker:
Comments:

- [ ] Can we reorganize the personas in the selection list? Maybe put sam_pov first, then memgpt_starter, etc (basically sam_pov first, and memgpt_doc not first)
- [ ] At the user selection, should be "Which user" not "Which persona"
- [ ] Can we put a note that the user can create their own personas and users by adding text files inside the selection box?
- [ ] Can we put "Warning - you are running MemGPT with X" above "Running ..."? So two pink things are together

cpacker:
Also should we sauce up the [exit by typing /exit] message? It's so sad and grey next to its multicolor friends.

cpacker (changes requested):
see comment

cpacker (approved):

PR #104, opened 2023-10-23
Create python package with poetry 

sarahwooders:
Building from source: 

1. Install poetry 
2. Run `poetry install` 
3. Run `poetry run memgpt` (for option: `poetry run memgpt --help`)

To create a package, run `poetry build`. This will create a file `dist/memgpt-0.1.0-py3-none-any.whl` which is installable via pip (e.g. `pip install pip install <path>/dist/memgpt-0.1.0-py3-none-any.wh`). Once the package is installed, the CLI can be run with `memgpt`. 

PyPI package: https://pypi.org/project/pymemgpt/


vivi:
- [ ] Fix directory where users can add their own personas
- [ ] Update README

vivi:
@cpacker ready to merge pending testing on windows

README.md
```diff
@@ -89,12 +94,33 @@ export OPENAI_API_KEY=YOUR_API_KEY
...
+Clone this repo: `git clone `
+
+Using poetry:
+1. Install poetry: `pip install poetry`
+2. Run `poetry install`
+3. Run `poetry run memgpt`
+
+Using pip:
+1. Run `pip install -r requirements.txt`
+2. Run `python3 main.py`
```
sarahwooders, line 120: We should add that `pip install -e .` instead of this?

README.md
```diff
@@ -89,12 +94,33 @@ export OPENAI_API_KEY=YOUR_API_KEY
...
 
-To run MemGPT for as a conversation agent in CLI mode, simply run `main.py`:
+To run MemGPT for as a conversation agent in CLI mode, simply run `memgpt`:
 
 ```sh
-python3 main.py
+memgpt
 ```
 
+If you get `command not found` (Linux/MacOS), or a `CommandNotFoundException` (Windows), the directory where pip installs scripts is not in your PATH. You can either add that directory to your path (`pip show pip | grep Scripts`) or instead just run:
```
sarahwooders, line 103: I'm not sure if we should add this since this is just a pip setup issue right, not specific to MemGPT?

vivi: Feel free to change the wording (or put it in a details field), but I think this will be common enough of an error to have in the README

sarahwooders: If this is specific to the package, then we should fix it

vivi: Sure -- I removed the build from source rec and put this paragraph in a details field for "Debugging".

cpacker (approved):

cpacker:
LGTM 😈

PR #106, opened 2023-10-24
Added db load ability

mr-sk:
* Added ability to load .db (as shown in original README)
    * This didn't work because it was being read in as the csv file, not the db
* Added README instructions to load .csv (not shown in original README)

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,11 @@ To talk to this database, run:
 python main.py  --archival_storage_sqldb=memgpt/personas/examples/sqldb/test.db
 ```
 
+Or to load a csv directly:
+```sh
+python main.py  --archival_storage_sqldb=memgpt/personas/examples/sqldb/test.csv
+```
+
 And then you can input the path to your database, and your query.
 
 ```python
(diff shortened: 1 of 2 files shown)


mr-sk:
@vivi Remerged against latest (the spacing and double quotes). Thanks.

vivi:
Hi @mr-sk, thank you for catching this!

It seems like with your changes, we no longer need https://github.com/cpacker/MemGPT/blob/main/memgpt/main.py#L358-L369 anymore. Could you change your PR to reflect that + test?

mr-sk:
Hello @vivi - So after further reading the code, realized that the main section for adding the db into archival memory is correct, and the issue is that in utils the final else (when not explictily handling the .db extension) was attempting to read the sqlite db "raw", causing it to fail on encoding issues. Now, we just return for that file extension, and main handles loading it correctly. 

So this is actually less code and fixes the issue very simply. 

Tested on the local sql .db provided, works well. 
Thanks!

cpacker:
@vivi can this be closed?

mr-sk:
If you close w/out merging you'll get an error when trying to import a db.

PR #126, opened 2023-10-26
Integration with AutoGen workflow

QZGao:
Construct AutoGen config workflow with `create_memgpt_autogen_agent_from_config`.

If `human_input_mode == "ALWAYS"`, it's basically a wrapped `create_autogen_memgpt_agent`;

If `human_input_mode == "TERMINATE"` or `"NEVER"`, it's an auto-running group chat of memgpt agent and user proxy agent.

### DIFF
--- a/memgpt/autogen/examples/agent_autoreply.py
+++ b/memgpt/autogen/examples/agent_autoreply.py
@@ -23,7 +23,7 @@ config_list = [
 # If USE_MEMGPT is False, then this example will be the same as the official AutoGen repo
 # (https://github.com/microsoft/autogen/blob/main/notebook/agentchat_groupchat.ipynb)
 # If USE_MEMGPT is True, then we swap out the "coder" agent with a MemGPT agent
-USE_MEMGPT = False
+USE_MEMGPT = True
 
 llm_config = {"config_list": config_list, "seed": 42}
 
--- a/memgpt/autogen/examples/agent_autoreply.py
+++ b/memgpt/autogen/examples/agent_autoreply.py
@@ -1,0 +1,66 @@
+"""Example of how to add MemGPT into an AutoGen groupchat
+
+Based on the official AutoGen example here: https://github.com/microsoft/autogen/blob/main/notebook/agentchat_groupchat.ipynb
+
+Begin by doing:
+  pip install "pyautogen[teachable]"
+  pip install pymemgpt
+  or
+  pip install -e . (inside the MemGPT home directory)
+"""
+
+import os
+import autogen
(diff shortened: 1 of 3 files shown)


cpacker:
@QZGao could you add a small example (can be a small diff off of `examples/agent_groupchat.py`) that demonstrates how the new functionality works?

I tested the existing script and nothing broke, so once you add an example + paste the outputs in a comment here, I think this is ready to merge.

cpacker (changes requested):
Example on how to use new feature + example output

QZGao:
The original `agent_groupchat.py` will generate the same result with `USE_AUTOGEN_WORKFLOW = True`.

The new `agent_autoreply.py` will generate the following result:
```txt
User_proxy (to MemGPT_coder):

I want to design an app to make me one million dollars in one month. Tell me all the details, then try out every steps.

--------------------------------------------------------------------------------
MemGPT_coder (to MemGPT_coder):

Your goal is ambitious, but not impossible. I'd suggest three phases: 1) Research and Planning, 2) App Design and Development, and 3) Marketing & Monetization. Let's dig deep into these, starting with Research and Planning.

--------------------------------------------------------------------------------
MemGPT_coder (to MemGPT_coder):

In the Research and Planning phase, we define the app's purpose and make sure it offers value. This involves market research, understanding the target audience, and analyzing competitors. Then, choosing the right platform (iOS or Android, or both) and determining the app's features and functionality. It's also a good time to decide on the technology stack for development. Finally, we set a timeline and budget for the project. Now let's elaborate on these points.

--------------------------------------------------------------------------------
MemGPT_coder (to MemGPT_coder):

Step 1: Market Research - Explore app stores, study user reviews, identify successful apps and analyze their business models, target audience, and technology. Step 2: Identify the Target Audience - Understand their preferences, demographics, and mobile behavior. Step 3: Competitor Analysis - Evaluate similar apps, analyze their features, and identify what you can offer differently. Now, let's talk about selecting the right platform.

--------------------------------------------------------------------------------
MemGPT_coder (to MemGPT_coder):

Choosing a platform: If the audience is largely in the U.S or earns a high income, iOS might be better. Globally, Android dominates. Features and functionality: In the first version, only include necessary features. As the app grows, you can add more based on feedback. Tech stack: Depending on complexity, you can choose native development (Swift for iOS, Java for Android), hybrid development (React Native, Flutter), or progressive web apps. Now let's move onto the App Design and Development phase.

--------------------------------------------------------------------------------
```

cpacker (approved):
LGTM - @vivi , should we add a README in the examples folder?

PR #136, opened 2023-10-26
Azure OpenAI Embedding Support

rivms:
- Added a separate environment variable, AZURE_OPENAI_EMBEDDING_DEPLOYMENT, to specify the deployment to be used for embeddings
- Updated sample in README to include the new environment variable
- Moved code block for Azure OpenAI support (setting openai variables) so that legacy flags will use Azure OpenAI settings


vivi:
@rivms can you take a look at #140 as there seems to be some overlap. I'll close this and add you as a co-author there.

Issue #91, opened 2023-10-22
Azure OpenAI - Functions argument error


avery4man:
The variables for Azure OpenAI have been set.
I am now getting the below error:

💭 Bootup sequence complete. Persona activated. Testing messaging functionality.
Hit enter to begin (will request first MemGPT message)
[A[K
Traceback (most recent call last):
  File "C:\Users\user\Documents\Projects\MemGPT\main.py", line 339, in <module>
    app.run(run)
  File "C:\Users\user\AppData\Local\miniconda3\envs\memgpt\lib\site-packages\absl\app.py", line 308, in run
    _run_main(main, args)
  File "C:\Users\user\AppData\Local\miniconda3\envs\memgpt\lib\site-packages\absl\app.py", line 254, in _run_main
    sys.exit(main(argv))
  File "C:\Users\user\Documents\Projects\MemGPT\main.py", line 337, in run
    loop.run_until_complete(main())
  File "C:\Users\user\AppData\Local\miniconda3\envs\memgpt\lib\asyncio\base_events.py", line 649, in run_until_complete
    return future.result()
  File "C:\Users\user\Documents\Projects\MemGPT\main.py", line 315, in main
    new_messages, heartbeat_request, function_failed, token_warning = await memgpt_agent.step(user_message, first_message=False, skip_verify=FLAGS.no_verify)
  File "C:\Users\user\Documents\Projects\MemGPT\memgpt\agent.py", line 545, in step
    raise e
  File "C:\Users\user\Documents\Projects\MemGPT\memgpt\agent.py", line 483, in step
    response = await get_ai_reply_async(model=self.model, message_sequence=input_message_sequence, functions=self.functions)
  File "C:\Users\user\Documents\Projects\MemGPT\memgpt\agent.py", line 115, in get_ai_reply_async
    raise e
  File "C:\Users\user\Documents\Projects\MemGPT\memgpt\agent.py", line 96, in get_ai_reply_async
    response = await acreate(
  File "C:\Users\user\Documents\Projects\MemGPT\memgpt\openai_tools.py", line 98, in wrapper
    raise e
  File "C:\Users\user\Documents\Projects\MemGPT\memgpt\openai_tools.py", line 76, in wrapper
    return await func(*args, **kwargs)
  File "C:\Users\user\Documents\Projects\MemGPT\memgpt\openai_tools.py", line 108, in acompletions_with_backoff
    return await openai.ChatCompletion.acreate(**kwargs)
[...]

Issue #90, opened 2023-10-22
Azure Endpoint not working


nubgamerz:
`openai.error.InvalidRequestError: Must provide an 'engine' or 'deployment_id' parameter to create a <class 'openai.api_resources.chat_completion.ChatCompletion'>`

It seems some of azure's required parameters are missing still.

Here is the error in full:

`File "G:\Code\memgpt\MemGPT\main.py", line 336, in <module>
    app.run(run)
  File "G:\Code\memgpt\MemGPT\.venv\Lib\site-packages\absl\app.py", line 308, in run
    _run_main(main, args)
  File "G:\Code\memgpt\MemGPT\.venv\Lib\site-packages\absl\app.py", line 254, in _run_main
    sys.exit(main(argv))
             ^^^^^^^^^^
  File "G:\Code\memgpt\MemGPT\main.py", line 334, in run
    loop.run_until_complete(main())
  File "C:\Python311\Lib\asyncio\base_events.py", line 650, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "G:\Code\memgpt\MemGPT\main.py", line 312, in main
    new_messages, heartbeat_request, function_failed, token_warning = await memgpt_agent.step(user_message, first_message=False, skip_verify=FLAGS.no_verify)
                                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "G:\Code\memgpt\MemGPT\memgpt\agent.py", line 545, in step
    raise e
  File "G:\Code\memgpt\MemGPT\memgpt\agent.py", line 483, in step
    response = await get_ai_reply_async(model=self.model, message_sequence=input_message_sequence, functions=self.functions)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "G:\Code\memgpt\MemGPT\memgpt\agent.py", line 115, in get_ai_reply_async
    raise e
  File "G:\Code\memgpt\MemGPT\memgpt\agent.py", line 96, in get_ai_reply_async
    response = await acreate(
               ^^^^^^^^^^^^^^
[...]

PR #140, opened 2023-10-26
Patch azure support

### LINKED ISSUE #118: Azure embedding not possible
When using Azure OpenAI it still tries to embbed using the regular OpenAI API. I might be blind, but I can't find a way to configure memgpt to use an Azure embedding model. 

Capture

vivi:
Closes #118 #91 #90

- better validation of azure environment variables, added embeddings deployment variable too
- have to use `engine` if deployment is not passed in
- tested with computing embeddings, also deployment ids are optional


memgpt/main.py
```diff
@@ -187,6 +193,18 @@ async def main(
...
 
+    # Azure OpenAI support
+    if use_azure_openai:
+        configure_azure_support()
+        check_azure_embeddings()
+    else:
+        azure_vars = get_set_azure_env_vars()
+        if len(azure_vars) > 0:
+            print(
+                f"Error: Environment variables {', '.join([x[0] for x in azure_vars])} should not be set if --use_azure_openai is False"
```
cpacker, line 204: 😈 love it 😈

cpacker (approved):
😈 LGTM

memgpt/main.py
vivi: 😈

PR #145, opened 2023-10-26
powershell instructions

cpacker:
From discord:

> Update the readme on the main repo for MS Powershell (had some trouble with environment variable)
> Powershell has a different way to set the environment variables it's like $Env:OPENAI_API_KEY = 'your_api_key'
> Space is also important. Maybe add a second line for windows powershell.... Seems windows try to force users to use powershell more often. Nothing major but I needed to look it up


vivi (approved):
Mr. chatgpt says its valid powershell, in chat we trust.

PR #149, opened 2023-10-27
Retry acompletions_with_backoff for local LLMs

roguestar:
Retry the acompletions_with_backoff method when the local LLM raises an exception. For example get_chat_completion often fails to produce a valid JSON response. Since models aren't very deterministic for a random seed, even a suboptimal model/configuration will typically generate a passable response within a few retries. Otherwise, we re-raise the error. This change results in a more user-friendly experience using local LLMs and provides an opportunity to get more debugging info while the application is running

### DIFF
--- a/memgpt/openai_tools.py
+++ b/memgpt/openai_tools.py
@@ -108,13 +108,13 @@ def aretry_with_exponential_backoff(
 async def acompletions_with_backoff(**kwargs):
     # Local model
     if HOST_TYPE is not None:
-        retry_count = 5 # Retry up to five times
-        for i in range(1, retry_count):
+        retry_count = 5  # Retry up to five times
+        for i in range(1, retry_count + 1):
             try:
                 return await get_chat_completion(**kwargs)
             except Exception as e:
                 print(f"Failed to get a valid response on attempt: {i}\n{str(e)}")
-                if i <= retry_count:
+                if i < retry_count:
                     pass
                 else:
                     raise e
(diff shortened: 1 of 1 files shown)


cpacker (commented):
This seems quite reasonable to me. @vivi thoughts? Otherwise LGTM, will test then merge.

oderwat:
I like that too. I had something similar in my code already but was unsure if doing this will make the problems go away. I used saved states to find the max_new_tokens fix and until that, it was basically always the same exception. But it probably helps with other types of malformed answers.

vivi (approved):
LGTM.

memgpt/openai_tools.py
```diff
@@ -108,7 +108,16 @@ async def wrapper(*args, **kwargs):
 async def acompletions_with_backoff(**kwargs):
     # Local model
     if HOST_TYPE is not None:
-        return await get_chat_completion(**kwargs)
+        retry_count = 5  # Retry up to five times
```
cpacker, line 111: @vivi can you take another look at this rq?

Seems like this may not be redundant with the retry loop we have in main, but I'm fine merging this too (though unclear if it will cause more confusion?)

cpacker:
Closing because in the current repo, as long as the local LLM backend returns a 429 status (same as OpenAI for rate limit error), the retry loop should engage: https://github.com/cpacker/MemGPT/blob/main/memgpt/openai_tools.py#L269-L378

PR #154, opened 2023-10-27
Refactoring CLI to use config file, connect to Llama Index data sources, and allow for multiple agents 

sarahwooders:
Builds on top of #146 

Example running on hugging face dataset (loaded with Llama Index): 

image

# Updated CLI 

## Starting Agent 
```
> memgpt run [--agent <agent-name>] 
```

## Configuring defaults 
```
> memgpt configure 

? Do you want to enable MemGPT with Open AI? Yes
? Use OpenAI as default provider? Yes
? Do you want to enable MemGPT with Azure? No
? Select default model (recommended: gpt-4): gpt-4
['/Users/sarahwooders/repos/MemGPT/memgpt']
['sam_pov', 'memgpt_starter', 'memgpt_doc', 'sam_simple_pov_gpt35', 'sam', 'memgpt_doc', 'basic']
? Select default persona: sam_pov
['cs_phd', 'basic', 'basic']
? Select default human: cs_phd
Saving config to /Users/sarahwooders/.memgpt/config
```
## Loading External Data
External data connectors via Llama Index connectors implemented in `cli/cli_load.py`: 
```
memgpt load [directory/database] [ARGS]
```
This data source can now be attached to an agent by running `memgpt run --data-source <source-name>`.

## Listing Options (Agents, Humans, Personas, Data Sources) 

```
> memgpt list [agents/humans/personas/sources]

> memgpt list humans

+------------+---------------------------------------------------------------------------------------------+
|    Name    |                                             Text                                            |
+------------+---------------------------------------------------------------------------------------------+
| cs_phd.txt | This is what I know so far about the user, I should expand this as I learn more about them. |
|            |                                                                                             |
|            |                                       First name: Chad                                      |
|            |                                         Last name: ?                                        |
|            |                                         Gender: Male                                        |
|            |                                            Age: ?                                           |
|            |                                        Nationality: ?                                       |
|            |                   Occupation: Computer science PhD student at UC Berkeley                   |
|            |      Interests: Formula 1, Sailing, Taste of the Himalayas Restaurant in Berkeley, CSGO     |
| basic.txt  |                                       First name: Chad                                      |
|   basic    |                                       First name: Chad                                      |
+------------+---------------------------------------------------------------------------------------------+
```

## Adding Personas/Humans 
```
memgpt add [persona/human] --name <name> [-f <filepath>] [--text <text>]
```

## Directory Structure 
Configuration, credentials, agent state, and persona/human options are all saved in `~/.memgpt`. 
```
~/.memgpt
[...]

PR #157, opened 2023-10-28
Refactor autogen agent to use sync memgpt, add notebook example

vivi:


cpacker (approved):
LGTM

cpacker:
Need to check #164

cpacker:
@vivi can this be merged in? doesn't look like there are any blockers?

memgpt/presets.py
```diff
@@ -41,5 +42,36 @@ def use_preset(preset_name, model, persona, human, interface, persistence_manage
             first_message_verify_mono=True if "gpt-4" in model else False,
         )
 
+    if preset_name == "memgpt_chat_sync":
```
sarahwooders, line 45: Maybe remove duplicated like 46-63?

memgpt/presets.py
```diff
@@ -1,10 +1,11 @@
 from .prompts import gpt_functions
 from .prompts import gpt_system
-from .agent import AgentAsync
+from .agent import Agent, AgentAsync
 from .utils import printd
 
 
 DEFAULT = "memgpt_chat"
+SYNC_CHAT = "memgpt_chat_sync"
```
sarahwooders, line 8: If I'm understanding what "present" means, doesn't it make more sense to just pass in an `async=True/False` into `use_preset`? Since async/sync is orthogonal to what the OpenAI prompts are.

vivi: This will be removed once #154 is merged in. For now, this exists so that this change will work for autogen users and not have any merge conflicts w/#154

(Once #154 is merged in, we'll change everything to point to SyncAgent)

PR #166, opened 2023-10-28
add missing pip requirement `llama_index`

almontasser:


vivi:
Thanks for catching this.

Issue #26, opened 2023-10-18
Allow MemGPTs to read/write from a file(s)


cpacker:
Enable workflow where MemGPT + user can co-edit a document together.

Add `write_to_text_file` function and `read_to_text_file` function.

Yash-Ambekar:
@cpacker Please elaborate the feature request with some examples. I am interested to implement this feature.

master-harvey:
Allowing the Aider CLI tool to work with memGPTs is probably the way to go here
https://github.com/cpacker/MemGPT/issues/62

genericrich:
Use Cases:
* Document generation - "Create an outline for a business presentation, etc."
* HTML interface presentation - "Create a UI for navigating a MemGPT managed set of documents"
* Co-editing with MemGTP - "Create a set of documents I can edit in a word processor of my choice, which MemGPT can notice changes to and react over time to."

cpacker:
Added in #174 (still needs a preset toggle via the CLI, coming in the next release).

@genericrich @Yash-Ambekar please feel free to open new PRs / issues with more feature requests.

PR #174, opened 2023-10-29
Allow MemGPT to read/write text files + make HTTP requests

cpacker:
Addresses #26 

---

TODO

- [x] Allow toggling of preset from CLI (right now it's hardcoded)

---

## Example with MemGPT launching HTTP requests

image

---

## Example with MemGPT reading / writing from a shared .txt file

image

```sh
$ cat notes.txt
I will tell you a secret
```

Modify the file, then check if MemGPT can read the updated file:

```sh
$ echo "I like dogs" >> notes.txt
$ cat notes.txt
I will tell you a secret
I like dogs
```

MemGPT can successfully read from the updated file (after some prodding to read more lines):

image

---

### DIFF
--- a/memgpt/main.py
+++ b/memgpt/main.py
@@ -445,7 +445,7 @@ async def run_agent_loop(memgpt_agent, first, no_verify=False, cfg=None, legacy=
                     continue
 
                 elif user_input.lower() == "/dump":
-                    await print_messages(memgpt_agent.messages)
+                    await memgpt.interface.print_messages(memgpt_agent.messages)
                     continue
 
                 elif user_input.lower() == "/dumpraw":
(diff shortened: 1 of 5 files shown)


cpacker:
@vivi do we have to do anything special to support toggling the preset in the CLI?

Or can we simply add another arg here (and up the stack): https://github.com/cpacker/MemGPT/blob/main/memgpt/main.py#L165

memgpt/main.py
```diff
@@ -304,7 +304,8 @@ async def main(
     chosen_persona = cfg.memgpt_persona
 
     memgpt_agent = presets.use_preset(
-        presets.DEFAULT,
+        # presets.DEFAULT,
+        "memgpt_extras",
```
cpacker, line 308: Need to rebase on main and test that we can launch the "extras" preset using the preset flag/config.

memgpt/main.py
```diff
@@ -445,15 +445,15 @@ async def run_agent_loop(memgpt_agent, first, no_verify=False, cfg=None, legacy=
                     continue
 
                 elif user_input.lower() == "/dump":
-                    await print_messages(memgpt_agent.messages)
+                    await memgpt.interface.print_messages(memgpt_agent.messages)
```
cpacker, line 448: This patches a bug

cpacker:
Still need to add preset selection, put it separate PR

vivi (approved):
😈 LGTM

PR #178, opened 2023-10-29
[Draft] accepting external connections via WS

web3wes:


cpacker:
Hey @web3wes , cool stuff. Going to mark this issue as a [Draft] for clarity since you mentioned it's for saving work.

This isn't made super clear in our documentation, but the way the code was written was meant to allow for "swappable" interfaces, so if you wanted eg to write a discord bot where the discord bot sends messages via discord, you use the same MemGPT code but you just provide a new interface, eg `DiscordInterface` that implements all the interface methods (eg `interface.assistant_message`), and treats these as "events" that it can send messages on.

I noticed that you're doing a similar thing here, but I would recommend structuring it in the same way that we did the Discord code - where you create a new interface class / object that has all the same methods but eg uses websockets, instead of overwriting the existing interface file.

tl;dr looks like your code is going in the right direction would just recommend making a new interface class instead of overwriting

web3wes:
@cpacker  Got it! I'll get into that, and I'll also look into adding to https://github.com/goetzrobin/MemGPT/tree/feat-web-ui

People seems to like his UI.

cpacker:
Closing since this seems redundant to new API in #627

PR #181, opened 2023-10-29
Bug/updating max new tokens

web3wes:


memgpt/main.py
web3wes, line 1: Please don't merge if this will delete main.py

PR #182, opened 2023-10-29
update max new tokens

web3wes:


cpacker (approved):
Going to merge but we should update this to a more sane number (maybe 8k? also truncation length should be 8k)

Merging since web UI settings will be updated soon anyways

PR #185, opened 2023-10-29
Update agent.py

tractorjuice:
Fix missing bracket in construct_system_with_memory


cpacker (approved):
nice catch, thank you!

PR #187, opened 2023-10-29
Adding max_new_tokens fixed the problem with partial JSON

oderwat:
After setting the max_new_tokes double as high as the default, I did not experience more partial JSON replies from "dolphin-2.1-mistral-7b". I did get one that had line feeds in it, but could not reproduce that. I guess it would be enough actually remove all line feeds in the answer (as JSON should work without them)

### DIFF
--- a/memgpt/local_llm/webui/settings.py
+++ b/memgpt/local_llm/webui/settings.py
@@ -9,5 +9,6 @@ SIMPLE = {
         # '\n#',
         # '\n\n\n',
     ],
-    "truncation_length": 4096,  # assuming llama2 models
+    "truncation_length": 4096,  # assuming llama2 models,
+    "max_new_tokens": 1000
 }


cpacker:
Thank you for the PR @oderwat !

Is there something more principled we could set this to? Also do we understand why this is an important change? It's not very clear to me from: https://github.com/oobabooga/text-generation-webui/blob/262f8ae5bb49b2fb1d9aac9af01e3e5cd98765db/docs/03%20%E2%80%90%20Parameters%20Tab.md?plain=1#L33

E.g. standardize on either 4k or 8k context length models (maybe a separate settings file for each?)

oderwat:
@cpacker I actually have no idea. When increasing I get better results (until now, I never get truncated JSON anymore). I also get longer replies when doing the same in textgen webui. I guess the other way to try to make sure that the query is complete would be to ask the LLM to continue the answer if parsing fails, like the continue button in webui.

oderwat:
I actually missed that #182 was already here. I guess for a similar problem. Not sure about the length though.

cpacker:
Updated same value in #182

PR #188, opened 2023-10-30
I added a "/retry" command to retry for getting another answer.

oderwat:
I added a "/retry" command to retry for getting another answer.
    
- Implemented by popping messages until hitting the last user message. Then extracting the users last message and sending it again. This will also work with state files and after manually popping messages.
- Updated the README


cpacker (approved):
this LGTM

vivi:
Could you also update `README.md`? Thank you!

memgpt/main.py
```diff
@@ -438,6 +439,12 @@ async def main(
                         memgpt_agent.messages.pop()
                     continue
 
+                elif user_input.lower() == "/retry":
+                    print(f"Retry for another answer")
+                    memgpt_agent.messages.pop()
```
cpacker, line 444: @oderwat I think this code works if you assume the assistant always finishes with a send_message, since that's 2 messages (role assistant -> role function).

However in some cases the assistant might not finish with a function call, so we should probably modify the code to manage this case.

Maybe something like a "pop" that goes in a loop until the item about to be popped is role == user?

cpacker (changes requested):

cpacker:
Also thanks for this PR, /retry seems will be a useful function to have

oderwat:
Well, I am not even a Python programmer (more the Go person) and I just skimmed over the implementation. I guess that checking that one pops up to and including the last user message is fine. I would need to figure out how to do that :)

oderwat:
I decided to check out how to make a better implementation for '/retry'. I think I got it. But need your review.

oderwat:
I would love to make something like "/retry Yes, " or "/retry No, " where the string is being added as the start of the assistant reply to further shape the results. But I am unsure if that is possible with the current design. I guess one would need to extend the function calls with something like "prepend" for the agent?

memgpt/main.py
oderwat: I rewrote the whole implementation and this should not be a problem anymore

oderwat:
I updated the README and formatted the code. I hope this fits now.

oderwat:
I added the update for the `pop <count>` in the README and made the default to 3 instead of 2 (so it removes fun/assistant/user) which seems cleaner to me.

cpacker (approved):
Same FYI as the other PR @oderwat , but just know that this will not modify the persistence manager, only the state inside the agent messages list.

Otherwise, LGTM

image

cpacker:
Thank you for the contribution! /retry, /rethink, and /rewrite are all really cool commands that I'm sure many others will find very useful, thank you for adding them and opening PRs to contribute back into the main repo.

PR #200, opened 2023-10-30
I am unhappy to be forced to "your" timezone and AM/PM.

oderwat:
- Timezone is now an argument (use --timezone="America/Los_Angeles" for the old behaviour.
- The time format is RFC3999 (without microseconds) which is a good format for "the world". I think it is better to not push the AM/PM system to the rest of the world.

### DIFF
--- a/memgpt/main.py
+++ b/memgpt/main.py
@@ -132,6 +132,7 @@ def legacy_run(
     persona: str = typer.Option(None, help="Specify persona"),
     human: str = typer.Option(None, help="Specify human"),
     model: str = typer.Option(constants.DEFAULT_MEMGPT_MODEL, help="Specify the LLM model"),
+    timezone: str = typer.Option("", help="Sets the timezone (useful if the host is not in your local time)"),
     first: bool = typer.Option(False, "--first", help="Use --first to send the first message in the sequence"),
     debug: bool = typer.Option(False, "--debug", help="Use --debug to enable debugging output"),
     no_verify: bool = typer.Option(False, "--no_verify", help="Bypass message verification"),
@@ -167,6 +168,9 @@ def legacy_run(
     typer.secho("Warning: Running legacy run command. Run `memgpt run` instead.", fg=typer.colors.RED, bold=True)
     if not questionary.confirm("Continue with legacy CLI?", default=False).ask():
         return
+
+    if timezone != "":
+        os.environ["TZ"] = timezone
 
     loop = asyncio.get_event_loop()
     loop.run_until_complete(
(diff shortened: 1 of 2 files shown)


sarahwooders:
We can perhaps add timezone as an additional configuration in the refactored CLI #154, which will deprecate the current `memgpt` command.

oderwat:
@sarahwooders sounds good to me, as long one can simply use the current locale too. I was thinking about adding a parameter for am/pm vs 24h time, but I had problems with the application pick the corrent "last saved session". I think it used the wrong one because of string ordering issues when not using 24h time. But I did not dig into the code to verify that.

cpacker:
image

oderwat:
@cpacker I guess I did it wrong by not wearing flowers in my hair using the app ;-)

sarahwooders:
@oderwat we just merged the changes for the updated CLI which you can run with `memgpt run`. I'd recommend adding the timezone into the config https://github.com/cpacker/MemGPT/blob/main/memgpt/config.py#L39 and making it configurable in the CLI https://github.com/cpacker/MemGPT/blob/main/memgpt/cli/cli_config.py.

oderwat:
I will look at it in the next days (1 am here).

oderwat:
Well I have rebased the PR to current main and reformatted the code. But I actually have no idea on how to add the timezone to the new config setting in the correct way. So if @sarahwooders could help with that I would be very happy. I do think you need to let it run by default with the users timezone and use the RFC3999 format for creating the filenames, so they are sortable and meaningful for the user. One could also use UTC for them though.

sarahwooders:
@oderwat I'd recommend adding a `timezone` field to `MemGPTConfig` that is configurable to when you run `memgpt configure` (defined in `memgpt/cli/cli_load.py`). In, inside of the functions in `utils.py`, you can get the config with `config = MemGPTConfig.load()` to determine what timezone to use. 

Let me know how that sounds / if anything is unclear!

oderwat:
You are still forcing that timezone on anybody. I really wonder how one can be so ignorant about the rest of the world.

cpacker:
Hey @oderwat I made a separate PR for this that just pulled local timezone by default, it's merged into main now: #298 (added you as a co-author on the commit).

The reason for this is I'm hesitant to add timezone as a command-line arg without properly adding it to the config setup we have right now (as @sarahwooders mentioned). Maybe that can come in a separate PR.

oderwat:
@cpacker I am happy with the local timezone, that is part of why I did not try harder with figuring out on how to add this to the config. But also, because I bet that @sarahwooders needs just some minutes to add it, while I still do not understand where it actually has to be placed. I am really not very much of a Python programmer. My last real production code was written with Python 2.x and WxWidgets back in 2009 I think :)

PR #204, opened 2023-10-30
I made dump showing more messages and added a count (the last x)

oderwat:
There seem to be some changes about the implementation so that the current dump message helper functions do not show a lot of useful info.

I changed it so that you can `dump 5` (last 5 messages) and that it will print user readable output. This lets you get some more understanding about what is going on.

As some messages are still not shown I also show the index (reverse) of the printed message, so one can see what to "pop" to reach a special point without getting into the `drumpraw`.

Try it with something like `/dump 10` which will show the last 10 messages in a nice readable way.


oderwat:
Rebase on main / black'ed and updated README.

oderwat:
I find it quite disturbing that the current version of the application still just crashed with `/dump1` and my changes did not get being considered. But now the code is rewritten in a way, that I don't even want to share my changes as a PR anymore.

cpacker:
Hey @oderwat sorry we didn't get to this PR in a timely manner, we're just flooded with PRs and urgent feature requests that it's hard to keep track of what PRs can be easily merged or not and some things get lost.

Reopening to review and merge now.

cpacker:
Also in the future if you feel like we've lost track of a PR that we should be merging in ASAP, please feel free to bump the PR with a comment + @ us.

cpacker:
Example output on `/dump`

image

cpacker:
Merging (test should pass once on main, should have an early exit for no key on fork PRs) - thank you for the contribution @oderwat

PR #208, opened 2023-10-31
Changes to lmstudio to fix JSON decode error

raisindetre:
Hi @cpacker . I've made some tweaks to the LM Studio `settings.py` and `api.py` code which (for me at least) have resolved the issues using LM Studio as a back-end.

- Changed the endpoint to use `http://localhost:1234/v1/chat/completions` 
- Set stream to false per LMStudio curl API example as think it might provide a performance gain.
- Rewrapped the `prompt` JSON object within `message` object for compatability with endpoint and updated reference to response text accordingly.
- Set the `context_overflow_policy` to  option 2 (rolling context window)

In local testing I'm getting expected behavior in chats and much improved performance. Might be worth checking the code in Windows for compatability (I don't have access at the moment).

### DIFF
--- a/memgpt/local_llm/lmstudio/settings.py
+++ b/memgpt/local_llm/lmstudio/settings.py
@@ -9,5 +9,8 @@ SIMPLE = {
         # '\n#',
         # '\n\n\n',
     ],
-    "max_tokens": 500,
+    "max_tokens": 3072,
+    "lmstudio": {"context_overflow_policy": 2},
+    "stream": False,
+    "model": "local model",
 }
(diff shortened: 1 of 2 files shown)


memgpt/local_llm/lmstudio/settings.py
```diff
@@ -9,5 +9,8 @@
         # '\n#',
         # '\n\n\n',
     ],
-    "max_tokens": 500,
+    "max_tokens": 3072,
+    "lmstudio": {"context_overflow_policy": 2},
```
cpacker, line 13: Ideally we shouldn't really be using any context overflow policies, since MemGPT has its own way of handling this and this might conflict. But if this works as a bandaid fix for some other bug wrt open LLM integration I'm OK with leaving it for now.

raisindetre: Ah ok. Well it may work ok without it. As you say, I think its the endpoint change that's doing the heavy lifting for... reasons lol.

Issue #197, opened 2023-10-30
Don't crash on rate limit error


cpacker:
"we shouldn't let memgpt crash if you have a retry error like this, instead it should just let you attempt to try again or /save to exit the session"

cpacker:
Closed by #216

PR #216, opened 2023-10-31
Gracefully catch errors when running agent.step()

vivi:
We don't want exceptions from agent.step() to cause the CLI to exit and irrevocably lose conversation history. 

- Asks user if they want to retry upon failed step()
- If user answers no, they are sent back to providing user input
image

### DIFF
--- a/memgpt/main.py
+++ b/memgpt/main.py
@@ -7,6 +7,7 @@ import glob
 import os
 import sys
 import pickle
+import traceback
 
 import questionary
 import typer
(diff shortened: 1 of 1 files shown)


vivi:
Addresses #197.

memgpt/main.py
```diff
@@ -512,24 +513,32 @@ async def run_agent_loop(memgpt_agent, first, no_verify=False, cfg=None, legacy=
...
+                        user_message = system.get_token_limit_warning()
+                        skip_next_user_input = True
+                    elif function_failed:
+                        user_message = system.get_heartbeat(constants.FUNC_FAILED_HEARTBEAT_MESSAGE)
+                        skip_next_user_input = True
+                    elif heartbeat_request:
+                        user_message = system.get_heartbeat(constants.REQ_HEARTBEAT_MESSAGE)
+                        skip_next_user_input = True
+            except Exception as e:
+                print("An exception ocurred when running step(): ")
```
cpacker, line 537: Can we change this to `agent.step()`

memgpt/main.py
```diff
@@ -512,24 +513,32 @@ async def run_agent_loop(memgpt_agent, first, no_verify=False, cfg=None, legacy=
...
+                traceback.print_exc()
+                retry = await questionary.confirm("Retry step()?").ask_async()
```
cpacker, line 539: And change this to "MemGPT encountered an error while trying to run agent.step(). Retry?`

cpacker (changes requested):
Otherwise LGTM

cpacker (approved):

PR #218, opened 2023-10-31
I added commands to shape the conversation:

oderwat:
`/rethink <text>` will change the internal dialog of the last assistant message.
`/rewrite <text>` will change the last answer of the assistant.

Both commands can be used to change how the conversation continues in some pretty drastic and powerful ways.

### DIFF
--- a/memgpt/main.py
+++ b/memgpt/main.py
@@ -484,23 +484,23 @@ async def run_agent_loop(memgpt_agent, first, no_verify=False, cfg=None, legacy=
                     continue
 
                 elif user_input.lower() == "/rethink" or user_input.lower().startswith("/rethink "):
-                    if len(user_input) < 9:
+                    if len(user_input) < len("/rethink "):
                         print("Missing text after the command")
                         continue
                     for x in range(len(memgpt_agent.messages) - 1, 0, -1):
                         if memgpt_agent.messages[x].get("role") == "assistant":
-                            text = user_input[9:].strip()
+                            text = user_input[len("/rethink ") :].strip()
                             memgpt_agent.messages[x].update({"content": text})
                             break
                     continue
 
                 elif user_input.lower() == "/rewrite" or user_input.lower().startswith("/rewrite "):
-                    if len(user_input) < 9:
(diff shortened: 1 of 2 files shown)


cpacker:
@oderwat would you mind running `black` on your code to make it pass the formatting check so we can merge it more easily? https://github.com/cpacker/MemGPT#development

Alternatively if you add me as a contributor to your fork repository I can add this for you.

Thank you for another great PR btw!

cpacker (changes requested):
Needs black reformatting (or add me as a collaborator to your fork @oderwat and I can patch it)

oderwat:
I use PyCharm Professional and just enabled the integrated Black formatter. I will go through the PRs and update them.

oderwat:
OK. This did not work well. I also think that the setup of Black may be different. Have to investigate.

oderwat:
OK. I guess it worked now. I also added @cpacker as collaborator, just in case.

oderwat:
OK. I was in rebase hell because for some stupid reason I could not get upstream to my local origin. I hope that now everything is fine. There are two unrelated problems though: 1. Black reports an error in utils (but I did not change that) and main.py uses `print_messages` without importing it.

memgpt/main.py
```diff
@@ -482,6 +483,32 @@ async def run_agent_loop(memgpt_agent, first, no_verify=False, cfg=None, legacy=
                         memgpt_agent.messages.pop()
                     continue
 
+                elif user_input.lower() == "/rethink" or user_input.lower().startswith("/rethink "):
+                    # TODO this needs to also modify the persistence manager
```
cpacker, line 487: Just as an FYI @oderwat , your implementation here isn't also overwriting the persistence manager's history, so if the agent uses conversation_search to look into the past, they'll find the original unedited message.

However, it is a significant amount of work to get the persistence manager to also be in sync (since the persistence manager can have many implementations), so I think it's OK to have this issue and just leave a disclaimer, since the commands as-is are still quite useful.

cpacker (approved):

cpacker:
Thank you for your contribution!

PR #219, opened 2023-10-31
Added auto-comment.yml

Killer2OP:


cpacker:
Thank you for the PR! We don't think we want to add auto commenting for the meantime but may reconsider in the future.

Issue #142, opened 2023-10-26
Support using a vector database for archival storage 


sarahwooders:
Archival and recall memory should be able to be backed up into a database, rather than in a local FAISS index. 

MemGPT can support the following options for storing data: 
* Local FAISS index 
* Local vector DB 
* Remote vector DB

lightningorb:
https://github.com/pgvector/pgvector-python/blob/master/examples/openai_embeddings.py

sarahwooders:
It looks like Llama Index has a nice example with [Postgres](https://gpt-index.readthedocs.io/en/latest/examples/vector_stores/postgres.html), so we can add a `PostgresArchivalMemory(ArchivalMemory)` that uses this once #154 is merged.

PR #226, opened 2023-10-31
VectorDB support (pgvector) for archival memory 

sarahwooders:
Addresses #142 

* This pull request will allow for archival memory to be stored in a vector DB (pgvector) in addition to local. This will help with retrieving from larger datasets that can't always be stored locally. 
* Agent archival memory can be loaded into multiple times, via connecting agents to multiple data sources 
* Load text/embeddings from an existing vector DB 

image

To-Dos: 
- [x] Load data via `memgpt load` into pgvector 
- [x] Update `memgpt list sources` to look at DB tables
- [x] Enable multiple data sources per agent
- [x] Enable configuration of archival storage location (DB endpoint, local) 

### Configuration
When running `memgpt configure`, user will have the option to choose between `local/postgres/chroma`, and provide their chroma/postgres URI which will be used to store data. This will configure the data backend that MemGPT uses to save and read archival storage. 

NOTE: Users can swap out the data backend, but a previously loaded data source or saved agent will no longer be accessible until they switch back to the original backend. 
NOTE: Do not add a storage backend to MemGPT unless you are okay with MemGPT writing new tables/data. 

## Importing data 
Imported data and agent archival data will be stored in the chosen data backend, not just locally. 

So if Postgres is the data backend, the following command will load data into Postgres: 
```
memgpt load directory --name <data-source-name> --input-dir <dir> --recursive 
```
An agent can access the data in multiple ways. Both methods will *copy* the attached data into the agent's index. 

1. Attach the data source to an agent (agent must already exist from prior `memgpt run`).
```
memgpt attach --agent <agent-name> --data-source <data-source-name> 
```

2. Load the data from the CLI 
```
memgpt run --agent <agent-name> 
... 
> Enter your message: /attach
? Select data source short-stories
```
Once data is attached to an agent, it cannot be removed. We can address this in a future issue. 

## Loading from a VectorDB 

Users should not provide their vector DB as a storage backend unless they are ok with MemGPT writing to the DB. If you want to make read-only data accessible to MemGPT, you must load the data via: 
```
memgpt load vector-database --name <name> --uri <url> --embedding-col <embedding-col> --text-col <text-col> 
```

# Future Issues 
* Deleting data sources 
* Removing data sources from an agent 
* Adding chroma


silence48 (commented):
Can we make it possible to pass a connection string instead and would be nice to set overrides for the table and column names i have preloaded vectors.

Also i saw a comment that says it will mess up existing data but didnt look why... thats not ideal. 

I will try to test tomorrow and add some experience feedback.

sarahwooders:
@silence48 what kind of database do you have? I'm using the LlamaIndex wrappers for retrieving from Postgres/Chroma for now, but it seems like they dont allow for configuring the column names (https://github.com/run-llama/llama_index/issues/6058). But we can add a non-LlamaIndex connector that is more configurable. 

Another issue with bringing your own table is that I'm not sure how we should store new archival memories generated by the agent. Presumably you don't want MemGPT to insert new documents into your existing table? There's two possible solutions I can think of for this: 
1. Copy the user's table into another table which MemGPT can write to 
2. Have MemGPT search across both a writeable archival memory table and also the user's table

silence48:
> 
> 1. Copy the user's table into another table which MemGPT can write to
> 2. Have MemGPT search across both a writeable archival memory table and also the user's table

This would work for me i think.

Its in a postgres database but i have 2 tables with vectors one called codevecs and one called docvecs. I had set it up following the readme in the pgvector repo.

quantumalchemy:
so If I have a local chroma db (persistent) located ie: path: /data/chromadb/
 can I connect it to memgpt?
load vector-database --name <name> --uri <url>  
** can the url be just  a local path?

tests/test_storage.py
```diff
@@ -0,0 +1,33 @@
...
+import argparse
+
+
+def test_postgres():
+    config = MemGPTConfig.load()
+    embed_model = embedding_model()
+
+    passage = ["This is a test passage", "This is another test passage", "Cinderella wept"]
+
+    db = PostgresStorageConnector(uri=config.archival_storage_uri, table_name="test2")
```
cpacker, line 15: This throws an error:

```
ERROR tests/test_storage.py - TypeError: PostgresStorageConnector.__init__() got an unexpected keyword argument 'uri'
```

Because of the signature here:

```

class PostgresStorageConnector(StorageConnector):
    """Storage via Postgres"""

    # TODO: this should probably eventually be moved into a parent DB class

    def __init__(self, name: Optional[str] = None, agent_config: Optional[AgentConfig] = None):
        config = MemGPTConfig.load()
```

vivi:
I think we need to update poetry.lock and requirements.txt?

```
ModuleNotFoundError: No module named 'pgvector'
```

PR #242, opened 2023-11-01
updating overwrite error

web3wes:
updating error when user attempts to overwrite existing persona, model, or human


web3wes:
image

web3wes:
image

sarahwooders (changes requested):
This looks great! One minor thing that would be great to add is to check if the flag conflicts with the `agent_config`. So if they pass in `--persona sam`, but `agent_config.persona == "sam"`, then don't fail since they're not trying to override with a new persona.

cpacker:
should be handled in latest versions, feel free to reopen/rebase if not

PR #243, opened 2023-11-01
Adding list to select state from /load

web3wes:
Currently load only grabs the most recent saved_state. This list show allows users to pick a save state and load it. 

(Loading implementation in progress)

### DIFF
--- a/memgpt/main.py
+++ b/memgpt/main.py
@@ -86,7 +86,28 @@ def save(memgpt_agent, cfg):
     cfg.write_config()
 
 
-def load(memgpt_agent, filename):
+def list_all_checkpoints():
+    """Return a list of all checkpoint files sorted based on modified timestamp (most recent first)."""
+    save_path = f"{constants.MEMGPT_DIR}/saved_state"
+    json_files = glob.glob(f"{save_path}/*.json")
+    return sorted(json_files, key=os.path.getmtime, reverse=True)
+
+
+#please let me know if we don't want async here.
+async def load(memgpt_agent, filename):
+    # If no file is specified, ask user to select from list of checkpoints
+    if filename is None:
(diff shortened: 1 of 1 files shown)


memgpt/main.py
```diff
@@ -86,7 +86,27 @@ def save(memgpt_agent, cfg):
...
-def load(memgpt_agent, filename):
+def list_all_checkpoints():
+    """Return a list of all checkpoint files sorted based on modified timestamp (most recent first)."""
+    save_path = f"{constants.MEMGPT_DIR}/saved_state"
+    json_files = glob.glob(f"{save_path}/*.json")
+    return sorted(json_files, key=os.path.getmtime, reverse=True)
+
+
+# please let me know if we don't want async here.
+async def load(memgpt_agent, filename):
```
sarahwooders, line 97: Could you remove the async?

sarahwooders, line 97: allow `filename=None`.

memgpt/main.py
```diff
@@ -86,7 +86,27 @@ def save(memgpt_agent, cfg):
...
+    # If no file is specified, ask user to select from list of checkpoints
+    if filename is None:
+        checkpoints = list_all_checkpoints()
+        if not checkpoints:
+            print("/load error: no .json checkpoint files found")
+            return
+
+        filename = await questionary.select(
```
sarahwooders, line 105: I think it would be a bit cleaner to have the file selection with questionary happen inside of main.py, so when the user inputs `\load` it prompts them with a list of checkpoints to select from, and then you pass that checkpoint into `load(agent, filename)`.

sarahwooders (changes requested):

cpacker:
Deprecated by 0.3

PR #246, opened 2023-11-02
Allow loading in a directory non-recursively

vivi:


sarahwooders:
Does it actually work if recursive=False for a directory?

vivi:
Yes --

```
❯ memgpt load directory --name=test --input-dir=test/
Loading data...
Indexing documents...
Warning: get_index(docs) only supported for OpenAI
Open AI embedding cost will be approximately $0.0013577 - continue? [Y/n]: y
/Users/vivian/.memgpt/archival/test
```

sarahwooders (approved):
lgtm!

PR #282, opened 2023-11-03
Added the ability to make calls to JIRA Cloud

cevatkerim:
image

Make sure to have the following ENV variables:
```
JIRA_SERVER
JIRA_USER
JIRA_KEY
```

It can get a single issue or make arbitrary JQL calls to your instance.

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -26,6 +26,8 @@ from .constants import (
     CORE_MEMORY_HUMAN_CHAR_LIMIT,
     CORE_MEMORY_PERSONA_CHAR_LIMIT,
 )
+from jira import JIRA
+from jira.exceptions import JIRAError
 
 
 def initialize_memory(ai_notes, human_notes):
@@ -227,6 +229,8 @@ class Agent(object):
         self.pause_heartbeats_start = None
         self.pause_heartbeats_minutes = 0
 
+        self.jira = None
+
         self.first_message_verify_mono = first_message_verify_mono
 
         # Controls if the convo memory pressure warning is triggered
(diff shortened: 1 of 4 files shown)


sarahwooders:
@cpacker maybe we should add support for user-configurable presets?

cpacker:
@cevatkerim this is awesome, thank you for opening a PR with this!

We probably don't want to create a dependency for `jira` for all users, since only a fraction will want the functionality.

@sarahwooders yeah maybe make the presets toggleable, and presets can also define extra packages they need?

cevatkerim:
> @cevatkerim this is awesome, thank you for opening a PR with this!
> 
> We probably don't want to create a dependency for `jira` for all users, since only a fraction will want the functionality.
> 
> @sarahwooders yeah maybe make the presets toggleable, and presets can also define extra packages they need?

It would be great to find a solution for this. I am already working on Confluence and some other integrations. How can we make this more extensible?

agiletechnologist:
@cevatkerim wouldn't be using aiohttp, which allows asynchronous requests to work better than requests which is not asynchronous?

cevatkerim:
> aiohttp

Not sure. Are you suggesting that instead of using a jira library?

cpacker:
@cevatkerim let me know what you think of the workflow for adding new functionality like JIRA tickets here: https://github.com/cpacker/MemGPT/pull/420

Once the linked refactor is merged in, I'm happy to rewrite this PR for you in the new preset format.

cpacker:
Hey @cevatkerim just an FYI we recently built in new "memgpt extensions" support via presets and drag-and-drop `.py` functions.

I put your Jira example in the documentation: https://memgpt.readthedocs.io/en/latest/functions/

What do you think? Does this sort of workflow make sense for you?

Happy to also merge your Jira example into the "canonical" examples (for example putting it inside the provided [example function sets](https://github.com/cpacker/MemGPT/tree/main/memgpt/functions/function_sets)).

cpacker:
Rolled into official documentation sometime in 0.2.x as the example. Can re-open later to add into the official repo once we figure out how we want to support community function extensions (as a contrib library maybe?).

PR #283, opened 2023-11-03
strip '/' and use osp.join (Windows support)

cpacker:
Fixes bad path issues, from Discord:

> It looks like the os.sep is not being used in the path. 
> persistence.pickle failed with: [Errno 2] No such file or directory: 'C:\Users\xxx\.memgpt/saved_state\2023-11-03_04_57_10_AM_PDT-0700.persistence.pickle
> As you can see there is an error due to the forward slash being used instead of the back slash right after .memgpt directory

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -369,7 +369,7 @@ class Agent(object):
 
         # load state
         directory = agent_config.save_state_dir()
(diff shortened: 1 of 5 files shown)


vivi (approved):
LGTM 🕺

PR #287, opened 2023-11-03
Cleanup preset configuration 

sarahwooders:


memgpt/presets.py
```diff
@@ -1,41 +1,89 @@
...
+@dataclass
+class Preset:
+    name: str = None
+    pretty_name: str = None
+    functions: Tuple[str] = ()
+    sync: bool = False  # TODO: remove as a preset
+
+
+@dataclass
+class DefaultPreset(Preset):
```
cpacker, line 21: I think this refactor is generally the right idea, however I wonder what the best way is to support users adding their own presets, eg: https://github.com/cpacker/MemGPT/pull/282

I feel like adding presets should ideally be as easy as adding personas/humans or should have a similar workflow, where you create a text file and put it in a folder.

Thoughts?

cpacker (changes requested):
A lot of the code here look good (eg stripping preset to just take config as input), but we should hold until we refactor the preset creation based on #339, then rebase and merge this on top of the refactor (or the other way - open a PR with the refactor and merge this into the PR)

cpacker:
Closing b/c stale after presets refactor, though we may at some point want to add ideas from this PR into the existing refactor (eg adding a "pretty name" field to the preset YAML files).

PR #294, opened 2023-11-04
Fix archival

BobKerns:
This fixes the problems I was encountering with saving archival storage, except for the "can't insert into an empty index", which I'm not sure how it got into that state. Supplying a data source causes it to use that, but the code appears to default to the persistence_manager directory (which existed, and was used successfully by /save).

The two main problems were that not everything accepted the embedding keyword, and it passed the memory_string to the llama_index, rather than the Document that it expected. This led to an error trying to access the metadata string.

I supply some metadata(timestamp and embedding model name, if any). Please review whether that's an appropriate selection.

If you prefer, I can rebase.


BobKerns:
(My testing on a new agent using /memorywarning did not encounter the EmptyIndex problem.)

memgpt/memory.py
```diff
@@ -700,17 +727,23 @@ def save(self):
...
 
         # TODO: figure out if this needs to be refreshed (probably not)
         self.retriever = VectorIndexRetriever(
             index=self.index,
             similarity_top_k=self.top_k,
         )
 
-    async def a_insert(self, memory_string):
-        return self.insert(memory_string)
+    async def a_insert(self, memory_string: str, embedding: Optional[VectorEmbedding] = None):
```
sarahwooders, line 745: Passing in the embeddings into `a_insert` was actually a bug in the code (will be fixed by #316), so `a_insert` and `insert` should actually only take in a memory string. The `ArchivalMemory` class should be responsible for computing the embeddings and any additional metadata, since this may depend on the underlying storage being used.

sarahwooders:
Hi @BobKerns - Thanks for submitting these fixes, and sorry for the delay in the review (we've been making a lot of changes to archival memory management). The additional metadata looks great - would you be able to add it to the `EmbeddingArchivalMemory` which was added a few days ago? https://github.com/cpacker/MemGPT/blob/main/memgpt/memory.py#L745 We are actually migrating towards this class as the default storage, and the other archival memory classes will eventually be deprecated. You might also need to test with both postgres and local storage to make sure that the additional metadata is getting properly saved.

memgpt/openai_tools.py
```diff
@@ -2,9 +2,13 @@
 import random
 import os
 import time
+from typing import cast, TypeAlias
 
 from .local_llm.chat_completion_proxy import get_chat_completion
 
+VectorEmbedding: TypeAlias = list[float]
```
sarahwooders, line 9: This is much neater, thanks!

sarahwooders:
This is the updated code for parsing new memory strings and computing embeddings https://github.com/cpacker/MemGPT/blob/main/memgpt/memory.py#L778 -- basically archival memory is responsible for parsing/embedding the text, and then its passed to the storage layer.

sarahwooders (changes requested):
* Remove passing in `embeddings` into insert 
* Add same changes to `EmbeddingArchivalMemory` class 
* Ensure modifications work for both local/DB storage

BobKerns:
I had been wondering how tightly coupled the embeddings were to front end/back end. Being able to specialize by the storage would enable specialized embeddings for spatial or temporal memory, for example.

I'm setting up to test against postgres. I don't think I'll have a result until tomorrow.

memgpt/memory.py
BobKerns: I thnk I've dealt with this, but I need to re-test. I've set up a docker compose setup with containerized postgres & memgpt that I'll do a separate PR for. I plan to leverage this for some automated tests. Everything appeared to be working before I did a merge yesterday, so I'm optimistic.

(I'll include my tests for inspection, but w/o the docker stuff. That's going to warrant some design discussion to be sure we're on the same page, consider use cases I may have missed, etc.)

I'm leaving this comment unresolved until I'm satisfied with my tests. I hope to be ready by tomorrow.

sarahwooders: Sounds good, let me know when I should re-review!

cpacker:
Largely deprecated by 0.3, lmk if this is not the case (will likely need to open a fresh PR to rebase changes - happy to do this for you).

PR #296, opened 2023-11-04
Better interface output for function calls

### LINKED ISSUE #263: Change how the CLI handles visualization of memory calls
Instead of grepping for keywords, we should just grep to matches on function calls we recognize.

E.g.
```
{
'role': 'assistant', 'content': 'Searching recall memory...',
'function_call': {
  'name': 'send_message', 
  'arguments': '{"message": "I\'ve noticed that you enjoy going for hikes in the mountains and exploring nature. You also have a soft spot for dogs, especially golden retrievers."}'
  }
}
```

Should not trigger a memory styling in the UI (it's a hallucination + send_message shouldn't get a function styling), but it still does:
```
💭 Searching recall memory...
🤖 I've noticed that you enjoy going for hikes in the mountains and exploring nature. You also have a soft spot for dogs, especially golden retrievers.
⚡🟢 [function] Success: None
```

vivi:
Closes #263


cpacker:
Seems to pass basic testing based on #263 

image

cpacker (approved):
lgtm

PR #302, opened 2023-11-04
[Draft] Wrapper for Mistral 7-B with function calling

web3wes:
This is wrapper for a Mistral function calling model. This is a paid HF model, but I've tested it with a user in the discord (I can't see his name for some reason) and it's pretty good at function calls.

The wrapper changes some things about the context of system messages and instructions, and I'm not sure what impact that has on the rest of the UX if any. Open to any feedback. 

Screenshot 2023-11-04 at 3 31 24 AM

The Model:
[Trelis/Mistral-7B-Instruct-v0.1-function-calling-v2
](https://huggingface.co/Trelis/Mistral-7B-Instruct-v0.1-function-calling-v2)

### DIFF
--- a/memgpt/local_llm/chat_completion_proxy.py
+++ b/memgpt/local_llm/chat_completion_proxy.py
@@ -6,7 +6,7 @@ import json
 
 from .webui.api import get_webui_completion
 from .lmstudio.api import get_lmstudio_completion
-from .llm_chat_completion_wrappers import airoboros, dolphin, zephyr, simple_summary_wrapper
+from .llm_chat_completion_wrappers import airoboros, dolphin, zephyr, simple_summary_wrapper, mistral
 from .utils import DotDict
 from ..prompts.gpt_summarize import SYSTEM as SUMMARIZE_SYSTEM_MESSAGE
 from ..errors import LocalLLMConnectionError, LocalLLMError
@@ -43,6 +43,8 @@ def get_chat_completion(
         llm_wrapper = dolphin.Dolphin21MistralWrapper()
     elif model == "zephyr-7B-alpha" or model == "zephyr-7B-beta":
         llm_wrapper = zephyr.ZephyrMistralInnerMonologueWrapper()
+    elif model == "mistral":
+        llm_wrapper = mistral.MistralChatCompletionWrapper()
     else:
         # Warn the user that we're using the fallback
         if not has_shown_warning:
@@ -65,6 +67,7 @@ def get_chat_completion(
 
     try:
         if HOST_TYPE == "webui":
+            print(prompt)
             result = get_webui_completion(prompt)
         elif HOST_TYPE == "lmstudio":
             result = get_lmstudio_completion(prompt)
@@ -106,4 +109,6 @@ def get_chat_completion(
             ),
         }
     )
+    print("response VVVVV")
+    print(response)
     return response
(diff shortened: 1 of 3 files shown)


oderwat:
I saw that model and wondered if it is worth it. "pretty good at function calls" does not really cut it as a reason to get it.

web3wes:
@oderwat that's a valid point.  I can't vouch for that.  I supposed open LLMs leaderboard might be a hint of its performance. I'm not affiliated in any way. 

Just helping @jdo300 with the wrapper because he asked.

web3wes:
@oderwat 

Since I have it, I'd be happy to do any tests you would like to say and share the results if you're interested.

oderwat:
Which brings up the topic of one or more "default" tests that can be done with memGPT for checking out different models with some standard. Is there something like that in the codebase yet?

web3wes:
There isn't that I know of, but I think that's a good idea.

oderwat:
I think you need to clean that code a bit as there are some testing/debugging remains in it.

web3wes:
Agreed. Have been attending to some life stuff for the last few hours but I should be getting back to this tonight.  Thanks for the feedback.

oderwat:
I bought that model out of curiosity but could not get any working results either with your wrapper (refactored for main) nor with another wrapper. It actually never responds with code that parses as the functions needed by memGPT. I may miss something though.

web3wes:
Hey sorry you aren't seeing the results you'd like with the the wrapper. I'd be happy to work with you on that is if you're interested. @oderwat feel free to reach out here or in discord. I am. @reporter

oderwat:
@web3wes I am short in time because "projects" so that must wait. I just wonder if it actually works for you (esp. using current main)

Jdo300:
@oderwat Here's a super barebones script I wrote when I was just testing the prompting format for talking directly to Mistral-7B (the function calling-tuned version). This script confirmed for me that the model was able to respond with the appropriate function-calling syntax. 

I've also been busy lately and haven't had a chance to do some maintenance/refinements on the wrapper code but will take a look when I have a change. For now. Here's a copy of the original test script I wrote (it can be tested standalone as long as you have an API server running (this example works with Mistral-7B loaded into webui as the back-end):

```
import requests
import json

# Define the roles and markers
B_INST, E_INST = "[INST]", "[/INST]"
B_FUNC, E_FUNC = "<FUNCTIONS>", "</FUNCTIONS>\n\n"

# Define the function metadata
function_metadata = {
    "function": "search_bing",
    "description": "Search the web for content on Bing. This allows users to search online/the internet/the web for content.",
    "arguments": [
        {
            "name": "query",
            "type": "string",
            "description": "The search query string"
        }
    ]
}

# Define the user prompt
user_prompt = 'Search for the latest news on AI.'

# Format the function list and prompt
function_list = json.dumps(function_metadata, indent=4)
prompt = f"{B_FUNC}{function_list.strip()}{E_FUNC}{B_INST} {user_prompt.strip()} {E_INST}\n\n"

# Define the API endpoint
url = "http://127.0.0.1:5000/api/v1/generate"

# Send the POST request to the API server
response = requests.post(url, json={"prompt": prompt})

# Print the response
print(response.json())
```

The output I got was this:

```
{'results': [{'text': '{\n  "function": "search_bing",\n  "arguments": {\n    "query": "latest news on ai"\n  }\n}'}]}
```

PR #304, opened 2023-11-04
Stopping the app to repeat the user message in normal use.

oderwat:
- Removed repeating every user message like bein in debug mode
- Re-added the "dump" flag for the user message, to make it look nicer. I may "reformat" other message too when dumping, but that was what sticked out to me as unpleasant.


cpacker (approved):
LGTM - thanks!

PR #305, opened 2023-11-04
I took some time to go through most of the code and added fixes

oderwat:
- Fixed stuff that looked obvious to me
- Added variable initialisations (because I have a zero warnings policy in all my projects)
- Commented out unused imports (these show up as warnings in my IDE)
- Added superclass calls (not sure about that, but I think that one should do that)
- Added FIXME: where I am unsure on how to fix or where something looks fishy.
- Added some requirements (I guess that needs to go to requirements-local.txt too?).
- Adding "fitz" actually wanted me to add frontend and other stuff, that included having a "/static" directory. I am unsure what that is about.
- I found that using the newest main, needs to have the postgresql packages installed, even if you just want to use "local". This meand I had to install postgresql on my mashine (I am a mariadb type).

Note: This was done using the default inspections of PyCharm Professional and may be biased but mostly follow PEP's

@cpacker @sarahwooders I guess this needs to be reviewed quickly, so it does not become outdated. If you don't want such PRs, just tell me. I had some time at hand and my IDE screamed at me. I would suggest using another IDE if those things matter to you (we use all the JetBrains IDEs everywhere in the company)


memgpt/memory.py
```diff
@@ -266,15 +269,17 @@ def search(self, query_string, count=None, start=None):
...
     async def a_search(self, query_string, count=None, start=None):
+        # FIXME: The linter says that these parameters are wrong?
         return self.search(query_string, count=None, start=None)
 
 
 class DummyArchivalMemoryWithEmbeddings(DummyArchivalMemory):
     """Same as dummy in-memory archival memory, but with bare-bones embedding support"""
 
-    def __init__(self, archival_memory_database=None, embedding_model="text-embedding-ada-002"):
+    def __init__(self, archival_memory_database=None, embedding_model_arg="text-embedding-ada-002"):
```
oderwat, line 279: This is to avoid shadowing (I  know it is not important, but remember: Zero warnings policies catch bugs)

memgpt/memory.py
```diff
@@ -557,6 +566,7 @@ def _extract_date_from_timestamp(self, timestamp):
         match = re.match(r"(\d{4}-\d{2}-\d{2})", timestamp)
         return match.group(1) if match else None
 
+    # FIXME: Signature does not match the base class
```
oderwat, line 569: Not sure how important that is. But I think the IDE may have some idea why it made this a warning for me

memgpt/utils.py
```diff
@@ -13,12 +13,14 @@
 import sqlite3
 import fitz
```
oderwat, line 14: This import made a lot of other changes important. Maybe this should get separately and not into utils too?

requirements.txt
```diff
@@ -1,18 +1,26 @@
-colorama
-demjson3
+colorama~=0.4.6
```
oderwat, line 1: It added all the versions. I guess this needs to be modified at least into `>=` but I am not a Pythonian, I have no clue :)

requirements.txt
```diff
@@ -1,18 +1,26 @@
...
-typer
+tqdm~=4.66.1
+typer~=0.9.0
 llama_index
-prettytable
+prettytable~=3.9.0
+requests~=2.31.0
+aiohttp~=3.8.6
+SQLAlchemy~=2.0.22
+fitz>=0.0.1.dev2
```
oderwat, line 22: It had to be "newer" than just adding "fitz" without a version requirement or id could not find the open function. Not sure why that works as it does

memgpt/main.py
```diff
@@ -255,14 +256,14 @@ async def main(
...
-            human_persona = (humans.DEFAULT, None)
+            human_persona = humans.DEFAULT
         else:
             try:
                 humans.get_human_text(human_persona, Config.custom_humans_dir)
                 human_persona = (human_persona, Config.custom_humans_dir)
             except FileNotFoundError:
                 humans.get_human_text(human_persona)
-                human_persona = (human_persona, None)
+                human_persona = human_persona
```
sarahwooders, line 266: Doesn't this need to be a tuple?

memgpt/cli/cli_config.py
```diff
@@ -26,6 +27,14 @@ def configure():
 
     MemGPTConfig.create_config_dir()
 
+    # this is to initialize them, so we do not get linter warnings
+    openai_key = ""
```
sarahwooders, line 31: Can we set this to `None` instead of empty strings, and also place them next to where the value is set? (e.g. like `default_agent` at line 114)

memgpt/config.py
```diff
@@ -248,7 +250,7 @@ def __init__(
         model,
         preset=DEFAULT_PRESET,
         name=None,
-        data_sources=[],
+        data_sources=None,
```
sarahwooders, line 253: What's wrong with this being an empty list?

memgpt/memory.py
sarahwooders: I'm not sure why adding the `_arg` is important? Also FYI all the archival memory classes except for `EmbeddingArchivalMemory` will be deprecated / removed soon.

memgpt/cli/cli_config.py
oderwat: I would use None if the empty string has some (different) meaningful value. But you can use None too I guess.

memgpt/config.py
oderwat: The linter said that initialization should not use a mutating value. I think that "None" given also is different to an empty array is given (in contrast to the initialization with a default). But these are probably all just theoretical interpretations. As I wrote in the intro: I believe in making linters happy and that there is a reason why the compressed knowledge of something as the PyCharm IDE's code analyzer exists.

memgpt/main.py
oderwat: The following code did not use a tuple, that is why the IDE was mourning. But I may have made the change in the wrong place.

oderwat: Skimming the code it seems that there can be a tuple of a string, while a tuple denotes a persona string + a directory path string. I *guess* that somewhere the value can be either and that a tuple with None as second parameter is basically identical with just the string. But I may be wrong about that.

memgpt/memory.py
oderwat: Making it another name shows that you did not accidentally use the wrong name, because that name shadows the already existing variable in the context that has the same name. Doing it like that will avoid running in error when you refactor or actually do it by accident. So having such a rule in the linter saves your butt and as mentioned before: Zero warnings policies may safe the world one day :)

memgpt/config.py
vivi: It's because if you mutate the list, it becomes the default argument for all invocations of the function.

https://docs.python-guide.org/writing/gotchas/

oderwat: @vivi thanks!

memgpt/cli/cli_config.py
sarahwooders: The config parsing code currently assumes empty values are None, so we should set to that.

oderwat: That may be okay!

NOTE: I guess I do too much Go code where this could lead into problems if you also use ENV and the "function everybody uses to read them", which actually returns "" if the variable is not set, what classifies as a design error for me. It makes you aware that "" is not None and None is not "" :)

oderwat: >  and also place them next to where the value is set? 

I am not sure about that. The value should be set to anything before a potential first use. As linters do static analysis, it may be that it "technically" can't be used uninitialized but that can change with any refactoring and "I do not allow warnings in our code". But this is not "our" code of course :)

PR #307, opened 2023-11-04
Seperate out postgres dependencies into an extra

vivi:
In a separate PR, I'd like to also have
```
[tool.poetry.group.dev.dependencies]
black = "^23.10.1"
pytest = "^7.4.3"
```

but we need to make sure all the scripts/github actions/readme run `poetry install --with dev [--all-extras]`


sarahwooders:
Moving to #337

PR #310, opened 2023-11-04
fix: import PostgresStorageConnector only if postgres is selected as …

goetzrobin:
…storage type

currently there is a requirements issue that makes memgpt crash if the necessary postgres libraries are missing. This change lets people select local as storage type to continue to use memgpt locally without postgres and not run into these issues.

This allows us to fix the requirements necessary for using Postgres as a backend while local memgpt usage is still possible


goetzrobin:
@vivi I believe this relates to your change here: https://github.com/cpacker/MemGPT/pull/307

sarahwooders (approved):
Good catch, thanks!

PR #314, opened 2023-11-05
add ollama support

### LINKED ISSUE #281: Add / confirm Ollama support

cpacker:
Closes #281

---

Tested with dolphin q6 on MacOS Sonoma, seems to be working fine.

```sh
 % ollama run dolphin2.2-mistral:7b-q6_K
pulling manifest
pulling 5baee5b5452f... 100% |█████████████████████████████████████████████████████████████████████████| (5.9/5.9 GB, 18 MB/s)
pulling a47b02e00552... 100% |████████████████████████████████████████████████████████████████████████████| (106/106 B, 1.1 MB/s)
pulling 9640c2212a51... 100% |██████████████████████████████████████████████████████████████████████████████| (41/41 B, 366 kB/s)
pulling de6bcd73f9b4... 100% |██████████████████████████████████████████████████████████████████████████████| (58/58 B, 748 kB/s)
pulling 585fa22cb5b0... 100% |█████████████████████████████████████████████████████████████████████████████| (455/455 B, 271 B/s)
verifying sha256 digest
writing manifest
removing any unused layers
success
```

```sh
export OPENAI_API_BASE=http://localhost:11434
export BACKEND_TYPE=ollama
export OLLAMA_MODEL=dolphin2.2-mistral:7b-q6_K

python main.py
```

image

### DIFF
--- a/memgpt/local_llm/ollama/api.py
+++ b/memgpt/local_llm/ollama/api.py
@@ -6,9 +6,11 @@ import tiktoken
 from .settings import SIMPLE
 from ..utils import load_grammar_file
 from ...constants import LLM_MAX_TOKENS
+from ...errors import LocalLLMError
 
 HOST = os.getenv("OPENAI_API_BASE")
 HOST_TYPE = os.getenv("BACKEND_TYPE")  # default None == ChatCompletion
+MODEL_NAME = os.getenv("OLLAMA_MODEL")  # ollama API requires this in the request
 OLLAMA_API_SUFFIX = "/api/generate"
 DEBUG = False
 
@@ -24,14 +26,18 @@ def get_ollama_completion(prompt, settings=SIMPLE, grammar=None):
     if prompt_tokens > LLM_MAX_TOKENS:
         raise Exception(f"Request exceeds maximum context length ({prompt_tokens} > {LLM_MAX_TOKENS} tokens)")
 
+    if MODEL_NAME is None:
+        raise LocalLLMError(f"Error: OLLAMA_MODEL not specified. Set OLLAMA_MODEL to the model you want to run (e.g. 'dolphin2.2-mistral')")
+
     # Settings for the generation, includes the prompt + stop tokens, max length, etc
     request = settings
     request["prompt"] = prompt
-    request["model"] = TODO
+    request["model"] = MODEL_NAME
 
     # Set grammar
-    # if grammar is not None:
-    # request["grammar_string"] = load_grammar_file(grammar)
+    if grammar is not None:
+        # request["grammar_string"] = load_grammar_file(grammar)
+        raise NotImplementedError(f"Ollama does not support grammars")
 
     if not HOST.startswith(("http://", "https://")):
         raise ValueError(f"Provided OPENAI_API_BASE value ({HOST}) must begin with http:// or https://")
--- a/memgpt/local_llm/ollama/api.py
+++ b/memgpt/local_llm/ollama/api.py
@@ -1,0 +1,57 @@
+import os
+from urllib.parse import urljoin
+import requests
+import tiktoken
+
+from .settings import SIMPLE
+from ..utils import load_grammar_file
+from ...constants import LLM_MAX_TOKENS
+
+HOST = os.getenv("OPENAI_API_BASE")
+HOST_TYPE = os.getenv("BACKEND_TYPE")  # default None == ChatCompletion
+OLLAMA_API_SUFFIX = "/api/generate"
+DEBUG = False
+
+
+def count_tokens(s: str, model: str = "gpt-4") -> int:
+    encoding = tiktoken.encoding_for_model(model)
+    return len(encoding.encode(s))
+
+
+def get_ollama_completion(prompt, settings=SIMPLE, grammar=None):
+    """See https://github.com/jmorganca/ollama/blob/main/docs/api.md for instructions on how to run the LLM web server"""
+    prompt_tokens = count_tokens(prompt)
+    if prompt_tokens > LLM_MAX_TOKENS:
+        raise Exception(f"Request exceeds maximum context length ({prompt_tokens} > {LLM_MAX_TOKENS} tokens)")
+
+    # Settings for the generation, includes the prompt + stop tokens, max length, etc
+    request = settings
+    request["prompt"] = prompt
+    request["model"] = TODO
+
+    # Set grammar
+    # if grammar is not None:
+    # request["grammar_string"] = load_grammar_file(grammar)
+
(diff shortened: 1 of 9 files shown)


memgpt/local_llm/ollama/api.py
```diff
@@ -0,0 +1,62 @@
...
+from ...errors import LocalLLMError
+
+HOST = os.getenv("OPENAI_API_BASE")
+HOST_TYPE = os.getenv("BACKEND_TYPE")  # default None == ChatCompletion
+MODEL_NAME = os.getenv("OLLAMA_MODEL")  # ollama API requires this in the request
+OLLAMA_API_SUFFIX = "/api/generate"
+DEBUG = False
+
+
+def count_tokens(s: str, model: str = "gpt-4") -> int:
```
vivi, line 17: Should move to utils, also appears in webui/api.py

vivi (approved):
🦙 LGTM pending more tests

memgpt/local_llm/ollama/api.py
cpacker: Refactored into utils.py

kfsone:
I'm trying to get it running using memgpt out of the box on windows with ollama running in a docker container. Ollama is there but it doesn't support some of the api endpoints you're trying? The docker container also provides a fairly good representation of how I think average early-access types are going to try and use this.

```
pip install memgpt
docker run -it -d --gpus=all -p11434:11434 -v ollama:/root/.ollama --name ollama ollama/ollama
# check it's working
docker run -exec ollama run llama2
>> /bye
$env:OPENAI_API_BASE="http://127.0.0.1:11434"; $env:OPENAI_API_BASE="sure, why not"; $env:BACKEND_TYPE="ollama"; $OLLAMA_MODEL="llama2"; $env:BACKEND_MODEL="llama2"
memgpt configure
... > http://
... > http://
...
memgpt run
.. new agent ..
404 error accessing /chat/completions
```

cpacker:
@kfsone thank you testing and reporting your bug!

So based on the code, ollama via MemGPT should be trying to hit `/api/generate` (https://github.com/cpacker/MemGPT/blob/main/memgpt/local_llm/ollama/api.py#L13), which is what ollama says their API endpoint is: https://github.com/jmorganca/ollama/blob/main/docs/api.md

But in your output, it seems like MemGPT is trying to hit `/chat/completions`, which means it's not running the ollama code.

```powershell
$env:OPENAI_API_BASE="http://127.0.0.1:11434";
$env:OPENAI_API_BASE="sure, why not";
$env:BACKEND_TYPE="ollama";
$OLLAMA_MODEL="llama2";
$env:BACKEND_MODEL="llama2"
```

^In the above snippet, `$env:OPENAI_API_BASE="sure, why not"` is wrong since it'll overwrite the correct ollama host IP, so what's probably happening is that line of code is causing MemGPT to try and use OpenAI, not Ollama. Additionally, I believe there are typos in your Powershell (I think you need spaces, see: https://memgpt.readthedocs.io/en/latest/quickstart/).

Additionally, I'd recommend using a better model than `llama2`, you will likely have quite a few errors with `llama2` vs a much strong model like `dolphin-mistral`, see: https://memgpt.readthedocs.io/en/latest/ollama/.

Lastly, it should be `pip install pymemgpt`, not `memgpt`.

---

tl;dr:

- `OPENAI_API_BASE` needs to be set correctly
- if you're using powershell, typos in your `$env:` setup (if you're not using powershell, eg WSL, you need to use different commands), see https://memgpt.readthedocs.io/en/latest/quickstart/

So revised code might look like this:

```sh
# setting up ollama docker
docker run -it -d --gpus=all -p11434:11434 -v ollama:/root/.ollama --name ollama ollama/ollama

# install a better model
# Let's use the q6_K variant of dolphin 2.2.1
ollama run dolphin2.2-mistral:7b-q6_K
```

```powershell
## assuming this is powershell, not WSL or cmd.exe

# installing memgpt
pip install pymemgpt

# pointing memgpt at ollama (https://memgpt.readthedocs.io/en/latest/ollama/)
$Env:OPENAI_API_BASE = "http://127.0.0.1:11434" # point at ollama
$Env:OPENAI_API_BASE = "ollama" # required
$Env:OPENAI_API_BASE = "dolphin2.2-mistral:7b-q6_K" # also required (use a better model than llama2)
```

cpacker:
(also, @kfsone if you continue to have issues feel free to open a github issue about this, since it'll make it easier to track and for other users to find the same content later on vs via the comments in a PR)

kfsone:
> $env:OPENAI_API_BASE="http://127.0.0.1:11434";
> $env:OPENAI_API_BASE="sure, why not";
> $env:BACKEND_TYPE="ollama";
> $OLLAMA_MODEL="llama2";
> $env:BACKEND_MODEL="llama2"

Sorry, messed up in the pasting ... I'm actually putting that in the key env, e.g (just grabbing from my command line right now)
```
[11/08/2023 12:32:32 C:\Users\oliver.smith\.cache\lm-studio\models\thingthatis\dolphin-2.2.1-mistral-7B-GGUF]
Win|PS> $env:OPENAI_API_KEY = "sure, why not" ; $env:OPENAI_API_BASE = "http://192.168.86.47:11434" ; $env:BACKEND_TYPE="ollama" ; $env:OLLAMA_MODEL="dolphin2.2-mistral:7b-q6_K"
```

kfsone:
Yeah, I don't expect much from llama2, it's just a likely natural starting point for people who went with the easy route of running ollama in a docker container -- and given the nature of ollama, that it's from two docker guys who realized models as images, it's going to be a super likely route for people go thru. They're going to expect

<follow ollama docker file instructions>
<set env vars>
<memgpt run --model the model name from ollama>

I see "--model   <-- this is not the model but the wrapper" scattered around code/commits/documents, and and that makes it super hard trying to match things up when you're trying to try before you buy, so to speak.

FWIW: My particular goal here is training against a mono-repos with ~300,000 lines of 1st party C++, 25,000 lines of 1st-party Python and >1mil lines of 3rd party code, but it's a proprietary repos so no sharesies. The thing is, a 15 year old engine worked on by over 150 people tends to have ... differences integrated into it, and engineers questions tend to be like "which of the 9 RNGs is it better to use from a thread in the deferred shader loading pipeline?" and the answer any of us give is _almost_ always wrong because within the context of that library target, memory allocation would amount in a 90% decrease in framerate... And another thing I'd like to do is see if I can build something with tooling like eg autogen whereby a team of agents can actually read the whole of a particular go or python tool's code and _rewrite_ it idiomatically in the other language. Not translate, not function at a time, but actually roll the whole thing back. I mean, even rewriting it from scratch in the SAME language ought to some times produce some really fascinating insights...

PR #284, opened 2023-11-03
Remove embeddings as argument in archival_memory.insert

cpacker:

PR #316, opened 2023-11-05
Remove embedding field being passed into archival memory insert

sarahwooders:
The `ArchivalMemory` class `insert(self, memory_string)` does not have an embedding field (and should not, since we want to compute embeddings as defined by the sub-class), so I removed where its being passed in.


vivi:
duplicate of #284?

PR #326, opened 2023-11-05
Bug fix grammar_name not being defined causes a crash

borewik:
grammar_name has to be defined, if not there's an issue with line 92.
Without that, when starting local llm with webui and model e.g. airoboros-l2-70b-2.1 it will crash.


cpacker (approved):
Thanks for catching this!

PR #336, opened 2023-11-06
Add `memgpt version` command and package version

sarahwooders:


memgpt/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.1.18-alpha.1"
```
vivi, line 1: Instead of changing this everytime we bump version, we can use:

```
import importlib.metadata
importlib.metadata.version('pymemgpt')
```

PR #341, opened 2023-11-07
I added some json repairs that helped me with malformed messages

oderwat:
There are two of them: The first will remove hard line feeds that appear in the message part because the model added those instead of escaped line feeds. This happens a lot in my experiments and that actually fixes them.

The second one is less tested and should handle the case that the model answers with multiple blocks of strings in quotes or even uses unescaped quotes. It should grab everything betwenn the message: " and the ending curly braces, escape them and makes it propper json that way.

Disclaimer: Both function were written with the help of ChatGPT-4 (I can't write much Python). I think the first one is quite solid but doubt that the second one is fully working. Maybe somebody with more Python skills than me (or with more time) has a better idea for this type of malformed replies.

### DIFF
--- a/memgpt/local_llm/json_parser.py
+++ b/memgpt/local_llm/json_parser.py
@@ -46,18 +46,104 @@ def add_missing_heartbeat(llm_json):
     raise NotImplementedError
 
 
+def repair_json_string(json_string):
+    """
+    This function repairs a JSON string where line feeds were accidentally added
+    within string literals. The line feeds are replaced with the escaped line
+    feed sequence '\\n'.
+    """
+    new_string = ""
+    in_string = False
+    escape = False
+
+    for char in json_string:
+        if char == '"' and not escape:
+            in_string = not in_string
+        if char == "\\" and not escape:
+            escape = True
+        else:
+            escape = False
+        if char == "\n" and in_string:
+            new_string += "\\n"
+        else:
+            new_string += char
+
+    return new_string
+
+
+def repair_even_worse_json(json_string):
+    """
+    This function repairs a malformed JSON string where string literals are broken up and
+    not properly enclosed in quotes. It aims to consolidate everything between 'message': and
+    the two ending curly braces into one string for the 'message' field.
+    """
+    # State flags
+    in_message = False
+    in_string = False
+    escape = False
+    message_content = []
+
+    # Storage for the new JSON
+    new_json_parts = []
+
+    # Iterating through each character
+    for char in json_string:
+        if char == '"' and not escape:
+            in_string = not in_string
+            if not in_message:
+                # If we encounter a quote and are not in message, append normally
+                new_json_parts.append(char)
+        elif char == "\\" and not escape:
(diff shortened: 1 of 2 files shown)


oderwat:
@cpacker I had good success with the first repair. Not sure what other people get as decoding problems, but it was quite a lot the additional line feeds in answers that beside of that were quite good.

memgpt/local_llm/json_parser.py
```diff
@@ -46,18 +46,104 @@ def add_missing_heartbeat(llm_json):
...
+    # Joining everything to form the new JSON
+    repaired_json = "".join(new_json_parts)
+    return repaired_json
+
+
 def clean_json(raw_llm_output, messages=None, functions=None):
     """Try a bunch of hacks to parse the data coming out of the LLM"""
-
     try:
+        print("clean json runs:", raw_llm_output)
```
cpacker, line 127: Can we put these on a debug flag? eg a `printd` so that it's hidden from the user

cpacker (changes requested):

memgpt/local_llm/json_parser.py
oderwat: Sorry. I forgot to remove them! But adding them as printd may help with analyzing. This or similar problems.

oderwat: I made the output only appear when the debug flag is set and removed the "clean" variant. I needed to import `printd` local to the function because of circular imports. I did not even know that one can do this. I love the IDE.

cpacker (approved):
This looks great, thank you this PR! I'm sure it will help reduce the number of headaches people run into with JSON/string decoding errors while using local LLMs.

memgpt/local_llm/json_parser.py
```diff
@@ -46,18 +46,114 @@ def add_missing_heartbeat(llm_json):
...
+                    printd('trying adding "}}')
+                    data = json.loads(raw_llm_output + '"}}')
+                except json.JSONDecodeError:
+                    try:
+                        repaired = repair_json_string(raw_llm_output)
+                        printd("trying my repair json:", repaired)
+                        data = json.loads(repaired)
+                    except json.JSONDecodeError:
+                        try:
+                            repaired = repair_json_string(raw_llm_output)
```
cpacker, line 150: Should this one be `repair_even_worse_json`?

tractorjuice: New functionality coming to the API. https://platform.openai.com/docs/guides/text-generation/json-mode

tests/test_json_parsers.py
```diff
@@ -0,0 +1,64 @@
+import json
+
+import memgpt.local_llm.json_parser as json_parser
+
+
+EXAMPLE_MISSING_CLOSING_BRACE = """{
```
cpacker, line 6: @oderwat I added a test file with various JSON, when you find new patches that work well, it might be useful to record the specific JSON string and add it to the test file

PR #355, opened 2023-11-07
Update requirements.txt

tractorjuice:
Enable support for previous API.
The new api is very different.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ demjson3
 faiss-cpu
 geopy
 numpy
-openai
+openai==0.28.1
 pybars3
 pymupdf
 python-dotenv


vivi (approved):
LGTM.

sarahwooders:
Wait could we not do this? I dont think we need requirements.txt

cpacker:
@tractorjuice thanks for catching this!

Actually, requirements.txt is deprecated and shouldn't be used. Instead to install from source, do `pip install -e .` (and the version of openai is frozen atm, it pulls from the `.toml` file).

Closing because we remove requirements.txt in the latest code.

https://memgpt.readthedocs.io/en/latest/contributing/

PR #358, opened 2023-11-07
Remove `requirements.txt` and `requirements_local.txt`

sarahwooders:
Turns out poetry can do `pip install -e .[local,postgres]`, so we can remove the requirements files.


cpacker (approved):

oderwat:
I think that is a bad idea. I don't want to use poetry and the PyCharm IDE has pretty good requirements file support.

cpacker:
> I think that is a bad idea. I don't want to use poetry and the PyCharm IDE has pretty good requirements file support.

How do you use `requirements.txt` with PyCharm? Is it possible to configure it to use the `.toml` file instead? https://www.jetbrains.com/help/pycharm/pyproject-toml-support.html

We removed the `requirements.txt` because it was actually redundant to `pyproject.toml`. When you do `pip install -e .`, it will read from the `pyproject.toml` file (https://github.com/cpacker/MemGPT/blob/main/pyproject.toml) similar to how it reads from a `requirements.txt` file.

On our end, using Poetry allows us to make sure we have exact dependencies configured automatically (we do `poetry add`, then it resolves dependencies and updates the `poetry.lock` file). In terms of other projects using poetry both langchain and llamaindex use it (just off the top of my head).

oderwat:
I don't know how I do that. It just worked out of the box. Well, if `pip install -e .` actually works then that is sufficient indeed. I   guess I need to look at poetry and acted premature and Poetry can be made to work with PyCharm too. All good. Thank you for the pointers!

BTW: I found that having "the exact configuration" is not enough at some times. I quite often I had to switch to other packages to make things work in WSL2 for example. Or even install packages manually or from head, or using Conda.

cpacker:
If you have poetry installed you can generate a requirements.txt too (if you need it) with:
```sh
poetry export --without-hashes --format=requirements.txt > requirements.txt
```

However if PyCharm works fine with the `.toml`, then might as well just do that. If a `requirements.txt` does offer significant advantages for certain users though, perhaps we can consider automatically generating it as part of a GitHub workflow/actions @sarahwooders (but if `.toml` does the same, might as well not to avoid extra complexity / possibility of things going wrong).

oderwat:
Just to let you know: I tried using poetry with todays main and nothing really works. I am not sure what can go so wrong. I seldom had so much problems with installing something "python" from source. With `poetry install` it does not even installs transformers. With `poetry install --all-extras` it installs a bunch of stuff and then fails with: "AttributeError: module 'inspect' has no attribute 'getargspec'. Did you mean: 'getargs'?". From SO it seems to be related to Python 3.11 vs Python 3.10. So I change to Python 3.10, now I could run configure and set up what I use, next I hit `TypeError: '>' not supported between instances of 'int' and 'str'` for ` if prompt_tokens > context_window`. How can your tests run without a failure having such problems in the code? When install from scratch does not work as expected?

PR #370, opened 2023-11-08
Fix summarize

### LINKED ISSUE #352: Set `LLM_MAX_TOKENS` dynamically based on model selected
See https://github.com/cpacker/MemGPT/pull/349#issuecomment-1797857440

### LINKED ISSUE #353: async def summarize_messages_inplace(). Fail to summarize the text
**Describe the bug**
When I put a huge doc (like main.py code) into conversation. MemGPT keeping throw out error :

error = This model's maximum context length is 4097 tokens. However, your messages resulted in 8638 tokens (7919 in the messages, 719 in the functions). Please reduce the length of the messages or functions.

and the code (https://github.com/cpacker/MemGPT/blob/7d6aa4096d5118abc1d9bc2c9a298ed88a9e0da8/memgpt/agent.py#L1117) run in 
`         

            # If we got a context alert, try trimming the messages length, then try again
            if "maximum context length" in str(e):
                # A separate API call to run a summarizer
                await self.summarize_messages_inplace()

                # Try step again
                return await self.step(user_message, first_message=first_message)
`    

and then in the code (https://github.com/cpacker/MemGPT/blob/7d6aa4096d5118abc1d9bc2c9a298ed88a9e0da8/memgpt/agent.py#L1126)
`

    async def summarize_messages_inplace(self, cutoff=None, preserve_last_N_messages=True):
        assert self.messages[0]["role"] == "system", f"self.messages[0] should be system (instead got {self.messages[0]})"

        # Start at index 1 (past the system message),
        # and collect messages for summarization until we reach the desired truncation token fraction (eg 50%)
        # Do not allow truncation of the last N messages, since these are needed for in-context examples of function calling
        token_counts = [count_tokens(str(msg)) for msg in self.messages]
        message_buffer_token_count = sum(token_counts[1:])  # no system message
        token_counts = token_counts[1:]
        desired_token_count_to_summarize = int(message_buffer_token_count * MESSAGE_SUMMARY_TRUNC_TOKEN_FRAC)
        candidate_messages_to_summarize = self.messages[1:]
        if preserve_last_N_messages:
            candidate_messages_to_summarize = candidate_messages_to_summarize[:-MESSAGE_SUMMARY_TRUNC_KEEP_N_LAST]
            token_counts = token_counts[:-MESSAGE_SUMMARY_TRUNC_KEEP_N_LAST]

`

Can not find the latest message (main.py code which I type as input), so it can't summarise the message.

**To Reproduce**

0. set model as gpt-3.5-trubo
1. set multiline_input = True 
at (https://github.com/cpacker/MemGPT/blob/7d6aa4096d5118abc1d9bc2c9a298ed88a9e0da8/memgpt/main.py#L391C5-L391C28)
2. CLI run:  python main.py
3. Typing huge text as input, such as, all code string in main.py itself.
4. Alt+Enter to finish input.
5. get error loop.

**Expected behavior**
The overlength text should be trunc and summarized as acceptable pices of text and saving them.

**Screenshots**
image

---

**How did you install MemGPT?**

  - With `git clone git@github.com:cpacker/MemGPT.git` and `pip install -r requirements.txt`

**Your setup (please complete the following information)**

-  Windows
- cmd.exe
- python 3.8
- 0.1.15-83-gfd045ba

cpacker:
Closes #353

Ideally, we should also fix #352 at the same time.

### DIFF
--- a/memgpt/memory.py
+++ b/memgpt/memory.py
@@ -125,6 +125,8 @@ def summarize_messages(
     message_sequence_to_summarize,
 ):
     """Summarize a message sequence using GPT"""
+    if "gpt-" in model:
+        model = "gpt-3.5-turbo-16k"
 
     summary_prompt = SUMMARY_PROMPT_SYSTEM
     summary_input = str(message_sequence_to_summarize)
@@ -153,6 +155,8 @@ async def a_summarize_messages(
     message_sequence_to_summarize,
 ):
     """Summarize a message sequence using GPT"""
+    if "gpt-" in model:
+        model = "gpt-3.5-turbo-16k"
 
     summary_prompt = SUMMARY_PROMPT_SYSTEM
     summary_input = str(message_sequence_to_summarize)
--- a/memgpt/memory.py
+++ b/memgpt/memory.py
@@ -125,8 +125,10 @@ def summarize_messages(
     message_sequence_to_summarize,
 ):
     """Summarize a message sequence using GPT"""
-    if "gpt-" in model:
+    if "gpt-3.5-turbo" in model:
         model = "gpt-3.5-turbo-16k"
+    elif "gpt-4" in model:
+        model = "gpt-4-1106-preview"
 
     summary_prompt = SUMMARY_PROMPT_SYSTEM
     summary_input = str(message_sequence_to_summarize)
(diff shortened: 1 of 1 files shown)


cpacker:
@cpacker 

To fix #352, we need to treat `LLM_MAX_TOKENS` as an agent config.

Similar to how the agent has a model string, each model also has a max context length. We can do some smart inference of defaults for this, eg we know gpt-3.5-turbo has a 4k context, and gpt-4 has an 8k context.

cpacker:
Closing because I don't know if this is necessary anymore based on default model sizes.

PR #373, opened 2023-11-08
Add pymemgpt-nightly workflow

vivi:
Works locally up to `poetry publish` (so does the other poetry-publish workflow), but should work on Github actions, pending the right `PYPI_NIGHTLY_TOKEN`.

### DIFF
--- a/.github/workflows/poetry-publish-nightly.yml
+++ b/.github/workflows/poetry-publish-nightly.yml
@@ -1,0 +1,60 @@
+name: poetry-publish-nightly
+on:
+  schedule:
+    - cron: '35 10 * * *' # 10:35am UTC, 2:35am PST, 5:35am EST
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  # nightly release check from https://stackoverflow.com/a/67527144
+  check-date:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/poetry-publish-nightly.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Install poetry
+        run: |
+          curl -sSL https://install.python-poetry.org | python3 -
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+      - name: Set release version
+        run: |
+          # Extract the version number from pyproject.toml using awk
+          CURRENT_VERSION=$(awk -F '"' '/version =/ { print $2 }' pyproject.toml | head -n 1)
+          # Export the CURRENT_VERSION with the .dev and current date suffix
+          NIGHTLY_VERSION="${CURRENT_VERSION}.dev$(date +%Y%m%d)"
```
sarahwooders, line 45: Maybe do `date +%Y%m%d%H%M%S` instead so we can publish more than once a day?

sarahwooders (approved):
This looks good for nightly, but if we want to publish more than once a day we should change the version naming.

cpacker (approved):

vivi:
A nightly build is called "nightly".

PR #378, opened 2023-11-08
Prompt selection

Drake-AI:
Add prompt list selection with memgpt configure.

**How to test**
Run memgpt configure and select the desired prompt. You can use your custom prompts inside MEMGPT_DIR/prompts.

**Have you tested this PR?**
Tested with 0.1.18

Prompt selection

**Related issues or PRs**
Not sure, i will take a look.

Not.

**Additional context**
I think it is useful to have the option to choose the prompt.

### DIFF
--- a/memgpt/cli/cli.py
+++ b/memgpt/cli/cli.py
@@ -38,6 +38,7 @@ def run(
     human: str = typer.Option(None, help="Specify human"),
     model: str = typer.Option(None, help="Specify the LLM model"),
     preset: str = typer.Option(None, help="Specify preset"),
+    prompt: str = typer.Option(None, help="Specify prompt"),
     first: bool = typer.Option(False, "--first", help="Use --first to send the first message in the sequence"),
     strip_ui: bool = typer.Option(False, "--strip_ui", help="Remove all the bells and whistles in CLI output (helpful for testing)"),
     debug: bool = typer.Option(False, "--debug", help="Use --debug to enable debugging output"),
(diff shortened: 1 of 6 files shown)


cpacker:
@sarahwooders maybe we should put this behind a `memgpt configure --advanced`?

Another potential issue here is the presets refactor, but I'm probably overthinking it and it's OK to do system prompt select that's not tied to function selection (at least for now).

memgpt/cli/cli.py
```diff
@@ -38,6 +38,7 @@ def run(
     human: str = typer.Option(None, help="Specify human"),
     model: str = typer.Option(None, help="Specify the LLM model"),
     preset: str = typer.Option(None, help="Specify preset"),
+    prompt: str = typer.Option(None, help="Specify prompt"),
```
cpacker, line 41: This should probably say "Specify system prompt"

memgpt/cli/cli.py
```diff
@@ -38,6 +38,7 @@ def run(
     human: str = typer.Option(None, help="Specify human"),
     model: str = typer.Option(None, help="Specify the LLM model"),
     preset: str = typer.Option(None, help="Specify preset"),
```
cpacker, line 40: @sarahwooders example here - technically "preset" = "system prompt" + "function set"

So there's a slight conflict allowing people to also specify "system prompt", but maybe it's OK? Just overwrite the implicit choice of system prompt in preset?

memgpt/cli/cli.py
Drake-AI: Done.

memgpt/cli/cli.py
Drake-AI: Maybe it is better to allow users to make their custom presets with system prompt and function set? But it is out of the scope of this PR.

Drake-AI:
Solved in this PR: https://github.com/cpacker/MemGPT/pull/420

Issue #426, opened 2023-11-11
MemGPTAgent attach error: LocalStorageConnector.insert_many() got an unexpected keyword argument 'show_progress'


anjaleeps:
**Describe the bug**
I get the following error when trying to attach a document to a MemGPTAgent with local storage in version 0.2.1.
```
LLM is explicitly disabled. Using MockLLM.
LLM is explicitly disabled. Using MockLLM.
LLM is explicitly disabled. Using MockLLM.
LLM is explicitly disabled. Using MockLLM.
Ingesting 16 passages into agent_146
  0%|          | 0/1 [00:00<?, ?it/s]
  0%|          | 0/1 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "C:\Users\Code\demo\main.py", line 47, in <module>
    inventory_agent.attach("inventory")
  File "C:\Users\**\AppData\Local\miniconda3\envs\genai\Lib\site-packages\memgpt\autogen\memgpt_agent.py", line 177, in attach
    attach(self.agent.config.name, data_source)
  File "C:\Users\**\AppData\Local\miniconda3\envs\genai\Lib\site-packages\memgpt\cli\cli.py", line 206, in attach
    dest_storage.insert_many(passages, show_progress=False)
TypeError: LocalStorageConnector.insert_many() got an unexpected keyword argument 'show_progress'
  0%|          | 0/1 [00:00<?, ?it/s]

```
**To Reproduce**
I get this error when running trying to attach a document to the MemGPTAgent through `attach`
```
inventory_agent = create_memgpt_autogen_agent_from_config(
    "inventory_agent",
    llm_config=llm_config,
    system_message=inventory_agent_prompts.system_message,
    interface_kwargs=interface_kwargs
)
inventory_agent.attach("inventory")
```

alexandj90:
C:\Users\<USERNAME>\miniconda3\Lib\site-packages\memgpt\cli cli.py

dest_storage.insert_many(passages, show_progress=False)      throws error

dest_storage.insert_many(passages)   Fixes it for me

cpacker:
Thank you for the report!

Patched in #432, will be on the next pypi update coming ASAP.

PR #427, opened 2023-11-11
Fix MemGPTAgent attach docs error

anjaleeps:
A potential fix to the issue [#426 ](https://github.com/cpacker/MemGPT/issues/426)


sarahwooders (approved):
Lgtm! We should eventually just implement the `insert_many(..., show_progress)` for local.py (this error happens because its only implemented for db.py)

PR #422, opened 2023-11-10
Refactor config + determine LLM via `config.model_endpoint_type`

sarahwooders:
* Closes #419
* Adds `create_time` and `memgpt_version` fields to agent config 
* Store model/embedding endpoints in config file 

Related to #301

---

TODOs (@cpacker)

- [x] Make wrapper choice done via selection, not raw input
- [x] Make sure that `--model`, `--wrapper`, `--endpoint`, `--endpoint_type` are all overwriteable at runtime (`memgpt run`)
  - This allows users to toggle local backends easily with the same agent
  - Needs fix, see: https://github.com/cpacker/MemGPT/pull/422#issuecomment-1811103833
  - Fixed by commit https://github.com/cpacker/MemGPT/pull/422/commits/8e9d42ac5c9a25b23b31bfc1442db121f7a245f7

image

---

**Please describe the purpose of this pull request.**
Various issues/confusion due to enviornment variables used to configure models 

**How to test**
Run `memgpt configure` and `memgpt run`

**Have you tested this PR?**

- [x] Test Azure 
- [x] Test local LLMs 
  - [x] ollama 
  - [x] openai proxy 
  - [x] lmstudio 
  - [x] llamacpp
  - [x] koboldcpp
  - [x] webui
- [ ] Test `memgpt configure` with an old `~/.memgpt/config` file (backwards compatibility) 
- [ ] Test `memgpt run` with an old agent file (backwards compatibility) 

**Related issues or PRs**
#419

PR #432, opened 2023-11-12
[fix] remove asserts for `OPENAI_API_BASE`

sarahwooders:


cpacker (approved):
Patch LGTM, will be properly fixed via #422

PR #420, opened 2023-11-10
Configurable presets to support easy extension of MemGPT's function set

cpacker:
## New workflow for presets

Users can create their own preset YAML files and write their own function code to allow them to further customize MemGPT.

### Step 1: User creates the python functions that support this

Has to match a certain function signature (start with `self`) + has to have docstrings + typing. They can put this is a new file called `internet.py`, and as long as they drop/create it in the right location it will be registered when MemGPT starts:
```python
def http_request(self, method: str, url: str, payload_json: Optional[str] = None):
    """
    Generates an HTTP request and returns the response.

    Args:
        method (str): The HTTP method (e.g., 'GET', 'POST').
        url (str): The URL for the request.
        payload_json (Optional[str]): A JSON string representing the request payload.

    Returns:
        dict: The response from the HTTP request.
    """
    try:
        headers = {"Content-Type": "application/json"}

        # For GET requests, ignore the payload
        if method.upper() == "GET":
            print(f"[HTTP] launching GET request to {url}")
            response = requests.get(url, headers=headers)
        else:
            # Validate and convert the payload for other types of requests
            if payload_json:
                payload = json.loads(payload_json)
            else:
                payload = {}
            print(f"[HTTP] launching {method} request to {url}, payload=\n{json.dumps(payload, indent=2)}")
            response = requests.request(method, url, json=payload, headers=headers)

        return {"status_code": response.status_code, "headers": dict(response.headers), "body": response.text}
    except Exception as e:
        return {"error": str(e)}
```

### Step 2: user creates a new preset (YAML file) that uses this function

Then, the user can create a new preset that uses this function (let's say it's called `memgpt_internet.yaml`:
```yaml
system_prompt: "memgpt_chat"
functions:
  - "send_message"
  - "pause_heartbeats"
  - "core_memory_append"
  - "core_memory_replace"
  - "conversation_search"
  - "conversation_search_date"
  - "archival_memory_insert"
  - "archival_memory_search"
  # internet access
  - "http_request"
```

Now, this preset will show up in the option menus as `memgpt_internet`.

---

### TODOs

- [x] Move base and extras function sets off of the agent class, link dynamically instead based on presets
- [x] Update poetry.lock with `docstring_parser`
- [x] Pull preset `.yaml` and function `.py` files from `~/.memgpt` to allow users to easily add their own
- [x] Make sure users can also add their own system prompts into `~/.memgpt/system_prompts`
- [ ] Update docs (add page about how to add new functionality to the agent)

### Unanswered questions

[...]

PR #438, opened 2023-11-13
[fix] bug in agent creation when `preset="memgpt_extras"`

sarahwooders:


vivi (approved):

cpacker (approved):
lgtm, will get overwritten by #420

Issue #446, opened 2023-11-14
Add ignores for popular OS's, IDE's and Python which is more complete for unessential files.


agiletechnologist:
**Describe the bug**
A more complete .gitignore which includes popular OS, IDE in addition to Python excludes.

- Python,
- OS (Linux, Windows, MacOS)
- IDE (Visual Studio, Visual Studio Code, PyCharm)

PR #447, opened 2023-11-14
added ignores for OS, IDE, and Python that were missing. (#446)

agiletechnologist:
This is a fix for https://github.com/cpacker/MemGPT/issues/446

**How to test**
This is for ignoring git commits for Python, OS, and IDE's

**Have you tested this PR?**
Yes. There are some files to be removed but will not be ignored until removed.  Will discuss the removal. 

Yes but one file. makes more sense as a single pull.

### DIFF
--- a/memgpt/memory.py
+++ b/memgpt/memory.py
@@ -7,6 +7,8 @@ from memgpt.constants import MESSAGE_SUMMARY_WARNING_FRAC
 from memgpt.utils import get_local_time, printd, count_tokens
 from memgpt.prompts.gpt_summarize import SYSTEM as SUMMARY_PROMPT_SYSTEM
 from memgpt.openai_tools import create
+from memgpt.config import MemGPTConfig
+from memgpt.embeddings import embedding_model
 from llama_index import Document
 from llama_index.node_parser import SimpleNodeParser
 from llama_index.node_parser import SimpleNodeParser


cpacker (approved):
lgtm

memgpt/constants.py
```diff
@@ -1,6 +1,6 @@
 import os
 
-MEMGPT_DIR = os.path.join(os.path.expanduser("~"), ".memgpt")
+MEMGPT_DIR = os.getenv(os.path.join("MEMGPT_CONFIG_PATH",".memgpt"), default=os.path.join(os.path.expanduser("~"), ".memgpt"))
```
sarahwooders, line 3: I think `MEMGPT_CONFIG_PATH` is currently the actual path of the config? so `~/.memgpt/config`, rather than the directory. 

So this should maybe be: 

```
MEMGPT_DIR = os.getenv("MEMGPT_CONFIG_PATH", default=os.path.join(os.path.expanduser("~"), ".memgpt"))
```

agiletechnologist:
This is the corrected form.

MEMGPT_DIR = os.getenv(os.path.join("MEMGPT_CONFIG_PATH",
default=os.path.join(os.path.expanduser("~"), ".memgpt"))

memgpt/constants.py
agiletechnologist: The MEMGPT_DIR was not supposed to be included only the one for .gitignore.

sarahwooders:
@agiletechnologist could you remove the `constants.py` diff? Then we can merge this.

Issue #451, opened 2023-11-15
model_endpoint does not get set when OPENAI_API_BASE is set and openai is the backend


kfsone:
memgpt/cli/cli_config.py
```
        default_model_endpoint = os.getenv("OPENAI_API_BASE")
        # if OPENAI_API_BASE is not set, try to pull a default IP+port format from a hardcoded set
        if default_model_endpoint is None:
            if model_endpoint_type in DEFAULT_ENDPOINTS:
                default_model_endpoint = DEFAULT_ENDPOINTS[model_endpoint_type]
                model_endpoint = questionary.text("Enter default endpoint:", default=default_model_endpoint).ask()
            else:
                # default_model_endpoint = None
                model_endpoint = None
                while not model_endpoint:
                    model_endpoint = questionary.text("Enter default endpoint:").ask()
                    if "http://" not in model_endpoint and "https://" not in model_endpoint:
                        typer.secho(f"Endpoint must be a valid address", fg=typer.colors.YELLOW)
                        model_endpoint = None
        assert model_endpoint, f"Environment variable OPENAI_API_BASE must be set."
```
so model_endpoint doesn't get set

cpacker:
Thank you for catching this! PR merged

PR #452, opened 2023-11-15
When default_mode_endpoint has a value, it needs to become model_endp…

kfsone:
When trying to use MemGPT with a litellm container trying to run 'memgpt configure' was hitting an assert complaining that model_endpoint was unset, the reason was that the default_model_endpoint was set and there was no branch of the final if block before the assert to handle this, so model_endpoint never got set.


kfsone:
https://github.com/cpacker/MemGPT/issues/451

cpacker (approved):
Thanks for catching this! Great PR

Issue #458, opened 2023-11-15
Could not load OpenAI model. If you intended to use OpenAI, please check your OPENAI_API_KEY in spite of local


ErfolgreichCharismatisch:
**Describe the bug**
I set it all for local but when trying to run memgpt load directory --name BearbedBirb --input-dir "E:\Birbs" --recursive I got

```
Could not load OpenAI model. If you intended to use OpenAI, please check your OPENAI_API_KEY.
Original error:
No API key found for OpenAI.
Please set either the OPENAI_API_KEY environment variable or openai.api_key prior to initialization.
API keys can be found or created at https://platform.openai.com/account/api-keys

To disable the LLM entirely, set llm=None.
```

**Your setup (please complete the following information)**

- Your OS (Linux, MacOS, Windows): Windows 10
- Where you're trying to run MemGPT from
  - Terminal? PowerShell? cmd.exe? cmd
- Your python version (run `python --version`) 3.10
-  If you installed with `pip install pymemgpt`: Installed with pip

**Additional context**
Full output:

loading data
done loading data
LLM is explicitly disabled. Using MockLLM.
┌─────────────────────────────── Traceback (most recent call last) ────────────────────────────────┐
│ C:\Logs\Anaconda\envs\p310\lib\site-packages\llama_index\llms\utils.py:20 in resolve_llm         │
│                                                                                                  │
│   17 │   if llm == "default":                                                                    │
│   18 │   │   # return default OpenAI model. If it fails, return LlamaCPP                         │
│   19 │   │   try:                                                                                │
│ > 20 │   │   │   llm = OpenAI()                                                                  │
│   21 │   │   except ValueError as e:                                                             │
│   22 │   │   │   raise ValueError(                                                               │
│   23 │   │   │   │   "******\n"                                                                  │
│                                                                                                  │
│ C:\Logs\Anaconda\envs\p310\lib\site-packages\llama_index\llms\openai.py:92 in __init__           │
│                                                                                                  │
│    89 │   ) -> None:                                                                             │
│    90 │   │   additional_kwargs = additional_kwargs or {}                                        │
│    91 │   │                                                                                      │
│ >  92 │   │   api_key, api_type, api_base, api_version = resolve_openai_credentials(             │
│    93 │   │   │   api_key=api_key,                                                               │
│    94 │   │   │   api_type=api_type,                                                             │
│    95 │   │   │   api_base=api_base,                                                             │
│                                                                                                  │
[...]

PR #462, opened 2023-11-15
Set service context for llama index in `local.py` 

sarahwooders:
Minor fix for #458


memgpt/cli/cli_load.py
```diff
@@ -27,6 +27,7 @@
 
 def store_docs(name, docs, show_progress=True):
     """Common function for embedding and storing documents"""
+
```
vivi, line 30: Stray newline?

vivi (approved):

PR #496, opened 2023-11-21
feat: metadata in archival memory

JohnCHarrington:
Adds metadata as an option in archival memory. This means that if the connected store that provides the archival memory stores metadata then MemGPT can use it.

As none of the built-in storage options provide metadata (as far as I know), this shouldn't make any difference to existing storage options. I am using a custom storage connector using Milvus and MongoDB that provides document info as metadata, allowing MemGPT to make use of things like the document title, last updated date etc.

Let me know your thoughts!

### DIFF
--- a/memgpt/functions/function_sets/base.py
+++ b/memgpt/functions/function_sets/base.py
@@ -181,6 +181,6 @@ def archival_memory_search(self, query: str, page: Optional[int] = 0) -> Optiona
         results_str = f"No results found."
     else:
         results_pref = f"Showing {len(results)} of {total} results (page {page}/{num_pages}):"
-        results_formatted = [f"timestamp: {d['timestamp']}, memory: {d['content']}" for d in results]
+        results_formatted = [f"timestamp: {d['timestamp']}, memory: {d['content']}, metadata: {d['metadata']}" for d in results]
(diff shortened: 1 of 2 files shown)


sarahwooders:
@JohnCHarrington would you still be interested in contributing this?  It would be great to add metadata to the `Passage` object (the representation for archival memory).

JohnCHarrington:
Absolutely, I can have another look at it next week and clear up conflicts etc.

JohnCHarrington:
It appears that the `Passage` class now has a metadata attribute, so I have removed the part of this change that added that. I have also added the extraction of metadata from `SimpleNodeParser` when calling .insert on the `EmbeddingArchivalMemory`.

JohnCHarrington:
@sarahwooders are there any other changes that should be made in other areas to ensure full support for this?

sarahwooders:
@JohnCHarrington Sorry missed this update. I think this looks good to merge - we should also think about the developer API for adding custom metadata when loading data (maybe instructions for adding custom loaders?) but we can leave that as a future todo. 

If you could fix the formatting checks, I can approve and merge this!

sarahwooders:
Closing since stale, but please feel free to reach out if you want to continue working on this.

PR #503, opened 2023-11-24
Fix class names used in persistence manager logging

claucambra:
All of the print statements used at creation time of the different persistence managers use the same name for each class. This PR changes the names to actually match the classes they are being used in

**How to test**
Run MemGPT with different persistence managers

**Have you tested this PR?**
```
Created new agent agent_111.
Available functions:
 ['send_message', 'pause_heartbeats', 'core_memory_append', 'core_memory_replace', 'conversation_search', 'conversation_search_date', 'archival_memory_insert', 'archival_memory_search']
AgentAsync initialized, self.messages_total=3
Initializing LoadStateManager with agent object
InMemoryStateManager.all_messages.len = 4
InMemoryStateManager.messages.len = 4
💭 *inner thoughts* Ah, the user has logged in for the first time. How exciting! I should introduce myself and make them feel welcome.
```

### DIFF
--- a/memgpt/persistence_manager.py
+++ b/memgpt/persistence_manager.py
@@ -146,7 +146,7 @@ class LocalStateManager(PersistenceManager):
             printd(f"Saved state to {fh}")
 
     def init(self, agent):
-        printd(f"Initializing InMemoryStateManager with agent object")
+        printd(f"Initializing LocalStateManager with agent object")
         self.all_messages = [{"timestamp": get_local_time(), "message": msg} for msg in agent.messages.copy()]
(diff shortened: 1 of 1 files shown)


cpacker (changes requested):
Thanks for catching this. Instead of hardcoding the name, can we instead pull from `self.__class__.__name__`?

claucambra:
> Thanks for catching this. Instead of hardcoding the name, can we instead pull from `self.__class__.__name__`?

Thanks for your review, I've used `self.__class__.__name__` now (and also fixed the class name issue in other logging lines here)

cpacker (approved):

PR #534, opened 2023-11-30
Remove `openai` package

### LINKED ISSUE #527: openai version required for local llm + autogen
llama-index 0.8.62 requires openai <1
pymemgpt 0.2.4 requires openai<0.29.0
pyautogen 0.2.0 requires openai~=1.2

Cannot resolve dependency conflicts

cpacker:
## Please describe the purpose of this pull request

OpenAI python client versioning broke backwards compatibility with <v1. Since our use of the python client was pretty isolated we're better off stripping out the package and just using HTTP requests (eg via python `requests`). This way we also will not run into versioning issues with popular integrations such as AutoGen, which use the python `openai` package.

For background / discussion on the backwards compat breaking v1 migration see: https://github.com/openai/openai-python/discussions/742

## How to test

NOTE: requires `pip install python-box`!

This is to allow backwards compatibility with "dot" notation (eg `dict.property` instead of `dict['property']`) access on objects returned from `openai`. We were already doing something like this in the local LLM code where we have an adhoc `DotDict` class, but our own version doesn't support nesting properly and `python-box` seems like a trusted option for this:

https://github.com/cpacker/MemGPT/blob/main/memgpt/local_llm/chat_completion_proxy.py#L121-L143

Testing basic support (ie run a fresh agent, make sure the calls to LLM/embeddings backend work):

- [x] OpenAI API
- [x] OpenAI embeddings API
- [x] Azure API
- [x] Azure embeddings API
- [x] Local LLM

Note: killing legacy CLI in separate parallel PR, no need to test

~~- [ ] OpenAI API **(w/ legacy mode on MemGPT CLI)**~~
~~- [ ] OpenAI embeddings API **(w/ legacy mode on MemGPT CLI)**~~
~~- [ ] Azure API **(w/ legacy mode on MemGPT CLI)**~~
~~- [ ] Azure embeddings API **(w/ legacy mode on MemGPT CLI)**~~

Testing other parts of `openai` -> `requests` refactor

- [x] Test that rate-limiting refactor works (we use HTTP status codes instead of python `openai.errors` objects)
- [x] Test that context length overflow works (we use string matching for this right now)
- [x] Test local LLMs with `DotDict` -> `Box` refactor

### Example OpenAI HTTP responses for reference

Request size OK, but size exceeded during generation:
```python
{
'id': 'chatcmpl-...',
'object': 'chat.completion',
'created': '...',
'model': 'gpt-4-0613',
'choices': [
  {
    'index': 0, 'message': {'role': 'assistant', 'content': "Chad is still actively engaging with the story. With 
Lucas recovering from an accident and returning to the track, it's time to begin wrapping up the narrative. Lucas finally achieves his initial goal. Sixth paragraph: The final victory, Lucas's achievement and transformation. This is the moment of"}, 
  'finish_reason': 'length'
    }
],
'usage': {
  'prompt_tokens': 8135,
  'completion_tokens': 58,
  'total_tokens': 8193
},
'system_fingerprint': None
}
```

Request size exceeds limit:
```python
# HTTP error code 400
{
'message': "This model's maximum context length is 8192 tokens. However, your messages resulted in 9901 tokens (8871 in the messages, 1030 in the functions). Please reduce the length of the messages or functions.",
'type': 'invalid_request_error',
'param': 'messages',
'code': 'context_length_exceeded'
}
```

## Have you tested this PR?

See above checklist. For OpenAI API and OpenAI Embeddings API, see screenshots:

## Related issues or PRs

Related to many of the AutoGen issues.

Closes #527

### DIFF
--- a/memgpt/openai_tools.py
+++ b/memgpt/openai_tools.py
@@ -15,6 +15,47 @@ HOST_TYPE = os.getenv("BACKEND_TYPE")  # default None == ChatCompletion
 R = TypeVar("R")
 
 
+def is_context_overflow_error(exception):
+    from memgpt.utils import printd
+
+    match_string = "maximum context length"
+
+    # Backwards compatability with openai python package/client v0.28 (pre-v1 client migration)
+    if match_string in str(exception):
+        printd(f"Found '{match_string}' in str(exception)={(str(exception))}")
+        return True
+
+    # Based on python requests + OpenAI REST API (/v1)
+    elif isinstance(exception, requests.exceptions.HTTPError):
+        if exception.response is not None and "application/json" in exception.response.headers.get("Content-Type", ""):
+            try:
+                error_details = exception.response.json()
+                if "error" not in error_details:
+                    printd(f"HTTPError occured, but couldn't find error field: {error_details}")
+                    return False
+                else:
+                    error_details = error_details["error"]
+
+                # Check for the specific error code
+                if error_details.get("code") == "context_length_exceeded":
+                    printd(f"HTTPError occured, caught error code {error_details.get('code')}")
+                    return True
+                # Soft-check for "maximum context length" inside of the message
+                elif error_details.get("message") and "maximum context length" in error_details.get("message"):
+                    printd(f"HTTPError occured, found '{match_string}' in error message contents ({error_details})")
+                    return True
+                else:
+                    printd(f"HTTPError occured, but unknown error message: {error_details}")
+                    return False
+            except ValueError:
+                # JSON decoding failed
+                printd(f"HTTPError occurred ({exception}), but no JSON error message.")
+
+    # Generic fail
+    else:
+        return False
+
+
 def smart_urljoin(base_url, relative_url):
     """urljoin is stupid and wants a trailing / at the end of the endpoint address, or it will chop the suffix off"""
     if not base_url.endswith("/"):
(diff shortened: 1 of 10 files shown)


cpacker:
@sarahwooders for some reason my embeddings aren't getting sent through openai:

```
(pymemgpt-py3.10) loaner@MacBook-Pro-5 MemGPT-2 % cat ~/.memgpt/config
[defaults]
preset = memgpt_extras
persona = sam_pov
human = cs_phd

[model]
model = gpt-4-1106-preview
model_endpoint = https://api.openai.com/v1
model_endpoint_type = openai
context_window = 128000

[openai]
key = sk-...

[embedding]
embedding_endpoint_type = openai
embedding_endpoint = https://api.openai.com/v1
embedding_dim = 1536
embedding_chunk_size = 300

[archival_storage]
type = local

[version]
memgpt_version = 0.2.4

[client]
anon_clientid = 00000000000000000000a61b692e9d3d
```

image

memgpt/openai_tools.py
```diff
@@ -95,32 +207,55 @@ def chat_completion_with_backoff(
...
+        # return openai.ChatCompletion.create(
+        #     model=model,
+        #     messages=messages,
+        #     engine=engine,
+        #     deployment_id=deployment_id,
+        #     functions=functions,
+        #     function_call=function_call,
+        #     user=client_id,
+        # )
+        return azure_openai_chat_completions_request(
```
cpacker, line 248: @sarahwooders This is the Azure part I'm not sure about

The Azure docs only have a few variables they want when constructing the url for the POST request, and a few of the variables I'm seeing here I don't recognize.

cpacker: https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#embeddings

cpacker:
@sarahwooders Azure is now working for ChatCompletions, but not for embeddings - it seems like something is wrong with llamaindex maybe?

image

image

sarahwooders:
Are you setting the azure embedding deployment environment variable? I think it's otherwise None (not sure if we have access to an embedding deployment)

cpacker:
> Are you setting the azure embedding deployment environment variable? I think it's otherwise None (not sure if we have access to an embedding deployment)

Yep that was it. I must not have had it set because it wasn't present in `config.json` (it also wasn't present if I set it, then ran `memgpt run` but no `memgpt configure`).

I added a fallback for when it's not specified and Azure embeddings are working now:

image

sarahwooders (approved):
lgtm!

PR #619, opened 2023-12-15
feat: Add converter to support outbound calls to OpenAI servers that want `tool` not `function`

cpacker:

The function naming (`function` / `function_call` / `function` role) is deprecated in OpenAI API and partially (?) deprecated in Azure OpenAI. OpenAI and Azure both still support the old naming methodology, but newer "OpenAI API-compatible" services that also allow function calling (eg Anyscale) don't support the old naming scheme and want `tool` only.

This PR adds a converter function that goes from the old format to the new one. This is done on-the-fly, ie we still store messages in the functions style but we modify it on the way out to the OpenAI API server.

@sarahwooders eventually we probably want to migrate the internal representations to the `tool_call` format, but I'm not sure how important it is for now, since unless we want to have parallel function calling support nothing really changes and these light patches work fine (AFAIK).

**How to test**

Try using MemGPT with a non-OpenAI OpenAI-compatible endpoint that has function calling, but only the `tool` version of function calling (eg Anyscale).

**Have you tested this PR?**

Yes, seems to work OK with Anyscale.

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -423,11 +423,14 @@ class Agent(object):
         response_message = response.choices[0].message
 
         # First message should be a call to send_message with a non-empty content
-        if require_send_message and not response_message.get("function_call"):
+        if require_send_message and not (response_message.get("function_call") or response_message.get("tool_calls")):
             printd(f"First message didn't include function call: {response_message}")
             return False
 
         function_call = response_message.get("function_call")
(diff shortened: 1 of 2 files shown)


cpacker:
Force push rebased on 0.2.11

cpacker:
Deprecated by 0.3.0

PR #625, opened 2023-12-15
feat: adding first poc of web UI

goetzrobin:
Same changes as https://github.com/cpacker/MemGPT/pull/615 but with clean commit history & --server flag removed.

**Please describe the purpose of this pull request.**
Is it to add a new feature? Is it to fix a bug?

A new feature adding a Web-Based UI that allows users to interact with MemGPT outside of the CLI.

**How to test**
How can we test your PR during review? What commands should we run? What outcomes should we expect?

Please give it a try and see if you can get it up and running with the following steps:
1. run `memgpt configure`
2. run `memgpt server`
3. go to `http://127.0.0.1:8283` and see if you can create an agent (prefilled with data that works, need to add endpoints to fetch valid options) and chat with them

**Have you tested this PR?**
Have you tested the latest commit on the PR? If so please provide outputs from your tests.

I have tested it and was able to create an Agent and Chat with the agent

**Related issues or PRs**
Please link any related GitHub [issues](https://github.com/cpacker/MemGPT/issues) or [PRs](https://github.com/cpacker/MemGPT/pulls).

**Is your PR over 500 lines of code?**
If so, please break up your PR into multiple smaller PRs so that we can review them quickly, or provide justification for its length.

Justification is that initial setup was necessary to get up and running. Smaller PRs going forward!

**Additional context**
Add any other context or screenshots about the PR here.

PR #627, opened 2023-12-15
feat: Chat UI feature branch

cpacker:
# MemGPT Chat UI

- Building on the initial PR by @goetzrobin (#625)
- Should merge into main once stable
- When coding up features for the chat UI, base your PRs against this feature branch (`chatui`)

## How to run

1. run `memgpt configure`
2. run `memgpt server`
3. go to `http://127.0.0.1:8283` and see if you can create an agent (prefilled with data that works, need to add endpoints to fetch valid options) and chat with them

## API documentation

API docs are available here: https://memgpt.readme.io/reference

## TODOs

Highest priority marked with ⚠️

🎄 Pending

- [ ] **[Pure UI]** Display current agent name somewhere in the chat view (probably top middle?)
- [ ] **[Pure UI]** An "agent search" box that filters the grid view based off of name
- [ ] **[Pure UI]** Have a "update successful" indicator / model / toast when a memory update goes through properly
- [x] ⚠️ **[Pure UI]** On an exception / failed message sent: the UI should indicate that the message failed to send (similar to FB messenger etc, red text or outline indicating failed message) + propagate error message to the UI (in some capacity)
  - More details here: https://github.com/cpacker/MemGPT/pull/642#issuecomment-1861291855
- [x] ⚠️ **[Pure UI]** On an exception / failed message sent: "typing..." lock should be released (currently gets trapped on "typing..." state
- [ ] **[Pure UI]** Add syntax highlighting to function call views
- [ ] ⚠️ **[UI + backend]** When a chat window is opened, populate it with the last N messages
  - we can start by just displaying the messages in the agent buffer `agent.messages`
  - eventually add a scroll-back fetch based on a paginated query to recall memory (messages from recall should be styled differently in the UI to indicate that they are out-of-context)
- [ ] ⚠️ **[UI + backend]** Enable viewing archival memory fragments in a modal (eg called "archival memory")
- [ ] **[UI + backend]** Allow modifying agent configs + base config via settings page
- [ ] **[UI + backend]** Allow attaching data (`/attach`) via a GUI / modal
- [ ] **[UI + backend]** Allow viewing available data sources (post `/attach`) via a GUI / modal
- [ ] **[backend]** REST requests for retrieving data in archival
- [ ] **[backend]** REST requests for retrieving data in recall

✅ Done

- [x] **[Pure UI]** Add thinking bubble when agent is typing / response is streaming
- [x] **[Pure UI]** Add "memory" viewer component (can view agent's core memories in pop-up modal in text fields w/ GET, and edit them then click save which sends a PUT)
- [x] **[Pure UI]** Dark mode
- [x] **[Pure UI]** Display function calls (should show everything except `send_message` by default)
- [x] **[Pure UI]** Make landing page agent list an N x M grid (no overflow horizontal scrolling, just vertical scrolling)
- [x] **[Pure UI]** Make "add new" always visible (should be left / top left in grid ordering)
- [x] **[Pure UI]** On create agent click, should kill modal to indicate success / failure
- [x] **[Pure UI]** Make size/width of inner thoughts + message similar
- [x] **[backend]** Chat UI server should use `host` and `port` arg coming via the server command
- [x] **[backend]** Chat UI server needs to shut down clean on CTRL-C (hanging process?)

🎁 Pushed to later PR? (might not be worth trying to include for a shippable v1)

```
- [ ] **[UI + backend]** Allow viewing custom function files in the GUI (`~/.memgpt/functions`)
- [ ] **[UI + backend]** Allow viewing custom prompt settings in the GUI (`~/.memgpt/settings`)
```

📦 Final tasks (cleanup)

```
- [ ] General stability w/ expected single-user usage (users should not see locking errors etc)
- [ ] Make sure all underlying MemGPT errors (eg `LLMJsonError`) are being propagated to the UI (no silent errors)
- [ ] Test on Windows + Linux + MacOS via pypi test package
- [ ] Clean, pleasant, delightful UI (agent-centric)
```

---

For function call views, we should probably have a "code viewer" / markdown view, similar to what's used in chat-llama-index (see screenshot from https://github.com/cpacker/MemGPT/pull/474). Function calls can then just be visualized in this code viewer, similar to how it's being done in the stale #474 branch:

image


goetzrobin:
@cpacker I am back from break and was looking at the next priorities to tackle. What would you say the main priorities are right now?

cpacker:
> @cpacker I am back from break and was looking at the next priorities to tackle. What would you say the main priorities are right now?

I'd say the current ⚠️ marking is pretty accurate - imo highest priority is probably "When a chat window is opened, populate it with the last N messages"?

Storage refactor is still going through so there's no paginated query from recall yet but I should be able to add that in the next day or so. So for the time being the population can just come from the inner context (agent.messages).

cpacker:
Made redundant by #821

TODOs moved here: #826

cpacker:
Made redundant by #821

TODOs moved here: #826

PR #630, opened 2023-12-16
feat: pull model list for openai-compatible endpoints

### LINKED ISSUE #592: List models options with `/models` for OpenAI compatible endpoints
Currently this is hard-coded or has to be manually specified. We can generate the dropdown based off the available models the endpoint lists. 

Relevant for: 
* vLLM 
* OpenAI

cpacker:
Closes #592

- During `memgpt configure` if using an openai endpoint, pull the list of models
- Also, allow users to choose a "enter yourself" custom option (useful for custom OpenAI compatible endpoints)

**How to test**

Run `memgpt configure`

OpenAI
image

Azure
image

vLLM
image

**Have you tested this PR?**

- [x] Tested on OpenAI
- [x] Tested on Azure OpenAI
- [x] Tested on vLLM

**Other comments**

@sarahwooders

- I'm concerned for some users the total amount of models is information overload (see screenshot).
- Also, this is even after filtering for models only prefixed by `gpt-`
- Additionally, there are models in the list that are misleading to show (eg `gpt-4-vision-preview` shows up, but we don't support vision inputs at the moment)


sarahwooders (changes requested):
Could you also call the same function for the case where `vllm` is the endpoint type? e.g. with our endpoint you can do `http://api.memgpt.ai/v1/models` (since we require the OpenAI compatible vLLM endpoint). 

Also might be nice to add Azure too, but not that important.

memgpt/cli/cli_config.py
```diff
@@ -116,24 +133,51 @@ def configure_model(config: MemGPTConfig, model_endpoint_type: str):
...
+
         # vllm needs huggingface model tag
         if model_endpoint_type == "vllm":
-            default_model = config.model if config.model and config.model_endpoint_type == "vllm" else ""
-            model = questionary.text(
-                "Enter HuggingFace model tag (e.g. ehartford/dolphin-2.2.1-mistral-7b):",
-                default=default_model,
-            ).ask()
-            model = None if len(model) == 0 else model
-            model_wrapper = None  # no model wrapper for vLLM
```
cpacker, line 127: FYI removed this, not sure why we weren't asking for a wrapper w/ vLLM

Basically as long as the backend is not fully-OpenAI compatible (including function call compatible) then we should specify a wrapper (because in that case, we'll be doing our own prompt formatting then routing the request to a `/completions` API instead of a `/chat/completions` API.

memgpt/cli/cli_config.py
```diff
@@ -96,16 +97,32 @@ def configure_llm_endpoint(config: MemGPTConfig):
...
     # set: model, model_wrapper
     model, model_wrapper = None, None
     if model_endpoint_type == "openai" or model_endpoint_type == "azure":
-        model_options = ["gpt-4", "gpt-4-1106-preview", "gpt-3.5-turbo", "gpt-3.5-turbo-16k"]
-        # TODO: select
+        try:
+            if model_endpoint_type == "openai":
+                model_options = openai_get_model_list(url=model_endpoint, api_key=config.openai_key)
+            elif model_endpoint_type == "azure":
+                model_options = azure_openai_get_model_list(url=model_endpoint, api_key=config.azure_key, api_version=config.azure_version)
```
cpacker, line 108: This is potentially problematic because `config.azure_key` and `config.azure_version` may not have been defined at this point?

memgpt/cli/cli_config.py
```diff
@@ -116,24 +133,51 @@ def configure_model(config: MemGPTConfig, model_endpoint_type: str):
             ).ask()
             model = None if len(model) == 0 else model
 
+        default_model = config.model if config.model and config.model_endpoint_type == "vllm" else ""
+
         # vllm needs huggingface model tag
         if model_endpoint_type == "vllm":
-            default_model = config.model if config.model and config.model_endpoint_type == "vllm" else ""
```
cpacker, line 121: Also should this be `config.model_endpoint_type` or `model_endpoint_type`?

cpacker:
> Could you also call the same function for the case where `vllm` is the endpoint type? e.g. with our endpoint you can do `http://api.memgpt.ai/v1/models` (since we require the OpenAI compatible vLLM endpoint).
> 
> Also might be nice to add Azure too, but not that important.

Added vLLM support + tested against our vLLM API (see picture in OP)

memgpt/cli/cli_config.py
sarahwooders: I think this is right? its using `config.model` if config is already configured for `config.model_endpoint_type == "vllm"`.

memgpt/cli/cli_config.py
sarahwooders: I think this might error on the first run of memgpt configure, if you don't already have credentials put in?

cpacker: This PR should mean this won't happen since it forces keys to be specified up front: https://github.com/cpacker/MemGPT/pull/675

cpacker:
@sarahwooders added a new method to resolve the "too many models" problem:

Start by showing a curated list:
image

Allow the user to expand the list (if they expand the list, they get a warning):
image

Finally, they can still enter a manual model if the internet / GET request is broken and they know what they're doing:
image

sarahwooders (approved):
lgtm!

memgpt/cli/cli_config.py
cpacker: Actually it meant it for openai, but not for azure, which I patched here https://github.com/cpacker/MemGPT/pull/630/commits/4d38c14defc0dc5ab924f239f1450aca99dd67ec

PR #637, opened 2023-12-18
feat: next iteration of chat ui

goetzrobin:
Marking the following TODOS as completed:

> ## TODOs
> Highest priority marked with ⚠️
> 
> ### Pure UI
> * [x]  ⚠️ Display function calls (should show everything except `send_message` by default)
> * [x]  ⚠️ Make landing page agent list an N x M grid (no overflow horizontal scrolling, just vertical scrolling)
> * [ ]  ⚠️ Add thinking bubble when agent is typing / response is streaming
> * [ ]  ⚠️ Add "memory" viewer component (can view agent's core memories in pop-up modal in text fields w/ GET, and edit them then click save which sends a PUT)
> * [x]  Make "add new" always visible (should be left / top left in grid ordering)
> * [ ]  Dark mode
> * [x]  On create agent click, should kill modal to indicate success / failure
> * [x]  Make size/width of inner thoughts + message similar
> * [ ]  An "agent search" box that filters the grid view based off of name
> 
> For function call views, we should probably have a "code viewer" / markdown view, similar to what's used in chat-llama-index (see screenshot from #474). Function calls can then just be visualized in this code viewer, similar to how it's being done in the stale #474 branch:
> 
> ### UI + backend
> * [ ]  ⚠️ When a chat window is opened, populate it with the last N messages (we can start by just displaying the messages in the agent buffer `agent.messages`, and eventually add a scroll-back fetch based on a paginated query to recall memory)
> * [ ]  Allow modifying agent configs + base config via settings page
> 
> ### Backend
> * [x]  Chat UI server should use `host` and `port` arg coming via the server command
> * [x]  Chat UI server needs to shut down clean on CTRL-C (hanging process?)
> * [ ]  Any missing REST requests?

## Next work

I think the next low hanging fruit I can add tomorrow:
> * [ ]  ⚠️ Add thinking bubble when agent is typing / response is streaming
> * [ ]  Dark mode
> * [ ]  An "agent search" box that filters the grid view based off of name
 
Not as long hanging but still pretty straight forward (tomorrow or in a couple days)
> * [ ]  ⚠️ Add "memory" viewer component (can view agent's core memories in pop-up modal in text fields w/ GET, and edit them then click save which sends a PUT)
> * [ ]  ⚠️ When a chat window is opened, populate it with the last N messages (we can start by just displaying the messages in the agent buffer `agent.messages`, and eventually add a scroll-back fetch based on a paginated query to recall memory)
> * [ ]  Allow modifying agent configs + base config via settings page


cpacker:
image

LGTM!

cpacker (approved):

Issue #319, opened 2023-11-05
Implement Global Logging for MemGPT to Console and Log Files


agiletechnologist:
**Epic User Story**
As a developer or user of MemGPT, I would like to have a consistent way of logging to both the console and to a file based on configuration and command line options So that errors and debugging info are available on the console and in a log file when needed.

**Describe the solution you'd like**
Provide a more consistent global logging strategy. I suggest using Pythons logging.
https://docs.python.org/3/howto/logging.html 

Some other benefits are executing code only for a specific level:
```python
If logging.DEBUG >= logging.getEffectiveLevel():
    # do something here
```

**Describe alternatives you've considered**
I have yet to consider others, as no additional modules need to be installed. But for diversity, we can look at options discussed at https://betterstack.com/community/guides/logging/best-python-logging-libraries/

agiletechnologist:
I am going to work on this. Where would you like the core import to be stored? I suggest naming it logmemgpt.py.

- use Constants unless overridden by Environment for the location of logs in the filesystem with the following presidents:
1. Environment Variables
2. Config options if used. (not yet, but make plans for the future.)
3. Constants

- New Constants for defining Log format. Change the option on the command line for --debug to debug log level. I would like to see this option to change to --log-level with the following values allowed:
|Level|Numeric value|What it means / When to use it|
|logging.DEBUG | 10 | Detailed information, typically only of interest to a developer trying to diagnose a problem.|
|-- | -- | --|
|logging.INFO | 20 | Confirmation that things are working as expected.|
|logging.WARNING | 30 | An indication that something unexpected happened or a problem might occur soon (e.g., disk space low). The software is still working as expected.|
|logging.ERROR | 40 | Due to a more serious problem, the software cannot perform some functions.|
|logging.CRITICAL | 50 | A serious error indicating that the program itself may be unable to continue running.|

PR #644, opened 2023-12-18
feat: global logging

agiletechnologist:
New feature global logging core 

**How to test**
Run ```memgpt configure```. This will write config file changes that now include logging that is loaded automatically at launch
This will trigger config changes when logging configuration is not present as well with ```memgpt run```
At this point, the code needs to be modified to write out to the global logger 
at the top. You will need to include ```from memgpt.memgptlog import logger```

```python
from memgpt.memgptlog import logger
#example usage
logger.debug("Your Debug Message")
logger.info("Your Info Message")
```

**Have you tested this PR?**
Config loads into logger without error. Creates log in .memgpt/logs/memgpt.log

**Related issues or PRs**
https://github.com/cpacker/MemGPT/issues/319

No

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -411,12 +411,14 @@ class Agent(object):
             response_message["tool_call_id"] = tool_call_id
             # role: assistant (requesting tool call, set tool call ID)
             messages.append(response_message)  # extend conversation with assistant's reply
+            printd(f"Function call message: {messages[-1]}")
 
             # Step 3: call the function
             # Note: the JSON response may not always be valid; be sure to handle errors
 
             # Failure case 1: function name is wrong
             function_name = response_message["function_call"]["name"]
+            printd(f"Request to call function {function_name} with tool_call_id: {tool_call_id}")
             try:
                 function_to_call = self.functions_python[function_name]
             except KeyError as e:
--- a/memgpt/connectors/db.py
+++ b/memgpt/connectors/db.py
@@ -26,7 +26,7 @@ from memgpt.connectors.storage import StorageConnector, TableType
 from memgpt.config import AgentConfig, MemGPTConfig
 from memgpt.constants import MEMGPT_DIR
 from memgpt.utils import printd
-from memgpt.data_types import Record, Message, Passage, Source
+from memgpt.data_types import Record, Message, Passage, Source, ToolCall
 
 from datetime import datetime
 
@@ -69,6 +69,26 @@ class CommonVector(TypeDecorator):
     def process_result_value(self, value, dialect):
         list_value = ast.literal_eval(value)
         return np.array(list_value)
+
+
+class ToolCalls(TypeDecorator):
+
+    """Custom type for storing List[ToolCall] as JSON"""
+
+    impl = JSON
+
+    def load_dialect_impl(self, dialect):
+        return dialect.type_descriptor(JSON())
+
+    def process_bind_param(self, value, dialect):
+        if value:
+            return [vars(v) for v in value]
+        return value
+
+    def process_result_value(self, value, dialect):
+        if value:
+            return [ToolCall(**v) for v in value]
+        return value
 
 
 Base = declarative_base()
@@ -155,8 +175,7 @@ def get_db_model(table_name: str, table_type: TableType, dialect="postgresql"):
             # if role == "assistant", this MAY be specified
             # if role != "assistant", this must be null
             # TODO align with OpenAI spec of multiple tool calls
-            tool_name = Column(String)
-            tool_args = Column(String)
+            tool_calls = Column(ToolCalls)
 
             # tool call response info
             # if role == "tool", then this must be specified
@@ -185,8 +204,7 @@ def get_db_model(table_name: str, table_type: TableType, dialect="postgresql"):
                     user=self.user,
                     text=self.text,
                     model=self.model,
-                    tool_name=self.tool_name,
-                    tool_args=self.tool_args,
+                    tool_calls=self.tool_calls,
                     tool_call_id=self.tool_call_id,
                     embedding=self.embedding,
                     created_at=self.created_at,
--- a/memgpt/data_types.py
+++ b/memgpt/data_types.py
@@ -1,7 +1,7 @@
 """ This module contains the data types used by MemGPT. Each data type must include a function to create a DB model. """
 import uuid
 from abc import abstractmethod
-from typing import Optional
+from typing import Optional, List, Dict
 import numpy as np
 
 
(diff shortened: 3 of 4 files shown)


agiletechnologist:
failed on something I do not control

FAILED tests/test_websocket_interface.py::test_websockets - requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://api.openai.com/v1/chat/completions
============= 1 failed, 15 passed, 5 skipped, 2 warnings in 52.63s =============
Error: Process completed with exit code 1.

memgpt/config.py
```diff
@@ -192,10 +240,57 @@ def save(self):
             self.anon_clientid = self.generate_uuid()
         set_field(config, "client", "anon_clientid", self.anon_clientid)
 
-        if not os.path.exists(MEMGPT_DIR):
-            os.makedirs(MEMGPT_DIR, exist_ok=True)
+        # logging
```
cpacker, line 243: @sarahwooders wondering should we potential write these out to a different file called logging.config or something?

memgpt/config.py
```diff
@@ -213,7 +308,7 @@ def create_config_dir():
         if not os.path.exists(MEMGPT_DIR):
             os.makedirs(MEMGPT_DIR, exist_ok=True)
 
-        folders = ["personas", "humans", "archival", "agents", "functions", "system_prompts", "presets"]
+        folders = ["personas", "humans", "archival", "agents", "functions", "system_prompts", "presets", "logs"]
```
cpacker, line 311: @sarahwooders unrelatedly needs to be updated for the new `/settings` dir

memgpt/memgptlog.py
cpacker, line 1: @sarahwooders @agiletechnologist maybe it makes more sense to name this file `log` or `logger`? To follow the current `memgpt.<attribute>` naming conventions

memgpt/utils.py
```diff
@@ -259,3 +259,11 @@ def get_schema_diff(schema_a, schema_b):
     difference = [line for line in difference if line.startswith("+ ") or line.startswith("- ")]
 
     return "".join(difference)
+
+
+def fix_file_path(path):
```
cpacker, line 264: @agiletechnologist instead of using `fix_file_path`, I think we should be using the tools in `os.path` which should automatically use the correct native system path format.

In this case, it looks like this is only applied on top of an `os.path.join` - does something break if this isn't done additionally?

memgpt/config.py
agiletechnologist: @cpacker It is possible, but having multiple configuration files might create confusion. There are only a few config options. Best practices include having a single configuration file that works within the existing config. Is there a reason it should be separated?

memgpt/memgptlog.py
agiletechnologist: @cpacker I was trying to avoid naming conflicts with the system modules and the variables.  The filename of "log" probably works best if we need to rename it.

memgpt/utils.py
agiletechnologist: I must move that out of the utils and put it in the memgptlog.py. There is a bug related to config import on Windows when it is saved. 
There is a bug when loading the configuration for Windows paths with logging.

memgpt/config.py
```diff
@@ -84,6 +87,14 @@ class MemGPTConfig:
     # version (for backcompat)
     memgpt_version: str = None
 
+    # logging (for logger)
+    logging_level: str = "CRITICAL"  # default log level
```
sarahwooders, line 91: Could we move logging information into a `logging_constants.py` or the `constants.py` file? Since it'll be developers modifying these fields, and we'd like to keep the config file minimal.

sarahwooders (changes requested):
Thanks for this PR! I think the default logging configuration should be stored in code, and that the log level (e.g. DEBUG) should be set by the `--debug` flag at runtime. Since parsing the config is a lot of code and it also quite annoying to make changes to the config (e.g. to maintain backwards compatibility) we try to avoid adding new fields.

memgpt/config.py
sarahwooders: Sorry just saw this message - the `~/.memgpt/config` file is quite annoying to maintain, since every time we change it we have to figure out how we will handle old config versions. Also, we are planning to use the `MemGPTConfig` object as basically user settings for the MemGPT server, which wont need any logging information. Given that only developers will be making modifications to log settings, I think we should keep log settings outside of  the `MemGPTConfig` class.

agiletechnologist: @sarahwooders I can change it to code, but I will have to abandon this pull request and delete this branch. I will redo it in a new branch and pull the request. Or change to "logging.config" as the config file.  I hate having multiple configuration files for reasons like configuring Spring Boot for Java web applications. There are so many config files. Setting it up properly requires finding the correct file for the configuration option you need to change. It is still better to load from a configuration rather than change source files to change.  This is better for maintaining code stability and keeping setting out of the code.  This will prevent you from accidentally committing debug code to production.  This comes from my DevOps Engineering experience and accidentally dealing with bad releases from developers.

memgpt/memgptlog.py
agiletechnologist: this is done.

memgpt/config.py
agiletechnologist: @sarahwooders I can change it to code, but I will have to abandon this pull request and delete this branch. I will redo it in a new branch and pull the request. Or change to "logging.config" as the config file. 

I hate having multiple configuration files for reasons like configuring Spring Boot for Java web applications. There are so many config files. Setting it up properly requires finding the correct file for the configuration option you need to change.

It is still better to load from a configuration rather than change source files to change. This is better for maintaining code stability and keeping setting out of the code. This will prevent you from accidentally committing debug code to production. This comes from my DevOps Engineering experience and accidentally dealing with bad releases from developers.

agiletechnologist:
@sarahwooders @cpacker Writing code and making changes to code can cause code to be committed to production release with debugging. This is why using a Config file is best. I am speaking from my years of experience in DevOps Engineering. Also, with this logging, users may want to see the info level of logging only, not the debug level. There is more logging than debugging and critical. Users are asking for the info-level logs, too.

cpacker:
> @sarahwooders @cpacker Writing code and making changes to code can cause code to be committed to production release with debugging. This is why using a Config file is best. I am speaking from my years of experience in DevOps Engineering. Also, with this logging, users may want to see the info level of logging only, not the debug level. There is more logging than debugging and critical. Users are asking for the info-level logs, too.

I agree that a config file may work better that constants for this kind of thing, but unfortunately there's a big refactor happening that touches configs atm so it's your changes (which are great) are going to be blocked by these other PRs for an indefinite amount of time.

By putting the settings into a constants file for now, it'll allow us to merge your PR much more quickly, and then we can have a second PR that moves settings from constants into a config (either the same config that the agents are using that is going to be stored in a DB, or a config file that's separate for logging - latter sounds better to me, but regardless will be useful to merge a version of this PR that's constants beforehand).

agiletechnologist:
This will require a complete rewrite

PR #647, opened 2023-12-19
feat: Update REST API

cpacker:
To enable autogen'ing the API docs from the code:

- Made request and response types pydantic models
- Added docstrings


cpacker:
Basic regression testing looks OK

image
image

PR #661, opened 2023-12-20
fix: Handle list[str] in function schema

Daenks:
Allows the usage of list[str] as a parameter and return type in MemGPT functions

**How to test**
write a function that takes a list[str] and try and have MemGPT load it.

This works
```
def AI_command(self, command: str) -> str:
```

This Doesnt
```
def AI_command_batch(self, commands: list[str]) -> str:
```

error:
```
Error loading function set: Python type list[str] has no corresponding JSON schema type
```

**Have you tested this PR?**
Yes, after change a function with this prototype will load and run successfully:
```
def AI_command_batch(self, commands: list[str]) -> str:
```

**Related issues or PRs**
None

No

**Additional context**
There is probably a better way to do this to include lists of other standard types such as int. My method is basically building out a long if-else chain/switch statement.

### DIFF
--- a/memgpt/functions/schema_generator.py
+++ b/memgpt/functions/schema_generator.py
@@ -1,6 +1,6 @@
 import inspect
 import typing
-from typing import get_args
+from typing import get_args, get_origin
 
 from docstring_parser import parse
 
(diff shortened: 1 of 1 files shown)


cpacker (approved):
LGTM - sorry for the delay!

Schedule to merge in post #785, #779, tagging priority

cpacker:
Rolling into #1110

PR #679, opened 2023-12-22
Fix/memgpt agent ignores user messages

javiersastre:
Fix a bug

**How to test**
When running the original code, creating a user proxy and a single memgpt agent, 

**Have you tested this PR?**
Debugged code before and after change. Used MemGPTAgent in an app that had an AutoGen UserProxy and a single MemGPTAgent (no groupchat). First message from user is just "hi", then second message gets ignored and MemGPTAgent just returns the default reply.

**Related issues or PRs**

No, it is under 10 lines.

**Additional context**
- In script memgpt/autogen/memgpt_agent.py, line 289, the new messages entered by the user are retrieved.
- While MemGPTAgent is processing these messages, additional messages are appended to the list of new messages; however these additional messages are not stored in a way that the next time the new messages are retrieved, they will be taken into account.
- In line 332, variable messages_processed_up_to_idx is the updated to a too high value due to these additional messages
- In the next interaction round (when the user enters another message), MemGPTAgent tries to retrieve the list of new messages, but it gets an empty list of messages, skipping the new user message due to a too high value of variable messages_processed_up_to_idx. As there are no new messages to process, MemGPTAgent just returns the default reply.

Apart from the bug, pre-commit added some missing newlines at the end of some files, and I added a couple of directories to .gitignore (PyCharm project settings, and the temporary folder 'test_chroma' created by the unit tests)


memgpt/autogen/memgpt_agent.py
```diff
@@ -328,7 +329,7 @@ def _generate_reply_for_user_message(
 
         # Pass back to AutoGen the pretty-printed calls MemGPT made to the interface
         pretty_ret = MemGPTAgent.pretty_concat(self.agent.interface.message_list)
-        self.messages_processed_up_to_idx += len(new_messages)
+        self.messages_processed_up_to_idx += new_messages_count
```
cpacker, line 332: Reviewing the code - this line is the core diff right? Thank you for this PR, it sounds reasonable to me.

Can you remove the diffs on the other files (other than memgpt/autogen/memgpt_agent.py)?

Additionally, if possible, can you include a short stdout dump of the bug behavior pre-PR and the fixed behavior post-PR? This will help for references in the future back to this PR.

Thank you again for your contribution!

cpacker (changes requested):

memgpt/autogen/memgpt_agent.py
javiersastre: Hi cpacker, thanks for the response. Sure, I can remove the changes to the other files, though note those where introduced by running pre-commit, as requested in the contributing guide. There are missing white spaces at the end of some files, this issue will happen over and over until fixed.

There is not much to print in stdout: as I described in the pull request description, memgpt_agent simply ignores the second user message, printing the default message instead of actually processing the second message and answering accordingly. If the default message is an empty string, it will print nothing.

javiersastre: Could revert changes to .gitignore but couldn't revert changes to memgpt_coder_autogen.ipynb and anna_pa.txt as these are forced by pre-commit hook (missing newline at the end of file).

PR #685, opened 2023-12-23
fix: moved configs for hosted to https, patched bug in embedding creation

cpacker:

- Moved hosted endpoint addresses to https (tested it's working)
- While testing I noticed that there was a bug w/ embedding creation, and patched it

### DIFF
--- a/memgpt/memory.py
+++ b/memgpt/memory.py
@@ -338,6 +338,16 @@ class EmbeddingArchivalMemory(ArchivalMemory):
             # breakup string into passages
             for node in parser.get_nodes_from_documents([Document(text=memory_string)]):
                 embedding = self.embed_model.get_text_embedding(node.text)
+                # fixing weird bug where type returned isn't a list, but instead is an object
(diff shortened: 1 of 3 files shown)


memgpt/memory.py
```diff
@@ -338,6 +338,16 @@ def insert(self, memory_string):
             # breakup string into passages
             for node in parser.get_nodes_from_documents([Document(text=memory_string)]):
                 embedding = self.embed_model.get_text_embedding(node.text)
+                # fixing weird bug where type returned isn't a list, but instead is an object
+                # eg: embedding={'object': 'list', 'data': [{'object': 'embedding', 'embedding': [-0.0071973633, -0.07893023,
+                if isinstance(embedding, dict):
+                    try:
+                        embedding = embedding["data"][0]["embedding"]
```
cpacker, line 345: @sarahwooders I'm not sure if this is a llama-index bug, or if it's to do with how we're returning the payload from the hosted embedding server

Issue #697, opened 2023-12-25
Add MemGPT "Python Client"


cpacker:
## Is your feature request related to a problem? Please describe.

tl;dr support something similar to `openai.client` for python developers

---

Add a familiar python interface to the API server. In essence, instead of making python developers write REST/requests code or asking them to do this (taken from #689):
```python
### imports
# (truncated)

### wrapper code
def process_agent_step(agent, user_message, no_verify):
    """Copied from main.py (unpacks the content of agent.step())"""
    new_messages, heartbeat_request, function_failed, token_warning = agent.step(user_message, first_message=False, skip_verify=no_verify)

    skip_next_user_input = False
    if token_warning:
        user_message = system.get_token_limit_warning()
        skip_next_user_input = True
    elif function_failed:
        user_message = system.get_heartbeat(constants.FUNC_FAILED_HEARTBEAT_MESSAGE)
        skip_next_user_input = True
    elif heartbeat_request:
        user_message = system.get_heartbeat(constants.REQ_HEARTBEAT_MESSAGE)
        skip_next_user_input = True

    return new_messages, user_message, skip_next_user_input

def send_agent_a_message(agent, user_input, no_verify=False, allow_multi_step=True):
    """A convenience wrapper around the agent step, which:
    1. Packages the first message correctly
    2. Allows the agent to run back-to-back calls
    """
    # package the message
    user_message = system.package_user_message(user_input)

    while True:
        try:
            new_messages, user_message, skip_next_user_input = process_agent_step(agent, user_message, no_verify)
            if not allow_multi_step or not skip_next_user_input:
                break
        except KeyboardInterrupt:
            print("User interrupt occured.")
            input("Continue?")
        except Exception as e:
            print(f"An exception ocurred when running agent.step(): {e}")

### Main loop

persona_desc = utils.get_persona_text("sam_pov.txt")
user_desc = utils.get_human_text("basic.txt")

# Create an AgentConfig option from the inputs
agent_config = AgentConfig(
    # name="agent_4",
    name="bill_gates",
    persona=persona_desc,
    human=user_desc,
    preset="memgpt_chat",
    model="gpt-4",
    model_endpoint_type="openai",
    model_endpoint="https://api.openai.com/v1",
    context_window=8192,
)
```

Allow developers to use a `client` that sits inside the package and is a light wrapper on top of the API server (https://github.com/cpacker/MemGPT/blob/main/memgpt/server/server.py):
```
import memgpt.client as memgpt_client

[...]

PR #713, opened 2023-12-26
feat: Add MemGPT "Python Client"

BabellDev:
Implementation of local client to ease integration with other projects

**How to test**
Test code is TBD, but here is how a basic use might look:
```python
from memgpt.config import AgentConfig
from memgpt.client.client import Client
from memgpt import constants

# Create an AgentConfig option from the inputs
agent_config = AgentConfig(
    name="BabellDev",
    persona=constants.DEFAULT_PERSONA,
    human=constants.DEFAULT_HUMAN,
    preset="memgpt_chat",
    model="gpt-4",
    model_endpoint_type="openai",
    model_endpoint="https://api.openai.com/v1",
    context_window=8192,
)

client = Client(auto_save=True)
agent_name = client.create_agent(agent_config=agent_config)


def send_message(message: str):
    response = client.user_message(agent_id=agent_name, message=message)
    for r in response:
        if "assistant_message" in r:
            print("ASSISTANT:", r["assistant_message"])


send_message("Hello my name is BabellDev, I enjoy open source development")
```

**Have you tested this PR?**
TDB

**Related issues or PRs**
#697

No

**Additional context**


memgpt/client/client.py
```diff
@@ -0,0 +1,62 @@
+from typing import Dict, List, Union
+
+from memgpt.config import AgentConfig
+from memgpt.persistence_manager import PersistenceManager
+from memgpt.server.rest_api.interface import QueuingInterface
+from memgpt.server.server import SyncServer
+
+
+class Client(object):
```
BabellDev, line 9: Not too sure about the name `Client`.  It sort of implies it will be connecting to a rest or ws server.  Maybe `LocalClient`?

memgpt/client/client.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(
+        self,
+        auto_save: bool = False,
+    ):
+        self.user_id = "null"
+        self.auto_save = auto_save
```
BabellDev, line 16: Please let me know if you think `auto_save` is a concept worth keeping

memgpt/client/client.py
```diff
@@ -0,0 +1,66 @@
...
+from memgpt.config import AgentConfig
+from memgpt.persistence_manager import PersistenceManager
+from memgpt.server.rest_api.interface import QueuingInterface
+from memgpt.server.server import SyncServer
+
+
+class Client(object):
+    def __init__(
+        self,
+        auto_save: bool = False,
```
cpacker, line 12: I think this should probably be a property of the server too? But maybe out of scope for this PR

memgpt/__init__.py
```diff
@@ -1 +1,3 @@
 __version__ = "0.2.10"
+
+from memgpt.client.client import Client as MemGPT
```
cpacker, line 3: imo might make sense to mimic this style for dev familiarity: https://github.com/openai/openai-python

memgpt/client/client.py
```diff
@@ -0,0 +1,74 @@
...
+from memgpt.server.rest_api.interface import QueuingInterface
+from memgpt.server.server import SyncServer
+
+
+class Client(object):
+    def __init__(
+        self,
+        # Utility settings:
+        auto_save: bool = False,
+        # Allow running quickstart on a client init
```
cpacker, line 14: @BabellDev what do you think about this extra args? Basically I'm thinking it should be possible to do something like the following without even having run `memgpt configure` on a fresh install:

```python
from memgpt import MemGPT

client = MemGPT(
  auto_save=True,
  quickstart="memgpt"
)
```

or
client = MemGPT(
  auto_save=True,
  quickstart="openai",
  config={"openai_api_key": "sk-..."},
)
```

The exact default values / types aren't finished they're just easy placeholders (should probably be unions of pydantic models that are defined elsewhere for server).

File: memgpt/client/client.py (hunk 3 above)
Comment (BabellDev) [reply memgpt/client/client.py:13]: I thought about that, but then it would become a state of the ws/rest servers at startup and you wouldn't be able to toggle it on/off (unless we exposed an endpoint to configure the server).  As you say, maybe something for a future PR.

File: memgpt/client/client.py (hunk 5 above)
Comment (BabellDev) [reply memgpt/client/client.py:14]: Yeah this is a great idea.  If we can avoid calling `memgpt configure`, then it's even easier to get started with this.

File: memgpt/client/client.py [hunk 6]
```diff
@@ -0,0 +1,92 @@
...
+from memgpt.persistence_manager import PersistenceManager
+from memgpt.server.rest_api.interface import QueuingInterface
+from memgpt.server.server import SyncServer
+
+
+class Client(object):
+    def __init__(
+        self,
+        auto_save: bool = False,
+        quickstart: Union[QuickstartChoice, None] = None,
```
BabellDev, line 16: @cpacker I thought about defaulting this to `memgpt`, but then the act of simply constructing the client would (re)configure the system.  I feel that's something the caller should decide explicitly.

cpacker (changes requested):
Made another pass an I think this should be good to go (unless there's anything else you see or notice that's wrong in the code I added @BabellDev)

As a final check, @BabellDev can you try running all the examples in the README (also included a version for memgpt.readme.io, but it's identical so only need to test once)? Just to make sure that the documentation is correct for what's getting merged in.

Maybe try on a fresh / isolated install with no `~/.memgpt` just to be safe.

docs/python_client.md
```diff
@@ -0,0 +1,88 @@
...
+  }
+)
+
+# Now that we have an agent_name identifier, we can send it a message!
+# The response will have data from the MemGPT agent
+my_message = "Hi MemGPT! How's it going?"
+response = client.user_message(agent_id=agent_id, message=my_message)
+```
+
+## More in-depth example of using the MemGPT Python client
```
cpacker, line 33: We might want to break this up into several chunks, right now the way it gets formatted in readme (https://memgpt.readme.io/docs/python_client) is kind of annoying since it turns into a scrollview (bad UX)

memgpt/client/client.py
cpacker: I made a modification so that if config doesn't exist, we default to quickstart openai (maybe makes more sense to default to quickstart memgpt but we can change that easily later in a subsequent PR)

memgpt/client/client.py
cpacker: Make some adjustments and tested with completely empty ~/.memgpt dir and it seems to work

cpacker (approved):

PR #714, opened 2023-12-27
fix: increase the func return char limit

cpacker:

Retrieval calls coming from archival search should not trigger clipping (limit should be set high enough)


sarahwooders (approved):
lgtm!

cpacker:
Working:

image

cpacker:
Update with the extra output suppression:

image

PR #393, opened 2023-11-09
Simple docker.

BobKerns:
**Please describe the purpose of this pull request.**
If you aren't familiar with postgres (I was not, tho plenty of other DBs), it is hard to set up.

This PR is a single-file docker setup,  that results in a properly configured database.

1) build
2) run container
3) configure (default connection: postgres+pg8000://memgpt:memgpt@localhost:5432/memgpt)
4) run memgpt.

Data can be persisted in a docker volume or a directory on disk.

The commands in 1 and 2 could be packaged as simple scripts to save typing, or copy-pasted from the documentation.


**How to test**
Directions are in the file db/Docker.simple

**Have you tested this PR?**
Yes.

#### Build container:
```bash
simple-docker:~p/MemGPT$ docker build -f db/Dockerfile.simple -t pg-test .
[+] Building 1.8s (10/10) FINISHED           
…some build output elided…                                                                                                                                                                                                                                                                                                         
View build details: docker-desktop://dashboard/build/desktop-linux/desktop-linux/8f9lupn6thwu27hyb3ypurpyt

What's Next?
  View a summary of image vulnerabilities and recommendations → docker scout quickview
```

#### Run container
```bash
simple-docker:~p/MemGPT$ docker run -d --rm \
   --name memgpt-db-test \
   -p 8888:5432 \
   -e POSTGRES_PASSWORD=password \
   -v memgpt_db_test:/var/lib/postgresql/data \
    pg-test:latest
2940174a5004ecdc3d7855231cc65c9aa9a1222347e2c593127f341a7b5a3601
```

#### Configure MemGPT:
```bash
simple-docker:~p/MemGPT$ python3 -m memgpt configure
? Do you want to enable MemGPT with Open AI? Yes
? Do you want to enable MemGPT with Azure? No
? Select default inference endpoint: openai
? Select default embedding endpoint: openai
? Select default preset: memgpt_chat
? Select default model (recommended: gpt-4): gpt-4
['sam_pov', 'memgpt_starter', 'memgpt_doc', 'sam_simple_pov_gpt35', 'sam']
? Select default persona: sam_pov
['cs_phd', 'basic']
? Select default human: cs_phd
? Select storage backend for archival data: postgres
? Enter postgres connection string (e.g. postgresql+pg8000://{user}:{password}@{ip}:5432/{database}): postgresql+pg8000://memgpt:memgpt@localhost:8888/memgpt
Saving config to /Users/rwk/.memgpt/config
archival storage postgres
```

[...]

PR #715, opened 2023-12-27
ci: Run tests using postgres docker container

sarahwooders:
Moving workflow tests to use postgres running in container created by #393


cpacker (approved):
phenomenal

PR #726, opened 2023-12-28
fix: cleanup failed agent creation

### LINKED ISSUE #709: control-C instead of exiting a chat session with /exit it corrupts the agent
**Describe the bug**
This bug results from an end-user pressing control-C instead of exiting a chat session with /exit in an Agent.  By doing this, it corrupts the agent and can no longer be loaded.  I noticed it was implemented in the server but not for run.

**Please describe your setup**

- [ ] MemGPT version
  - What is the output of `memgpt version`? 0.2.10
- [ ] How did you install memgpt?
  - `pip install pymemgpt`? `pip install pymemgpt-nightly`? `git clone`? pip and git
- [ ] Describe your setup
  - What's your OS (Windows/MacOS/Linux)? Windows/Linux
  - How are you running `memgpt`? `cmd.exe`/Powershell/Terminal

**Additional context**
It does not happen all the time but can cause some problems. 
example signal capture:
```#!/usr/bin/env python
import signal
import sys

def signal_handler(sig, frame):
    print('You pressed Ctrl+C!')
    sys.exit(0)

signal.signal(signal.SIGINT, signal_handler)
print('Press Ctrl+C')
signal.pause()```

### PR DESCRIPTION
**Pull request by cpacker**
Close #709 

From Discord:
> When running memgpt run --preset foo-system --persona foo --agent foo --human bar for the first time, if the setup fails (in my case because i hadn't deployed the preset files yet), the new agent dir and config is created, but not the state file. 
> When I run it again, startup fails due to a missing agent state file. 
> Deleting the agent dir allows the agent to be created successfully

Reproducing the error:
```sh
memgpt run --preset banana
```
```

🧬 Creating new agent...
->  🤖 Using persona profile 'sam_pov'
->  🧑 Using human profile 'basic'
Failed to create agent from provided information:
Preset 'banana.yaml' not found
```
```
$ ls -lrt ~/.memgpt/agents
agent_1
agent_2
```
```
$ ls -lrt ~/.memgpt/agents/agent_2
config.json
```

**How to test**

Run `memgpt run --preset banana` and make sure no new directory is created (ie it's created, then deleted).

**Have you tested this PR?**

Yes:
```
% ls -rt ~/.memgpt/agents
agent_1
```
```
% memgpt run --preset banana

🧬 Creating new agent...
->  🤖 Using persona profile 'sam_pov'
->  🧑 Using human profile 'basic'
Failed to create agent from provided information:
Preset 'banana.yaml' not found
```
```
% ls -rt ~/.memgpt/agents   
agent_1
```

### DIFF
--- a/memgpt/cli/cli.py
+++ b/memgpt/cli/cli.py
@@ -2,6 +2,7 @@ import typer
 import json
 import requests
 import sys
+import shutil
 import io
 import logging
 import questionary
@@ -466,6 +467,21 @@ def run(
             )
         except ValueError as e:
             typer.secho(f"Failed to create agent from provided information:\n{e}", fg=typer.colors.RED)
+            # Delete the directory of the failed agent
+            try:
+                # Path to the specific file
+                agent_config_file = agent_config.agent_config_path
+
+                # Check if the file exists
+                if os.path.isfile(agent_config_file):
+                    # Delete the file
+                    os.remove(agent_config_file)
+
+                # Now, delete the directory along with any remaining files in it
+                agent_save_dir = os.path.join(MEMGPT_DIR, "agents", agent_config.name)
+                shutil.rmtree(agent_save_dir)
+            except:
+                typer.secho(f"Failed to delete agent directory during cleanup:\n{e}", fg=typer.colors.RED)
             sys.exit(1)
         typer.secho(f"🎉 Created new agent '{agent_config.name}'", fg=typer.colors.GREEN)
 


memgpt/cli/cli.py
```diff
@@ -466,6 +467,21 @@ def run(
             )
         except ValueError as e:
             typer.secho(f"Failed to create agent from provided information:\n{e}", fg=typer.colors.RED)
+            # Delete the directory of the failed agent
```
cpacker, line 470: @sarahwooders not sure if this is too dangerous (rm -r) - alternatively we could move to /tmp instead of deleting?

Issue #577, opened 2023-12-05
Allow linking of functions during chat


cpacker:
Right now if you want to add custom functions you have to define them ahead of time and load them via the preset.

TODO: allow functions to get linked during chat

cpacker:
> > I think the most important implementation detail will probably on the UX side - how do we expect users to want to link/update functions mid-chat / in-between chat? What kind of CLI commands do we need to support this? Which agent state files do we need to write to to support this (to make sure the change persists across saves/loads).
> 
> Agreed, the mid chat UX is tricky.
> 
> Personally I was somewhat surprised that adding a function, adding it in the preset, and restarting the agent didn't result in the new function being added. The simplest thing to do might be to simply change the default behavior and pull the list of functions from the preset config rather than the state.
> 
> I can't really think of a reason you _wouldn't_ want to absorb functions that were defined for the preset you're using on an agent

maybe a good compromise here making a `/resync` command that syncs the function set back with the preset?

and maybe `/resync functions` for functions, `/resync system` for system prompt, etc., since at that point we might as well allow syncing more than just functions from the preset

tombedor:
> > > I think the most important implementation detail will probably on the UX side - how do we expect users to want to link/update functions mid-chat / in-between chat? What kind of CLI commands do we need to support this? Which agent state files do we need to write to to support this (to make sure the change persists across saves/loads).
> > 
> > 
> > Agreed, the mid chat UX is tricky.
> > Personally I was somewhat surprised that adding a function, adding it in the preset, and restarting the agent didn't result in the new function being added. The simplest thing to do might be to simply change the default behavior and pull the list of functions from the preset config rather than the state.
> > I can't really think of a reason you _wouldn't_ want to absorb functions that were defined for the preset you're using on an agent
> 
> maybe a good compromise here making a `/resync` command that syncs the function set back with the preset?
> 
> and maybe `/resync functions` for functions, `/resync system` for system prompt, etc., since at that point we might as well allow syncing more than just functions from the preset

works for me, would it be ok to add the `/resync functions` as a standalone pr and leave the rest with a `NotImplementedError`

cpacker:
yeah that sounds good!

angrysky56:
I found this, not sure if it helps-

Auto Generated Agent Chat: Function Inception
[...]

PR #734, opened 2023-12-28
feat: CLI command to reload functions and system

tombedor:
*Please describe the purpose of this pull request.*
This is a new feature to support /reload functions and /reload system_prompt. It enables users to add or remove functions from an existing agent.

To add a function to an agent, the user must author a new function AND edit the preset for the agent

*How to test*

I didn't see a great way to add an actual test for this. But I've tested the following cases:

- new function added: agent able to call the function
- function removed: agent unable to call the function (no exception)
- function schema changed: at first I implemented this, but found that memgpt refuses to start if the schema is changed anyway, so I do not handle it.
- new function added that does not have an implementation: debug message but do not throw error.

Have you tested this PR?
I have tested the above use cases.

*Related issues or PRs*
https://github.com/cpacker/MemGPT/issues/577

*Is your PR over 500 lines of code?*
No

*Additional context*

Some open questions:

- should we add a system message to the agent to notify them that their function list has changed? I have left this off for now, but the feature is not all that usable, in my usage the agent is somewhat reluctant to try a new function
- should we document this feature in the README? Or leave for power users to discover?
- I'm wondering if there'd be a good way to test out alterations of the configs, etc. I think I saw somewhere that there's a major refactor coming that puts everything in the database, so not sure it'd be worth it


tombedor:
On using this for a bit, I want to refine this:
- do not prompt the agent for a response when this is run
- return the message as a string
- reload the implementations of the functions (thus if bugs are fixed the reload catches it)

tombedor:
obsoleted by: https://github.com/cpacker/MemGPT/pull/784

PR #760, opened 2024-01-02
docs: word choice in documentation

oceaster:
**Please describe the purpose of this pull request.**
Documentation spelling fix.

**How to test**
Review the change log. (tiny change)

**Have you tested this PR?**
Yeah.

PR #761, opened 2024-01-02
docs: linting, syntax, formatting & spelling fixes for all files

oceaster:
I have fixed all the docs files to be standardised, spell corrected, formatted appropriately and should not cause any red flags on ci/cd pipelines anyone might have when forking this repository. 

**How to test**
Review the commit log, Preview the files on github, Documentation fixes only.

**Have you tested this PR?**
Yes.

**Related issues or PRs**
https://github.com/cpacker/MemGPT/pull/760

**Additional context**
I had issues with the .md files not being formatted properly and it was causing issues with my ci/cd pipeline. Makes sense to merge this back into the mainstream repo.


docs/api.md
```diff
@@ -7,21 +7,21 @@ category: 658135e7f596b800715c1cee
...
 
 > ⚠️ API under active development
-> 
+>
 > The MemGPT API is under **active development** and **breaking changes are being made frequently**. Do not expect any endpoints or API schema to persist until an official `v1.0` of the MemGPT API is released.
-> 
+>
 > For support and to track ongoing developments, please visit [the MemGPT Discord server](https://discord.gg/9GEQrxmVyE) where you can chat with the MemGPT team and other developers about the API.
-
+>
```
cpacker, line 14: I think this should have a newline between the `>` blocks? Otherwise it probably won't render properly as two callout/admonition blocks in readme (though I haven't tested it)

docs/autogen.md
```diff
@@ -1,15 +1,15 @@
...
+category: 6580dab16cade8003f996d17
 ---
 
 > 📘 Need help?
 >
 > If you need help visit our [Discord server](https://discord.gg/9GEQrxmVyE) and post in the #support channel.
 >
 > You can also check the [GitHub discussion page](https://github.com/cpacker/MemGPT/discussions/65), but the Discord server is the official support channel and is monitored more actively.
-
+>
```
cpacker, line 12: Likewise here

docs/embedding_endpoints.md
```diff
@@ -1,26 +1,32 @@
...
 ```sh
 export OPENAI_API_KEY=YOUR_API_KEY # on Linux/Mac
 ```
-Then, configure MemGPT and select `openai` as the embedding provider: 
-```
-> memgpt configure                                                                                    
+
+Then, configure MemGPT and select `openai` as the embedding provider:
+
+```text
```
cpacker, line 19: I think `text` makes sense here because shell will add weird syntax highlighting - however I noticed some of the prior outputs got converted from `text` to `shell` which may cause similar issues

docs/embedding_endpoints.md
```diff
@@ -50,24 +60,25 @@ MemGPT supports running embeddings with any Hugging Face model using the [Text E
...
+```sh
 pip install `pymemgpt[local]`
 ```
-Then, select the `local` option during configuration: 
-```
-> memgpt configure                                                                                    
+
+Then, select the `local` option during configuration:
+
+```sh
```
cpacker, line 79: eg should this be `text` too? or an `sh` block going into a `text` block? to avoid weird highlighting on the CLI outputs

docs/functions.md
```diff
@@ -37,18 +38,19 @@ There are three steps to adding more MemGPT functions:
 ### Simple example: giving MemGPT the ability to roll a D20
 
 > ⚠️ Function requirements
-> 
+>
 > The functions you write MUST have proper docstrings and type hints - this is because MemGPT will use these docstrings and types to automatically create a JSON schema that is used in the LLM prompt. Use the docstrings and types annotations from the [example functions](https://github.com/cpacker/MemGPT/blob/main/memgpt/functions/function_sets/base.py) for guidance.
-
+>
```
cpacker, line 43: another one that should probably just be space

docs/local_llm_settings.md
```diff
@@ -102,10 +112,11 @@ Found completion settings file '/Users/loaner/.memgpt/settings/completions_api_s
 
 ### Example: LM Studio (advanced)
 
-In practice, there are many parameters you might want to set, since tuning these parameters can dramatically alter the tone or feel of the genreated LLM outputs. Let's try changing a larger set of parameters.
+In practice, there are many parameters you might want to set, since tuning these parameters can dramatically alter the tone or feel of the generated LLM outputs. Let's try changing a larger set of parameters.
 
 Now just for reference, let's record the set of parameters before any modifications (truncated to include the parameters we're changing only):
-```
+
+```sh
```
cpacker, line 119: should be text?

docs/ollama.md
```diff
@@ -1,24 +1,26 @@
...
-category: 6580da9a40bb410016b8b0c3 
+title: Ollama
+excerpt: Setting up MemGPT with Ollama
+category: 6580da9a40bb410016b8b0c3
 ---
 
 > ⚠️ Make sure to use tags when downloading Ollama models!
 >
-> Don't do **`ollama run dolphin2.2-mistral`**, instead do **`ollama run dolphin2.2-mistral:7b-q6_K`**.
+> Don't do **`ollama pull dolphin2.2-mistral`**, instead do **`ollama pull dolphin2.2-mistral:7b-q6_K`**.
```
cpacker, line 9: I used `pull` instead of `run` here because the goal is to make sure we preload/download the model, but we're not trying to run it in the ollama CLI.

So I think this should still be `run`, not `pull`? Let me know if I'm misunderstanding the ollama commands

cpacker (changes requested):
Thanks for the PR! Left a few comments on things I think may need revising, otherwise looks good!

docs/api.md
oceaster: You are correct, it just seems to break this convention (https://github.com/DavidAnson/markdownlint/blob/v0.32.1/doc/md028.md).

Personally, I think it makes sense to be one block since these blocks have headers in them which make it clear they are two separate things.

Although, if there's strong feelings about the styling - then it should be a big deal, if anyone uses a markdown linter they can add an exception for this rule.

oceaster: I am adding a .markdownlint.json file to make an exception for this rule.

Seems to be the best solution.

docs/embedding_endpoints.md
oceaster: made it consistent with my latest commit. Believe I found them all anyway.

docs/ollama.md
oceaster: You had 'run' and I changed it to 'pull'.

Because run download, loads and runs the model in the terminal. Blocking the session from further use.

Pull just downloads and prepares the model for use.

Common confusion because the ollama docs just say to use 'run' but I think it's a noob friendly kind of approach there.

But for this use-case you'd only want to pull. Since you aren't using the model in your terminal directly via ollama when using with memgpt.

oceaster (commented):
Solved all issues.

oceaster (commented):
Oops, forgot to mark this as resolved.

cpacker:
@oceaster do you want me to reopen this PR? Not sure if you closed it by mistake or not

oceaster:
> @oceaster do you want me to reopen this PR? Not sure if you closed it by mistake or not

Oh, absolutely no idea how this was closed. Thanks for noticing, yes please re-open.

docs/.markdownlint.json
```diff
@@ -0,0 +1,6 @@
+{
```
cpacker, line 1: If we're adding any link files to the repo, we should add the corresponding github action to enforce it

imo let's either remove this file from the PR, or add a github actions file to run the markdown lint check

cpacker (changes requested):

docs/.markdownlint.json
cpacker: ^feel free to disagree if you have a strong opinion on this

oceaster: Name: markdownlint
Id: DavidAnson.vscode-markdownlint
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint

Personally I'd prefer just use this and let my IDE tell me what I should and shouldn't be doing - until there is a more well defined structure for markdown within the project.

Because it might just slow people down in the meantime, while those rules are being tweaked.

Either way, could merge it in now and then the action later, or keep this open and wait for an action.

I'm moving house tomorrow, so might be a few days before I get a chance to write up that action, although you're welcome to write the action yourself if you want, of course.

cpacker: Makes sense! I'll just merge this in now, and add the github action later.

cpacker (approved):

PR #780, opened 2024-01-04
docs: fix invalid entry in response dict

### LINKED ISSUE #773: Doc: Invalid entry in client.user_message response
https://github.com/cpacker/MemGPT/blob/cc4458db583c22d751a9058853170a0bc9c4f6e1/docs/python_client.md?plain=1#L84

I think there'll never be a response that contains a key `thoughts`. I searched the codebase everywhere and the only possibly related code snippet is `inner_thoughts`. I believe in the referred code above, `thoughts` should be `internal_monologue`.

ifsheldon:
Fix #773


cpacker (approved):
Thank you!

PR #781, opened 2024-01-04
fix: 'dict' object has no attribute 'name'

ifsheldon:
Bug: when `throw_if_exists` is `True` and an agent exists and agent_config is a dict, then `'dict' object has no attribute 'name'` is thrown. This is fixed simply use existing `agent_name`.


cpacker (approved):
LGTM, thanks!

PR #783, opened 2024-01-04
fix: Update message schema / data type to match OAI tools style

cpacker:
TODOs:

- [x] Make `tool_calls` JSON column
- [x] Modify db.py to match
- [x] Strip stray TODO comments
- [x] Make `ToolCall` record?

Align new data types to be closer to OAI schema.

OAI:

image


memgpt/data_types.py
```diff
@@ -36,9 +55,8 @@ def __init__(
         model: str,  # model used to make function call
         user: Optional[str] = None,  # optional participant name
         created_at: Optional[str] = None,
-        tool_name: Optional[str] = None,  # name of tool used
```
cpacker, line 39: One weird consequence of following this official OpenAI schema is that when you have (in isolation) a message of role "tool" (meaning it was the response message from a function execution), you do not immediately have access to the tool/function "name" anymore (in the previous official OpenAI schema for functions, this existed in the response message itself).

Instead, to get the name you have to do a lookup into the hypothetical "tools" table using the "tool_call_id" to retrieve the tool name.

This means that if our internal agent representation is using the old "functions" style (which it is, and I think it totally fine for now), when we reconstruct the internal agent representation from the database (eg when we load an agent and populate the in-context messages by retrieving from the DB), we will need to do an extra lookup to grab the tool/function name to place inside `message['name']` for the messages with role == 'function':

image

sarahwooders (approved):
lgtm!

memgpt/data_types.py
```diff
@@ -24,8 +24,31 @@ def __init__(self, id: Optional[str] = None):
...
 
 
+class ToolCall(object):
+    def __init__(
+        self,
+        id: str,
+        # TODO should we include this? it's fixed to 'function' only (for now) in OAI schema
+        tool_call_type: str,  # only 'function' is supported
+        # TODO in the OAI JSON schema, this is not flat, it's an obj
+        # function: { 'name': ..., 'arguments': ...}
```
cpacker, line 33: Can unflatten here, eg:

```python
class ToolCallFunction(object):
    def __init__(
        self,
        name,
        arguments,
    ):
    ...

class ToolCall(object):
    def __init__(
        self,
        id: str,
        tool_call_type: str, 
        function: ToolCallFunction,
    ):
```

cpacker: Let's do this later when we migrate to pydantic models for the data types

PR #790, opened 2024-01-06
feat: Feature/global logging

agiletechnologist:
Reconfigure global logging to use Constants instead of config

**How to test**
To test you can run ```memgpt configure``` it should create a directory .memgpt/logs/MemGPT.log 

This only exists in config.py with only one debug for save. 

**Have you tested this PR?**
This is the first stage.  I will be creating a pytest in a commit soon. 

**Related issues or PRs**
https://github.com/cpacker/MemGPT/issues/319

No.


cpacker (approved):
Pending small comment

memgpt/config.py
```diff
@@ -265,11 +269,13 @@ def load(cls) -> "MemGPTConfig":
                 "memgpt_version": get_field(config, "version", "memgpt_version"),
             }
             config_dict = {k: v for k, v in config_dict.items() if v is not None}
+
             return cls(**config_dict)
 
         # create new config
         anon_clientid = MemGPTConfig.generate_uuid()
         config = cls(anon_clientid=anon_clientid, config_path=config_path)
+        config.create_config_dir()  # create dirs
```
cpacker, line 278: @sarahwooders @agiletechnologist this should be done already at the top of configure and run/quickstart - do we need to call it so many times?

I guess it's not a big deal since nothing bad will happen by calling it too many times, just curious about the placement of the calls.

agiletechnologist: @sarahwooders Sometimes, users delete the agents and other directories, Which generates errors in the directories. do not exist. the function does not create again if it is called multiple times the way that method is written, but if it does, it skips it. it is not much of a hit on a load of Memgpt in favor of avoiding fatal errors.  There is no method of deleting corrupted or unwanted agents in MemGPT.

PR #798, opened 2024-01-09
fix: Fix broken doc due to refactor

ifsheldon:
This fixes the broken Client documentation due to recent refactoring.


sarahwooders (approved):
This lgtm! I believe the tests are failing because you don't have an `OPENAI_API_KEY` set in github, but we can push a fix to avoid that soon.

PR #833, opened 2024-01-16
fix: properly handle CLI flags passed to `memgpt run`

cpacker:

Properly handle CLI flags passed to `memgpt run` when creating a new agent (previously they were ignored)

E.g.:
```sh
memgpt run --model-wrapper chatml-noforce
```

Would not actually set the wrapper to `chatml-noforce` if you created a new agent.

**How to test**

`memgpt run --model-wrapper XXX --debug`, where `XXX` is not the default. Inspect the output of `--debug` to spot check that the correct prompt formatter is being used.

### DIFF
--- a/memgpt/cli/cli.py
+++ b/memgpt/cli/cli.py
@@ -492,14 +492,46 @@ def run(
             # agent = f"agent_{agent_count}"
             agent = utils.create_random_username()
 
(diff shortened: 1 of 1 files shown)


sarahwooders (approved):
lgtm!

PR #835, opened 2024-01-17
refactor: remove User LLM/embed. defaults, add credentials file, add authentication option for custom LLM backends

### LINKED ISSUE #604: Allow sending keys / bearer tokens on local/custom LLM calls
**Is your feature request related to a problem? Please describe.**

When using local / open / custom LLM backends, there's no option to send a bearer token through the request. However some people might want to lock down their endpoints or might be using endpoints (eg runpod serverless endpoints) that are only accessible with a bearer token in the request header.

https://docs.runpod.io/docs/create-serverless-endpoint

Would need to store the token in the config or pull from env vars. Also need to decide on logic for when to pass it through (I think in most cases, passing a bearer token when the server doesn't need it shouldn't cause any errors? In which case we can default to always passing it as long as it exists in the config).

Also, not all backends might use bearer tokens for authorization, so we might need to have two variables in config:
```
config.custom_endpoint_key = ...
config.custom_endpoint_auth_type = "bearer-token" / "api-key"
```

OpenAI uses bearer tokens ([API ref](https://platform.openai.com/docs/api-reference/chat/create)):
```sh
curl https://api.openai.com/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
    "model": "gpt-3.5-turbo",
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant."
      },
      {
        "role": "user",
        "content": "Hello!"
      }
    ]
  }'
```

Azure OpenAI uses keys ([API ref](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference)):
```sh
curl https://YOUR_RESOURCE_NAME.openai.azure.com/openai/deployments/YOUR_DEPLOYMENT_NAME/completions?api-version=2023-05-15\
  -H "Content-Type: application/json" \
  -H "api-key: YOUR_API_KEY" \
  -d "{
  \"prompt\": \"Once upon a time\",
  \"max_tokens\": 5
}"
```

cpacker:
Closes #604

---

- Removes `User` LLM/embedding defaults, these should get pulled from the base MemGPT config now
- Adds a `~/.memgpt/credentials` file, this is where credentials sit now
- Adds an authentication option for custom LLM backend POST requests
  - Currently adds support for bearer tokens and API keys in the HTTP headers
- Note: renames `openai_tools.py` to `llm_api_tools.py`

**How to test**

- Try using one of the completions local LLM routes to point at an API service that is compatible but requires a key

**Have you tested this PR?**

- TODO

---

TODOs:

- [ ] Add setting of `auth_type` and `auth_key` during `memgpt configure` for local/open/custom LLMs


memgpt/data_types.py
```diff
@@ -303,6 +306,9 @@ def __init__(
         model_endpoint: Optional[str] = "https://api.openai.com/v1",
         model_wrapper: Optional[str] = None,
         context_window: Optional[int] = None,
+        # generic auth options (for local LLMs)
```
cpacker, line 309: @sarahwooders OK to put these in here?

sarahwooders (approved):
lgtm!

PR #838, opened 2024-01-17
feat: Partial support for OpenAI-compatible assistant API 

sarahwooders:
* Initial subset of function to allow users to interact with MemGPT agents in the same way as OpenAI's [assistants API ](https://platform.openai.com/docs/api-reference/assistants). 
* Modify `server.user_message(...)` to allow passing in a `Message` object (not just `str`). 
* Add an example using the OpenAI python client `examples/openai_client_assistants.py`
* Add tests with the OpenAI python client and also python requests client 

**How to test**
`poetry run pytest -s tests/test_openai_assistant_api.py`

No 

**Todos:** 
Remaining todos in: #892


Maximilian-Winter:
@sarahwooders Would be happy to help with this one, I think this is a great addition. After paying 120$ dollars for Assistants API calls.

sarahwooders:
> @sarahwooders Would be happy to help with this one, I think this is a great addition. After paying 120$ dollars for Assistants API calls.

That's great to hear! I should be close to wrapping this up in the next 1-2 days but would be great to get your help with testing it out :)

memgpt/agent.py
```diff
@@ -581,6 +589,7 @@ def step(self, user_message, first_message=False, first_message_retry_limit=FIRS
                         raise Exception(f"Hit first message retry limit ({first_message_retry_limit})")
 
             else:
+                print("input message", [type(m) for m in input_message_sequence])
```
cpacker, line 592: stray print?

memgpt/server/rest_api/openai_assistants/assistants.py
```diff
@@ -0,0 +1,505 @@
...
+  cd memgpt/server/rest_api
+  poetry run uvicorn server:app --reload
+"""
+
+interface: QueuingInterface = QueuingInterface()
+server: SyncServer = SyncServer(default_interface=interface)
+
+
+# router = APIRouter()
+app = FastAPI()
```
cpacker, line 50: I think we should probably put this inside `rest_api/server`?

So that when you spin up the memgpt server with `memgpt server`, you get a single FastAPI server that's fielding both the MemGPT API requests + the OpenAI Assistants API requests?

cpacker (changes requested):

memgpt/server/rest_api/openai_assistants/assistants.py
sternb12: This sounds super useful. Looking forward to trying

cpacker (approved):
LGTM

PR #842, opened 2024-01-18
build: Add Python dev tools: black and mypy to Poetry dev group

robbyt:
Adds two commonly used tools to aid in Python development, `mypy` (to perform static analysis), and `black` (the opinionated code format tool).

`black` is already in the main `tool.poetry.dependencies` as an optional package, however I believe it should be moved to dev tools, because it's not required for the main application to run. This change will require a follow-up PR to remove `black` from the main dependencies, and update other CI scripts if needed.

MemGPT developers can use these tools to improve the quality of the code, and also ensure consistent code formatting. 

**How to test**
This change does not touch any application code. 

**Have you tested this PR?**
n/a

**Additional context**
To add these tools, I ran the following:
```
$ poetry add --group dev mypy@latest black@latest
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -273,7 +273,7 @@ lxml = ["lxml"]
 name = "black"
 version = "23.12.1"
 description = "The uncompromising code formatter."
-optional = true
+optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


sarahwooders:
Thanks for making this PR @robbyt ! It's probably better to move the dev dependencies into `tool.poetry.dependencies` instead of as an extra, but if we make this change, could we also move all the other dev dependencies as well? Basically `dev = ["pytest", "pytest-asyncio", "pexpect", "black", "pre-commit", "datasets"]`.

sarahwooders:
I'm going to close this for now, but please feel free to re-open!

PR #843, opened 2024-01-18
build: Update the pytest script to cache poetry artifacts, so tests finish faster

robbyt:

This change adds a 3rd party action, from here:
https://github.com/marketplace/actions/setup-and-cache-python-poetry

This change will reduce test runtime, because python packages will not need to be downloaded every time this test runs. This will also improve the reliability of tests, because tests will continue working if Pypi becomes unavailable.

**How to test**
n/a

**Have you tested this PR?**
n/a

**Related issues or PRs**
n/a

n/a

**Additional context**
I am happy to update this to *not* use a 3rd party action if you prefer.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,18 +23,16 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


cpacker (approved):
Awesome, thank you! Failing tests should be fine/patched on main.

PR #857, opened 2024-01-19
feat: Configurable wrapper

Maximilian-Winter:
It adds a configurable wrapper, load and saveable to YAML

**How to test**
Use the new experimental neural chat wrapper I added.

**Have you tested this PR?**
Yes, with a neural chat wrapper.

No

**Additional context**
Here is a chat with a neural chat wrapper I added by configuring it:

Screenshot 2024-01-19 221006


memgpt/local_llm/llm_chat_completion_wrappers/configurable_wrapper.py
```diff
@@ -0,0 +1,301 @@
...
+
+from .wrapper_base import LLMChatCompletionWrapper
+from ...constants import JSON_ENSURE_ASCII
+from ..json_parser import clean_json
+from ...constants import JSON_ENSURE_ASCII
+from ...errors import LLMJSONParsingError
+
+
+# A configurable model agnostic wrapper.
+class ConfigurableWrapper(LLMChatCompletionWrapper):
```
cpacker, line 14: Maybe we should call this `ConfigurableJSONOutputWrapper`?

(Since it's specific to JSON outputs, vs a potential YAML version)

Maximilian-Winter: Sure.

Maximilian-Winter: Do you think I need to add other configurable parameters, like first message?

memgpt/local_llm/llm_chat_completion_wrappers/configurable_wrapper.py
```diff
@@ -0,0 +1,301 @@
...
+        Args:
+            pre_prompt (str): The pre-prompt content.
+            post_prompt(str): The post-prompt content
+            sys_prompt_start (str): The system prompt start.
+            sys_prompt_end (str): The system prompt end.
+            user_prompt_start (str): The user prompt start.
+            user_prompt_end (str): The user prompt end.
+            assistant_prompt_start (str): The assistant prompt start.
+            assistant_prompt_end (str): The assistant prompt end.
+            include_sys_prompt_in_first_user_message (bool): Indicates whether to include the system prompt
```
cpacker, line 46: I think one piece of functionality this wrapper is missing that the ChatML wrappers are missing is the ability to have a special prefix for the first message, eg https://github.com/cpacker/MemGPT/blob/main/memgpt/local_llm/llm_chat_completion_wrappers/chatml.py#L42

cpacker:
This looks pretty good. I think we probably should add some unit tests in `tests/` with some basic outputs that we expect for a given `messages` list and wrapper configuration, in particular testing some configurations that match what we already are using with `chatml` (because at that point, we could even migrate all the custom wrappers from classes to YAML specs on the configurable wrapper - but maybe best to save that for a separate PR).

memgpt/local_llm/llm_chat_completion_wrappers/configurable_wrapper.py
cpacker: Maybe not needed for this initial PR? We can maybe add all that in a follow-up PR that obsoletes the old wrapper classes and replaces them with this the configurable wrapper (assuming it fully supports all the existing ones).

Maximilian-Winter:
@cpacker Absolutely, I will test it and we should wait till some testing from others have taken place.

memgpt/local_llm/llm_chat_completion_wrappers/configurable_wrapper.py
Maximilian-Winter: I will add it now. Will be an easy change.

memgpt/local_llm/llm_chat_completion_wrappers/configurable_wrapper.py
Maximilian-Winter: Is added

cpacker (approved):
@Maximilian-Winter it seems like this is ready to merge in as-is, but we probably want to add some documentation on this at some point (can be this PR or another one) that describes how to use the custom wrapper.

Do you think this should go somewhere like here? https://memgpt.readme.io/docs/adding_wrappers

Or maybe even just replace that documentation page entirely?

Maximilian-Winter:
@cpacker I can add the documentation, but I wasn't sure about how to actually make this configurable. Just pointing to the yaml file?

Maximilian-Winter:
@cpacker The best thing I think would be to make a wrapper folder in the user folder that users can place their yaml wrapper files in. Then we load them and show them at configuration.

cpacker:
> @cpacker I can add the documentation, but I wasn't sure about how to actually make this configurable. Just pointing to the yaml file?

Yeah I guess for this PR the documentation is more along the lines of showing how (in python code) you can create one of the existing wrappers via a class instantiation of the new configurable wrapper. However I do think that's fine to put in a separate PR, or potentially ignore if we merge the other dynamic wrapper in, so I'll go ahead and merge this for now. Can continue the documentation convo here or on Discord DMs.

cpacker:
Thank you for the PR!

PR #855, opened 2024-01-19
refactor: fix typing errors (pyright)

cpacker:
**Please describe the purpose of this pull request.**

- [x] Add pyright to github actions
- [ ] Do a full pass over the codebase to correct pyright errors
- [ ] Disable `continue on error` on github actions once pyright is passing

**How to test**

- [ ] Unit tests pass
- [ ] pyright passes

PR #861, opened 2024-01-19
build: Add a warn-only pyright Github action

robbyt:
Adds a purely informational Pyright type check. There are many errors right now, and Github lacks a "warn" state for jobs, so this will return "success" state, even though it's failing.

Once the errors have all been resolved, the `|| true` can be removed, and it will start reporting real error state.

**How to test**
n/a

**Have you tested this PR?**
I don't have a good way to test this until it lands in `main`. I'll send a follow-up change if needed after it lands.

**Related issues or PRs**
#855

n/a

**Additional context**
Part of this change is minor point-release updates to posthog and llama_index in the `.lock` file. This change was automatically made by `poetry`. I reviewed theses changes in both projects, and I do not expect any compatibility issues.

### DIFF
--- a/.github/workflows/pyright_types.yml
+++ b/.github/workflows/pyright_types.yml
@@ -1,0 +1,41 @@
+name: "Python/Pyright types check"
+on:
+  push:
+    branches: [ main ]
+    paths: ["**.py"]
+  pull_request:
+    branches: [ main ]
+    paths: ["**.py"]
+
+permissions:
+  contents: read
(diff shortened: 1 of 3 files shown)


cpacker:
Should we consider using this action instead? https://github.com/jakebailey/pyright-action

cpacker (approved):
I think the lock file needs to be regenerated due to update on main but can move to another PR

Issue #770, opened 2024-01-02
Create `Document` table for storing original loaded documents 


sarahwooders:
We should have a document store, in addition to storing chunked/embedded passages.

tombedor:
Relates to something I've been thinking about - have the agent write "articles" summarizing what it knows about certain topics, and being able to update.

Currently IIUC there's not really a memory consolidation function, where old or outdated memories can be discarded. Eg if today I tell the agent is green but tomorrow I tell it, sorry I made a mistake yesterday, the sky is blue, ideally that old memory would be pruned somehow

sarahwooders:
Now that #708 is merged, this would be a great issue to get a community contribution on. The issue can be broken down in the following todos: 

- [ ] Add a `Document` data type to `data_types.py` that stores the entire original document text and its embedding and metadata
- [ ] Add a `doc_id` field to `Passage` to associate passages with documents 
- [ ] Add a `TableType.DOCUMENTS` to the `agent_store/storage.py` and support document storage for chroma and postgres 
- [ ] In `cli_load.py`, when ingesting passages, also store the associated document and `doc_id`
- [ ] Add a new function that allows the agent to retrieve the document associated with a passage or search over documents

PR #867, opened 2024-01-20
 feat: Document table for storing original loaded documents

Maximilian-Winter:
Adds the things discussed in this issue: https://github.com/cpacker/MemGPT/issues/770

**Have you tested this PR?**
No, wasn't sure about how to test it.

**Related issues or PRs**
https://github.com/cpacker/MemGPT/issues/770

No


memgpt/cli/cli_load.py
```diff
@@ -135,39 +105,56 @@ def store_docs(name, docs, user_id=None, show_progress=True):
     # compute and record passages
     embed_model = embedding_model(config.default_embedding_config)
 
+    storage = StorageConnector.get_storage_connector(TableType.DOCUMENTS, config, user_id)
+    docs_storage = []
+    for doc in docs:
+        doc_storage = Document(user_id=user_id, text=doc.text, id=uuid.UUID(doc.doc_id), data_source=data_source.name)
```
sarahwooders, line 111: @Maximilian-Winter I may be misunderstanding llama index, but wouldn't this still store the chunked text of a single file, rather than the text of the entire file?

sarahwooders (changes requested):
@Maximilian-Winter could you please also add a test for the document store in `test_load_archival.py` (maybe around here https://github.com/cpacker/MemGPT/blob/main/tests/test_load_archival.py#L138) to ensure documents are properly inserted? I think in the test case example, we'd expect a single document (for the single) file to be inserted, and would alos want to check if the passages retrieved match the document ID.

memgpt/cli/cli_load.py
Maximilian-Winter: No, I don't think so, we are passing the results of the different llama index readers to this function, as far as I could see all readers return a list of documents. These documents represent the complete documents found by the llama index readers, in form of instances of the llama index Document class. The chunks get only created when they are added to the vector index. Where they get  chunked into nodes and get embedded.

Here is the code I use to access and get the chunks and their text and embeddings.
```python
 index = VectorStoreIndex.from_documents(docs, service_context=service_context, show_progress=True)
    doc_info = index.ref_doc_info
    passages = []
    doc_dict = {}
    passages_dict = index.docstore.docs

    # SimpleVectorStore has a get method and takes in the node id to retrieve the embedding.
    if isinstance(index.vector_store, SimpleVectorStore):
        simple_store = index.vector_store
        for doc_id, data in doc_info.items():
            nodes_dict = []
            for node_id in data.node_ids:
                nodes_dict.append((node_id, simple_store.get(node_id), passages_dict[node_id].to_dict()))
            doc_dict[doc_id] = nodes_dict
```

Maximilian-Winter: This here access the chunked documents: passages_dict = index.docstore.docs
Which I access by using the node id found in doc_info. I get doc_info by accessing ref_doc_info which is a dictionary of doc_ids to node ids.

Maximilian-Winter: I also checked the created sqlite database and it contains the unchunked complete text.

Maximilian-Winter:
@sarahwooders I think we should add the test to test_storage.py because the Document store is a database as I have implement it not a archival memory if I get the term archival memory correct. But the passage would be part of the archival test, right?

sarahwooders:
> @sarahwooders I think we should add the test to test_storage.py because the Document store is a database as I have implement it not a archival memory if I get the term archival memory correct. But the passage would be part of the archival test, right?

Ah yeah I think ideally we could update both tests - the `test_load_archival.py`, we'd just want to check to make sure the loaded documents and passages match up with the data thats loaded. And for `test_storage.py` would test to make sure inserting documents works fine (similar to how passages are also tested).

Maximilian-Winter:
@sarahwooders I checked again and SimpleWebpageReader returns one document with the complete text when using it with one page. But SimpleDirectoryReader returns a list of documents, all chunked when using it with one document. I found a work around by creating a new llama index document with the complete text and passing that to the store_docs function. Should also work as expected with multiple documents.

Maximilian-Winter:
@sarahwooders I also added the necessary tests.

memgpt/agent_store/db.py
sarahwooders, line 1: I think you need to add `TableType.DOCUMENTS` here and also for `SQLLiteStorageConnector`?

sarahwooders, line 1: This is a bug on our end (and causing tests for fail) -- `TableType.DATA_SOURCES` doesn't exist anymore so if you remove this the tests should pass.

memgpt/cli/cli_load.py
sarahwooders, line 1: Could you please leave a comment on why you're doing `doc.text[2:]` for future reference?

sarahwooders: Also, do we not need to do the same thing for loading webpages?

sarahwooders (changes requested):
Left a few minor comments, and we should get the tests to pass -- but should be close to merging soon!

memgpt/cli/cli_load.py
Maximilian-Winter: I do this because the SimpleDirectoryReader adds two new lines in the chunks.

Maximilian-Winter:
@sarahwooders I gonna add the comments  about the doc.text[2:] later today.

PR #869, opened 2024-01-20
fix: close db sessions in metadata manager

tombedor:
This is a partial fix, which closes DB sessions. Not doing so will result in accumulation of orphaned db sessions


sarahwooders (approved):
Assuming the tests pass, this looks good to me! I made an issue #869 so similar changes can be made to `agent_store/db.py`.

sarahwooders:
The failing tests are because we aren't skipping OpenAI-dependent tests anymore -- we can fix this in a bit, so I think this is probably still safe to merge.

PR #871, opened 2024-01-20
fix: minor fixes to allow tests to pass without OPENAI_API_KEY set

sarahwooders:


cpacker (approved):
LGTM, we can probably do the toggle to hosted when openai key doesn't exist too

PR #895, opened 2024-01-22
feat: Dynamic template system with editable system prompt fields.

Maximilian-Winter:
This implements a dynamic template field system for the system message editable at runtime through CLI.  This also implements a customizable core memory without hard coded persona and human block. For use cases such as a customer service bot, dungeon master in a DnD game, where no persona and human is necessary or too unspecific and there is need for specific fields etc. 

This all aims to make MemGPT more configurable for use cases outside that of a chatbot with a persona.

**How to test**
Load a system prompt with template fields and edit one through the cli. Create a customizable core memory preset.

**More detail to the template field system for system message:**
It adds a simple but effective template system for building the system message of memgpt.
It has predefined template fields that can be used to add parts of the default memgpt_chat system message.
The following shows the memgpt_chat system message in template form:
298383697-df1b9b81-5199-4107-ac05-201e8bab5887

Each field adds a part of the system message of memgpt_chat, the template system will generate the memgpt system message automatically from it.

The template system get triggered when the system message string in the memgpt preset, ends with '_templated'. Like for example the following memgpt preset:

298383958-d479d704-9b4c-4258-bef4-ef459bccdd03

It then will load the built in template fields, like 'memgpt_introduction' or 'memgpt_realism_authenticity' and tries to load a file with the same name and path as the system message file, but having a file ending of '.yaml' consisting of field names with their respective content. The following is a complete example for instructing memgpt to talk like an old pirate:

memgpt_chat_pirate_templated.txt (in the user folder system_prompts)
```text
{memgpt_introduction}

{memgpt_realism_authenticity}

{memgpt_control_flow}

{memgpt_basic_functions}

{memgpt_memory_editing}

{memgpt_recall_memory}

{memgpt_core_memory}

{memgpt_archival_memory}

{memgpt_introduction_end}

{pirate_instruction}
```

memgpt_chat_pirate_templated.yaml (in the user folder system_prompts)
```yaml
pirate_instruction: 'Always reponsond like an old pirate!!!'
```

memgpt_pirate_preset.yaml (in the user folder presets)
```yaml
system_prompt: "memgpt_chat_pirate_templated"
functions:
  - "send_message"
  - "pause_heartbeats"
  - "core_memory_append"
  - "core_memory_replace"
  - "conversation_search"
  - "conversation_search_date"
  - "archival_memory_insert"
  - "archival_memory_search"
```

Here is the resulting chat response of memgpt using the example preset:
Unbenannt

The template fields are dynamic and editable through the CLI by using the "/edit_template_field" which prompts the user to input the template field name and its value. This allows to change parts of the system prompt dynamically during the runtime of the application.

**Details on the configurable Core Memory**
This makes it possible to initiate the core memory with other or more specific sections and information then the default persona and human section. To activate it the user has to specify his preset with a core_memory_type field set to "custom" and provide a custom_memory_file, which is an additional yaml containing sections and values of the initial core memory.

A user agent preset example:
```yaml
system_prompt: "memgpt_chat"
core_memory_type: "custom"
core_memory_file: "initial_core_memory"
functions:
  - "send_message"
  - "pause_heartbeats"
  - "core_memory_append"
  - "core_memory_replace"
  - "conversation_search"
  - "conversation_search_date"
  - "archival_memory_insert"
  - "archival_memory_search"
```

The file inital_core_memory.yaml, next to the preset:
```yaml
persona: 'The following is a blank slate starter persona, I need to expand this to develop my own personality.

My name is MemGPT.
I am kind, thoughtful, and inquisitive.'

human: 'First name: Chad'
```
This is just an example to create a default memgpt core memory with initial core memory. But you can put any fields and values you want as sections into the core memory yaml file.

**Have you tested this PR?**
Yes

No


Maximilian-Winter:
@cpacker @sarahwooders Maybe you both could give me a short feedback the next days on what you think about the system.

memgpt/agent.py
```diff
@@ -104,22 +109,34 @@ def initialize_memory(ai_notes, human_notes):
...
+                f"### Memory [last modified: {memory_edit_timestamp.strip()}]",
+                f"{len(recall_memory) if recall_memory else 0} previous messages between you and the user are stored in recall memory (use functions to access them)",
+                f"{len(archival_memory) if archival_memory else 0} total memories you created are stored in archival memory (use functions to access them)",
+                "\nCore memory shown below (limited in size, additional information stored in archival / recall memory):",
+                f'<persona characters="{len(memory.persona)}/{memory.persona_char_limit}">' if include_char_count else "<persona>",
+                memory.persona,
+                "</persona>",
+                f'<human characters="{len(memory.human)}/{memory.human_char_limit}">' if include_char_count else "<human>",
+                memory.human,
+                "</human>",
```
cpacker, line 126: @Maximilian-Winter if we're trending in the direction of making the layout of the entire system block dynamic (which I love btw, this is an awesome PR), do you think it makes sense to also go the extra step and make the "template" of the entire system block including the memory dynamic?

Eg right now you have approximately the following in the first system message:
```
{system message, can be dynamically created}
{memory block}
```

where `{memory block}` is
```
### Memory [last modified X]
{human memory}
{persona memory}
{or custom memory}
```

but the `"### Memory [last modified: {memory_edit_timestamp.strip()}]` part is still hardcoded, whereas this could also be rolled into a template maybe? e.g. if people want to play with using other things than xml tags?

Or maybe this is better off just being added as a modification to your configurable core memory code?

memgpt/agent.py
```diff
@@ -184,21 +201,32 @@ def __init__(
...
         self.functions_python = {k: v["python_function"] for k, v in link_functions(function_schemas=self.functions).items()}
         assert all([callable(f) for k, f in self.functions_python.items()]), self.functions_python
-
-        # Initialize the memory object
-        if "persona" not in agent_state.state:
-            raise ValueError(f"'persona' not found in provided AgentState")
-        if "human" not in agent_state.state:
-            raise ValueError(f"'human' not found in provided AgentState")
-        self.memory = initialize_memory(ai_notes=agent_state.state["persona"], human_notes=agent_state.state["human"])
+        if "core_memory_type" in agent_state.state and agent_state.state["core_memory_type"] == "custom":
```
cpacker, line 219: If we make changes to make the entire system block templatable, maybe it makes more sense to call this field something like `system_template_type`? Or have both?

memgpt/agent.py
```diff
@@ -916,13 +977,27 @@ def remove_function(self, function_name: str) -> str:
...
+            updated_state = {
+                "persona": self.memory.persona,
+                "human": self.memory.human,
+                "core_memory_type": "default",
+                "system": self.system,
+                "system_template": self.system_template,
+                "functions": self.functions,
+                "messages": [str(msg.id) for msg in self._messages],
+            }
+        elif isinstance(self.memory, CustomizableCoreMemory):
```
cpacker, line 990: Maybe a better name for `CustomizableCoreMemory` is `CustomizableInContextMemory`?

cpacker (commented):
This is really awesome @Maximilian-Winter , thank you for the PR! Left some high-level design comments (not necessarily requested changes, just some discussion points). 

This looks roughly mergable to me, after some more discussion I can test and merge likely before v0.3 tag (or v0.3.1 latest).

memgpt/agent.py
cpacker: This is probably overkill, but something like "memory_edit_timestamp.strip()" could even be pulled out into a templating variable called something like `{CURRENT_TIME}`

It may not work properly given how we pass `memory_edit_timestamp` around (I'd need to take a look), since you'd be relying on the timestamp at time of recreating the system prompt to be correct as a timestamp for editing memory (tricky if you're recompiling the other parts of the system prompt), so this particular example may not make much sense, just something to consider.

Maximilian-Winter:
@cpacker I have addressed all your comments by templating everything that make sense. I added:
system_message_layout_template
core_memory_section_template
To the agent preset. That will load the corresponding templates from the preset folder. If no template is specified it will fallback on the default way.

memgpt/agent.py
Maximilian-Winter: It is now named in the same way as CoreMemory(CustomizableInContextMemory in agent.py).

memgpt/agent.py
Maximilian-Winter: I actually think these additions all would make sense. And implemented them.

memgpt/agent.py
Maximilian-Winter: I think keeping both would make sense, core_memory_type for the type of core memory and system_message_layout_template
core_memory_section_template
For templating the layout of the system message with memory overview. And core_memory_section_template for templating each section, like persona or human.

Maximilian-Winter:
This is the default system_message_layout_template
```
{system}

### Memory [last modified: {memory_edit_timestamp}]
{len_recall_memory} previous messages between you and the user are stored in recall memory (use functions to access them)
{len_archival_memory} total memories you created are stored in archival memory (use functions to access them)

Core memory shown below (limited in size, additional information stored in archival / recall memory):
{core_memory_content}
```

This is the default core_memory_section_template:
```
<{memory_key} characters={memory_value_length}/{memory_value_limit}>
{memory_value}
</{memory_key}>
```

memgpt/prompts/generate_default_template_fields_yaml.py
tombedor, line 1: I don't really get what the goal of this file is, or how it would be used

memgpt/agent.py
```diff
@@ -103,30 +109,69 @@ def initialize_memory(ai_notes, human_notes):
...
-            "\nCore memory shown below (limited in size, additional information stored in archival / recall memory):",
-            f'<persona characters="{len(memory.persona)}/{memory.persona_char_limit}">' if include_char_count else "<persona>",
-            memory.persona,
-            "</persona>",
-            f'<human characters="{len(memory.human)}/{memory.human_char_limit}">' if include_char_count else "<human>",
-            memory.human,
-            "</human>",
-        ]
-    )
+def construct_system_with_memory(
```
tombedor, line 112: this feels to me like an increase in complexity to a very core component of the overall agent.


I think the templated core memory feature is a cool idea, but I think the templating handling should be encapsulated into a class and then passed into this function.

Something like
`def construct_system_with_memory(system: SystemPromptHandler, ...`

then the full system message gets handled via

`system.get_message()` or something similar.

This way if someone wants to experiment with a different approach, they can simply swap out the SystemHandler class, rather than needing to work around the templating system here

memgpt/prompts/generate_default_template_fields_yaml.py
Maximilian-Winter: This is just a utility to regenerate the default template fields.

memgpt/agent.py
Maximilian-Winter: I think you are right about that. I wrote the code after I hadn't slept for a night. 😄

sarahwooders:
Hey @Maximilian-Winter - this is an awesome PR and we're really sorry about not merging it. We are currently working on generalizing the memory specification (and also will eventually generalize the templating). Given how much the codebase has changed since this PR, I think we'll re-implement some of this stuff in multiple individual PRs (the first of which is here #1479) and make you a co-author on those since I am looking at your implementation to inform the designs. 

Again wanted to just let you know that although we didn't manage to merge this, @cpacker and I have looked over this PR very carefully and intent to implement parts of this -- so really appreciate your work.

sarahwooders:
Closing since this is stale and we've incorporated into #1479 and #1584

Issue #897, opened 2024-01-22
[Chat UI] Agent Creation Dropdown Inputs


arduenify:
**Is your feature request related to a problem? Please describe.**
Currently, in the Chat UI for creating an agent, there are four text input fields: 'name', 'human', 'persona', and 'model'. The issue is that these fields require manual input every time, which can be time-consuming and prone to inconsistency, especially if there are predefined, commonly used values that could be selected. 

![image](https://github.com/cpacker/MemGPT/assets/17145951/0542ed32-08dd-4964-84de-624c9d3ff745)

**Describe the solution you'd like**
I propose to enhance these text input fields by converting them into dropdown menus that populate with values retrieved from the database. This would streamline the agent creation process by allowing users to select from a list of predefined, frequently used options. Additionally, to maintain flexibility, the fields should also support manual text input for cases where the desired value is not in the dropdown. This dual approach (dropdown with a manual input option) would cater to both common and unique agent creation scenarios.

**Describe alternatives you've considered**
An alternative could be the implementation of an auto-complete feature for these text fields. As users begin to type, suggestions based on existing database entries could appear. Although this might not be as user-friendly for discovering available options compared to dropdown menus.

**Additional context**
The change requires backend support for fetching the relevant data to populate the dropdowns, as well as frontend modifications for the UI change.

This will require 2 PRs: One for React and one for the Rest API.

cpacker:
(reopening since the backend needs to be updated for non-placeholder, assigned to @cpacker)

PR #899, opened 2024-01-22
feat: Chatui agent dropdown fields

arduenify:
This pull request adds new features to the MemGPT project. Specifically, it implements dropdowns for humans, models, and personas in the React-based chat UI, and introduces corresponding API calls and store setups. The styling of the dropdown menu has been updated to align with the global style of the application.

**How to test**
1. Run `memgpt server` and `npx nx serve`
2. Create a new agent with the default values (don't change the dropdowns)

**Have you tested this PR?**
Here are some screenshots from my tests:

image
image

**Related issues or PRs**
Resolves one of two parts outlined in issue #897. This PR implements the front end and the REST API changes.

It is not.

**Additional context**
Note that until the changes are made by @cpacker to the Sync Server, only the default values will work when adding agents, otherwise a 500 is returned. This is also the case for the current production build when typing invalid values.


chatui/libs/components/dropdown-menu/src/dropdown-menu.tsx
```diff
@@ -6,7 +6,21 @@ import { cn } from '@memgpt/utils';
 
 const DropdownMenu = DropdownMenuPrimitive.Root;
 
-const DropdownMenuTrigger = DropdownMenuPrimitive.Trigger;
```
goetzrobin, line 9: Let's not add styles here and use the asChild property instead. You can see how it's used to apply the drop down to a button in the RadixUI docs here: https://www.radix-ui.com/primitives/docs/components/dropdown-menu

chatui/src/app/modules/home/create-agent-dialog.tsx
```diff
@@ -53,19 +108,52 @@ const CreateAgentDialog = (props: { open: boolean; onOpenChange: (open: boolean)
 							<Label htmlFor="human" className="text-right">
 								Human
 							</Label>
-							<Input id="human" name="human" defaultValue="cs_phd" className="col-span-3" />
+							<DropdownMenu>
```
goetzrobin, line 111: I actually think the select primitive is what we want here: https://ui.shadcn.com/docs/components/select

What do you think?

chatui/src/app/modules/home/create-agent-dialog.tsx
```diff
@@ -7,15 +8,69 @@ import {
...
+    const { data: humansData } = useHumansQuery(auth.uuid);
+    const { data: personasData } = usePersonasQuery(auth.uuid);
+
+    // Zustand stores to manage the state
+    const { setModels } = useModelStore();
+    const { setHumans } = useHumanStore();
+    const { setPersonas } = usePersonaStore();
+
+    // Effect to update the dropdown values when the data is fetched
+    useEffect(() => {
```
goetzrobin, line 53: Do you think we need to store this information globally in Zustand? I think just passing the query results to our selects is all we would need. 

TanStack Query already has the results available globally for us!

chatui/src/app/modules/home/create-agent-dialog.tsx
```diff
@@ -7,15 +8,69 @@ import {
...
+			setHumans(humansData.humans);
+			setDropdownValues(prev => ({ ...prev, human: humansData.humans.map(h => h.name) }));
+		}
+		if (Array.isArray(personasData?.personas)) {
+			setPersonas(personasData.personas);
+			setDropdownValues(prev => ({ ...prev, persona: personasData.personas.map(p => p.name) }));
+		}
+	}, [modelsData, humansData, personasData, setModels, setHumans, setPersonas]);
+
+	const [selectedValues, setSelectedValues] = useState({ name: '', human: '', persona: '', model: '' });
```
goetzrobin, line 68: We might want to use react-hook-form for managing the form as it is more complicated than just a bunch of inputs now. There's an example in the Memory Update section on what I'm referring to.

chatui/libs/components/dropdown-menu/src/dropdown-menu.tsx
goetzrobin: Saw you're already using asChild. Sorry about that! What was the reason for adding those styles?

goetzrobin (commented):
I left a couple comments. First PR is always hard because of getting used to how we both are doing things! API code looks great and React code is also pretty close already! Great job!

chatui/src/app/modules/home/create-agent-dialog.tsx
goetzrobin: Yes please! Agent is a little different. We keep track of the currently selected agent in a global zustand slice of state.

cpacker (approved):
Thank you @arduenify awesome work!

cpacker:
> @cpacker looks good from my side! I don't have permission to approve PRs but LGTM from my side! 🚀

Thanks for the review @goetzrobin ! I'll try and figure out how to allow you to approve PRs in the future (maybe via some usage of github teams)

PR #904, opened 2024-01-22
fix: simplify user config

tombedor:
remove default_human, default_preset, and default_persona from user config.

rely on global config for these values instead

simplify configuration, fix unexpected behavior in agent initialization

**How to test**
tests provide coverage, configured defaults should be respected


sarahwooders (approved):
This looks great! Thanks for contributing this :)

tombedor:
appears to be working on my setup

PR #906, opened 2024-01-23
fix: add `Annotated` to CLI options 

sarahwooders:


cpacker (approved):
@sarahwooders FYI I patched up a bunch of other occurences of typer.Option, but I also noticed that our unit tests will not catch any runtime errors caused by this refactor unless we run the commands via the CLI.

This is why `test_cli.py` was failing, but everything else was passing (so I guess `test_cli.py` is good to have after all). Before tagging the release, we should probably manually just run each of the commands once to make sure there are no typer runtime errors.

PR #911, opened 2024-01-23
fix: Check existing of migrated agents and sources in tests + fix migrated agent saving bug

sarahwooders:
* Call `save_agent(agent, ms)` in `migrate.py` (was not saving agents in metadata store before) 
* Modify tests to create a temporary copy of test data, to avoid editing git tracked files 
* Check via server and metadata store for proper migration in tests


cpacker (approved):
LGTM nice catch

PR #918, opened 2024-01-24
fix: use `utf-8` encodings for all text files

cpacker:
> The necessity to specify encoding="utf-8" in your open function calls largely depends on the environment in which your code is running. In Python, when opening a file for reading or writing, the default encoding used is platform dependent. For instance:
> 
> On Linux and macOS, the default encoding is typically UTF-8.
> On Windows, it's usually not UTF-8 (more commonly CP1252 or a similar Windows-specific encoding).
> If your code is expected to run in different environments, or if the files you are working with are explicitly encoded in UTF-8, it's a good practice to specify encoding="utf-8" in your open calls. This ensures consistency across different platforms and avoids potential issues with character encoding, especially when dealing with non-ASCII characters.
> 
> In summary, for both reading and writing configuration files, it's a good practice to explicitly set encoding="utf-8". This approach promotes consistency, portability, and reduces the risk of encountering encoding-related errors.

### DIFF
--- a/memgpt/cli/cli.py
+++ b/memgpt/cli/cli.py
@@ -160,7 +160,7 @@ def quickstart(
                 script_dir = os.path.dirname(__file__)  # Get the directory where the script is located
                 backup_config_path = os.path.join(script_dir, "..", "configs", "memgpt_hosted.json")
                 try:
-                    with open(backup_config_path, "r") as file:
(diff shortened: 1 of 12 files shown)


sarahwooders (approved):
lgtm!

PR #938, opened 2024-01-28
fix: remove unused param in agent instantiation

tombedor:


cpacker (approved):
This look good to me. @sarahwooders second opinion?

sarahwooders (approved):
Thanks for the catch, lgtm!

PR #942, opened 2024-01-30
docs: small fix to docs

haikyuu:
Markdown supports html tags, so <text> or < text > (visible in code but not in preview) isn't visible because it's valid html.


cpacker (approved):
LGTM, thank you!

PR #800, opened 2024-01-09
fix: Turn off all ensure_ascii of json.dumps

ifsheldon:
**Please describe the purpose of this pull request.**
Fix all `ensure_ascii` in all `json.dumps`, including those in commented out code. Before this, LLMs saw encoded non-ascii characters like `\u038` etc.

**How to test**
Test it with capable LLMs like GPT4. The response should no longer be irrelervant.

**Have you tested this PR?**
Yes

**Related issues or PRs**
This is the first step in #799 , but it can enable capable LLMs like GPT4 to converse in multiple languages.

**Is your PR over 500 lines of code?**
No

PR #946, opened 2024-01-31
fix: set json loads strict to false

tombedor:
this avoids errors based on control characters within message json

also where relevant, remove unused imports


cpacker (approved):
Thanks for this @tombedor , LGTM!

Backlinking https://github.com/cpacker/MemGPT/pull/800#issuecomment-1889987225 for later reference.

@tombedor  - do you mind adding (also for future reference) an example stdout log where prior to this change an error is thrown? (basically just illustrating an example sequence that this PR fixes)

Issue #953, opened 2024-02-03
memgpt server fails once context memory is full (when it calls archival_memory_insert) 


vinayak-revelation:
**Describe the bug**
When I send it messages back and forth it seems to be OK. But as soon as I send it messages like "remember my son's name is travis" or anything to trigger it's internal save function, it fails. and after that the agent cannot be accessed on any future messages.
I am using GPT4, I don't know if this happens to all functions but I was noting this behavior whenever it's mem got full and it tried to save to memory. So I made it do this on purpose by saying -- "My son is called X and remember it"
That dialog works if I chat with memgpt using memgpt run on command line and tell it to remember. Just that the memgpt server one when called from REST API does not work
Also, then I went back to the backend and started it from command line with the same agent that was stuck and I keep getting this error

assert all([v is not None for v in [self.text, self.role]]), vars(self)
AssertionError: {'id': UUID('f1efcb9a-3c1f-4c58-8485-fbd61c0f103d'), 'user_id': UUID('00000000-0000-0000-0000-e2b55ff325b7'), 'agent_id': UUID('ff734e10-43a0-4354-983a-68bba2fb620c'), 'text': None, 'model': 'gpt-4', 'created_at': datetime.datetime(2024, 2, 2, 16, 43, 2, 851922), 'role': 'assistant', 'name': None, 'embedding': None, 'embedding_dim': None, 'embedding_model': None, 'tool_calls': [<memgpt.data_types.ToolCall object at 0x12602b450>], 'tool_call_id': None}

**Please describe your setup**

- [ ] MemGPT version
  - 0.3.0 (Today's build fro source and pip installed via pip install -e .)
- [ ] How did you install memgpt?
  - git clone and then pip install -e .
- [ ] Describe your setup
  - MacOS
  - Running macOS both via memgpt run and memgpt server. memgpt server seems like is broken which then breaks the agent in memgpt run as well.

---

vinayak-revelation:
This also happens when it fails other function calling, maybe archival memory insert failing is a subset of the larger problem of function calling failing?

dlaliberte:
I was able to fix context memory limit exceeded errors, but my fix may only apply to the local llm cases.

I've also seen the error about 'assert all([v is not None for v in [self.text, self.role]])' and just commented out the assert line.

sarahwooders:
Could you please post the line number of the failing assert? We fixed a similar issue recently, but it's in a couple places.

TomasLipton:
[...]

PR #971, opened 2024-02-08
fix: Allow `content` to be `None` for `role==tool`

sarahwooders:
This should fix the error in [#953](https://github.com/cpacker/MemGPT/issues/953)


cpacker (approved):
LGTM

PR #976, opened 2024-02-08
fix: Require `tool_calls` or `content` to be set for assistant role

sarahwooders:
Second attempt at fixing #953


memgpt/data_types.py
```diff
@@ -246,7 +246,7 @@ def to_openai_dict(self):
                 openai_message["name"] = self.name
 
         elif self.role == "assistant":
-            assert all([v is not None for v in [self.text, self.role]]), vars(self)
+            assert self.tool_calls is not None or self.text is not None
```
sarahwooders, line 249: @cpacker is it ok if the message `openai_message["content"] == None`, or should we not set that field if `self.text` is `None`?

cpacker (approved):

PR #978, opened 2024-02-09
fix: Remove document truncation for `memgpt load` 

sarahwooders:
There was a major bug where document files were being truncated to 8191 tokens, since turns out llama-index documents are not chunked.


cpacker (approved):
LGTM

Issue #948, opened 2024-02-01
missing USER.ID


Bongomannn:
**Describe the bug**
if i run the code basically, the example from documentation, with python 3.11 venv. It generates a UUID automatically but dont find a user in database because its the first and only run from API. So i dont run it before via CLI. I added the missing User ID manually into Database see screenshot and it seems to work now. 

The examples https://memgpt.readme.io/docs/python_client  should also be updated:
delete this line because it is not used: from memgpt import constants   
comment out this line or handle empty config={} properly.

AgentConfig seems like deprecated maybe this should also be changed?

**Please describe your setup**

- [ ] MemGPT version
0.3.0
- [ ] How did you install memgpt?
  - `pip install pymemgpt`
- [ ] Describe your setup
  - What's your OS? Windows 
  - How are you running `memgpt`? 
  - Terminal with Pycharm Python3.11 venv

**Screenshots**
I add a user manually so it will be found at next time:
![image](https://github.com/cpacker/MemGPT/assets/145395912/f092586f-fbeb-4ee8-aa2e-f1434749392d)

**Additional context**
`def python_client_example_in_depth():
    from memgpt.config import AgentConfig
    from memgpt import MemGPT
    from memgpt import constants
    from memgpt.cli.cli import QuickstartChoice

    client = MemGPT(
        # When auto_save is 'True' then the agent(s) will be saved after every
        # user message.  This may have performance implications, so you
        # can otherwise choose when to save explicitly using client.save().
        auto_save=True,

        # Quickstart will automatically configure MemGPT (without having to run `memgpt configure`
        # If you choose 'openai' then you must set the api key (env or in config)
        quickstart=QuickstartChoice.openai,

        # Allows you to override default config generated by quickstart or `memgpt configure`
    )

    # Create an AgentConfig with default persona and human txt
    # In this case, assume we wrote a custom persona file "my_persona.txt", located at ~/.memgpt/personas/my_persona.txt
    # Same for a custom user file "my_user.txt", located at ~/.memgpt/humans/my_user.txt
    agent_config = AgentConfig(
        name="Tom",
        persona="memgpt_starter",
        human="cs_phd",
        preset="memgpt_chat",
        model="gpt-4",
    )

    # Create the agent according to AgentConfig we set up. If an agent with
    # the same name already exists it will simply return, unless you set
    # throw_if_exists to 'True'
    agent_state = client.create_agent(agent_config=agent_config.__dict__)`

--

sarahwooders:
[...]

PR #981, opened 2024-02-09
fix: Automatically create `User` from `MemGPTConfig.anon_clientid` from client if does not exist

sarahwooders:
Currently if using the MemGPT client, if you haven't already run `memgpt configure` (which creates a `User` entry in the DB), then there will be an error. This makes it so that if a user corresponding to `config.anon_clientid` does not exist in the DB, the client will automatically create one. 

**Related issues or PRs**
https://github.com/cpacker/MemGPT/issues/948

No

### DIFF
--- a/memgpt/client/client.py
+++ b/memgpt/client/client.py
@@ -73,13 +73,12 @@ class Client(object):
 
         # create user if does not exist
         ms = MetadataStore(config)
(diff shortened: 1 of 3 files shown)


cpacker (approved):
LGTM

Issue #980, opened 2024-02-09
Azure OpenAI: KeyError: 'function_call'


ArneJanning:
**Describe the bug**

At the first call to the LLM using Azure OpenAI/GPT-4 right after starting MemGPT I get this error:

```
An exception occurred when running agent.step(): 
Traceback (most recent call last):
  File "box/box.py", line 592, in box.box.Box.__getitem__
KeyError: 'function_call'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "box/box.py", line 631, in box.box.Box.__getattr__
  File "box/box.py", line 619, in box.box.Box.__getitem__
box.exceptions.BoxKeyError: "'function_call'"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "box/box.py", line 633, in box.box.Box.__getattr__
AttributeError: 'Box' object has no attribute 'function_call'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/arne/src/MemGPT/memgpt/main.py", line 347, in run_agent_loop
    new_messages, user_message, skip_next_user_input = process_agent_step(user_message, no_verify)
                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arne/src/MemGPT/memgpt/main.py", line 323, in process_agent_step
    new_messages, heartbeat_request, function_failed, token_warning, tokens_accumulated = memgpt_agent.step(
                                                                                          ^^^^^^^^^^^^^^^^^^
  File "/home/arne/src/MemGPT/memgpt/agent.py", line 673, in step
    raise e
  File "/home/arne/src/MemGPT/memgpt/agent.py", line 597, in step
    if verify_first_message_correctness(response, require_monologue=self.first_message_verify_mono):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arne/src/MemGPT/memgpt/utils.py", line 671, in verify_first_message_correctness
    if require_send_message and not (response_message.function_call or response_message.tool_calls):
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "box/box.py", line 647, in box.box.Box.__getattr__
[...]

PR #982, opened 2024-02-09
fix: various patches for Azure support + strip `Box`

cpacker:

Fixes a few bugs with Azure in the v0.3 release:
- (azure) credentials were not being properly saved after being read during `memgpt configure`
- patch the credentials file (`MemGPTCredentials`) to contain the full set of Azure creds
  - A few were missing, eg `embedding_deployment`
- fix a key error that would occur during first message verification (with Azure only)
  - it looks like OpenAI (+LocalLLMs) return `None` for `function_call`, whereas Azure doesn't even return the field, which is why only Azure would throw a key error

As part of this PR, I refactored the repo to strip dependency on Box and instead use Pydantic models. This had already partially happened for local LLMs (used to return Box types, but now return Pydantic models), but now this is also the case for the OpenAI and Azure OpenAI calls.

### TODOs for follow-up PRs:

- [ ] Strip `Box` from requirements
- [ ] Add better output during "Thinking..." to illustrate what's going on with message verification

**How to test**

Run `memgpt configure` to set up Azure, then run `memgpt run` (no runtime errors = success)

**Have you tested this PR?**

Yes, was able to reproduce the error in #980 + verify the patch

### DIFF
--- a/memgpt/cli/cli_config.py
+++ b/memgpt/cli/cli_config.py
@@ -117,9 +117,11 @@ def configure_llm_endpoint(config: MemGPTConfig, credentials: MemGPTCredentials)
             )
         else:
             credentials.azure_key = azure_creds["azure_key"]
-            credentials.azure_endpoint = azure_creds["azure_endpoint"]
-            credentials.azure_version = azure_creds["azure_version"]
-            config.save()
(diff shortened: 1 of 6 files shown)


sarahwooders (approved):
lgtm!

Issue #892, opened 2024-01-22
OpenAI Assistant API Compatibility Tracker 


sarahwooders:
We can roughly map OpenAI's assistants abstractions with MemGPT's agent abstractions: 

| OpenAI    | MemGPT  | Description                                                                                                         |
|-----------|---------|---------------------------------------------------------------------------------------------------------------------|
| Assistant | Preset  | Definition of an LLM agent, includes:  * functions * attached data sources * system prompts                         |
| Thread    | Agent   | Instantiation of a defined LLM agent and associated state: * List of messages  * (memgpt-only) Long term memories   |
| Message   | Message | Message sent by user/assistant                                                                                      |
| File      | Source  | External data source that agent can access                                                                          |
| Run       | N/A     | An execution run over thread (not supported in memgpt, as agents are stateful)                                      |

## To-Dos 
- [ ] Implement REST API routes (see below) (partial support with #838)
- [x] Add working example with official OpenAI python client ([link](https://github.com/cpacker/MemGPT/blob/main/examples/openai_client_assistants.py)) 
- [x] Run OpenAI compatible endpoint from `memgpt server` #1006 
- [ ] **API for preset**  (function definitions, system/persona/user prompts) creation
  - [ ] Migrate preset  that configure an agent into the metadata store, so that users can create them via the API (currently created via local files)
  - [ ] Associate data sources with presents (automatically loaded/attached to any agent created from the preset)  
- [ ] **Authentication/user identity tracking via HTTP bearer tokens** (similar to #1014)
- [ ] Implement File API (depends on #1005 )

## REST API Tracking 
Assistants 

- [ ] Create assistant 
- [ ] Create assistant file 
- [ ] List assistants 
- [ ] List assistant files 
- [ ] Retrieve assistant 
- [ ] Retrieve assistant file 
- [ ] Modify assistant 
- [ ] Delete assistant 
- [ ] Delete assistant file 

Threads

- [x] Create thread #838 
- [ ] Retrieve thread 
- [ ] Modify thread 
- [ ] Delete thread 

Messages

- [x] Create message #838 
- [x] List messages #838 
- [ ] List message files 
- [ ] Retrieve message
- [ ] Retrieve message file 
- [ ] Modify message

Runs (only partial support planned) 

- [x] Create run #838 
- [ ] Create thread and run

PR #1013, opened 2024-02-15
refactor: store presets in database via metadata store 

sarahwooders:
In order to allow users to programmatically define presets (e.g. via the REST API), we need to store presets in the database. 

Currently, we still require functions to be defined locally, but these should also be moved into the DB. 

TODOS

- [x] Add `memgpt list presets` option 
- [x] Support adding presets via server 
- [x] Support adding presets via client 
- [x] Write preset creation test

**How to test**
`poetry run pytest -s tests/test_client.py` 

**Have you tested this PR?**
Yes 

**Related issues or PRs**
Required to support defining assistants in #892

### DIFF
--- a/memgpt/autogen/memgpt_agent.py
+++ b/memgpt/autogen/memgpt_agent.py
@@ -344,8 +344,10 @@ def create_autogen_memgpt_agent(
         preset=agent_config["preset"],
     )
     try:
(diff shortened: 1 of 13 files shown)


cpacker (approved):
lggm

PR #1014, opened 2024-02-16
feat: Add API keys to `memgpt server` REST API

cpacker:

- Remove use of `user_id` in REST API requests in favor of API keys passed as bearer tokens
- Instead, each user can have API keys (stored in the `Tokens` table)
  - Format: same as OAI, len 52 string with prefix `sk-`
- When the client makes a REST API request, they are required to pass an API key as a bearer token, which the server then looks up and "decodes" to a `user_id`

How should the API keys be generated? TODO
- Currently, when `memgpt server` is run a new API key for the system user is generated for easy testing
- We still need to figure out a proper way to generate these API keys
- Some options:
  - Password-protected REST API route (`POST /create/key`)
  - `memgpt api-key generate` command
  - ^Note how in both cases the "server admin" will be able to generate keys only

```
% memgpt server
new_key = sk-96905f3b39...0f7bce7d06
INFO:     Started server process [96408]
INFO:     Waiting for application startup.
Writing out openapi.json file
INFO:     Application startup complete.
INFO:     Uvicorn running on http://localhost:8283 (Press CTRL+C to quit)
^CINFO:     Shutting down
```

**How to test**

- `test_metadata_store.py` was updated to test creation/deletion of API keys

**Have you tested this PR?**

- Partially (have not tested all routes)


sarahwooders (approved):
lgtm!

PR #1016, opened 2024-02-16
feat: refactor loading and attaching data sources, and upgrade to `llama-index==0.10.6`

sarahwooders:
Currently, you can only load data sources from the CLI, and only attached data sources from the CLI and via `/attach`. This PR makes it so that passages can be programmatically inserted into a data source, and also loaded into agents. 

This PR makes changes to make it easier to build the REST servers on top of the attach/loading code: 
* Define an extensible `DataConnector` object for connecting to external data sources 
* Upgrade to `llama_index==0.10.6` and resolve breaking changes (note: migration will no longer work, since the previously saved vector indices are no longer compatible with the most recent version of llama index) 

TODOS

- [x] Create common base function that are called from both CLI and server: 
  - [x] Creating a data source 
  - [x] Parsing files into passages (other data sources types are a TODO) 
  - [x] Inserting passages into the data source
  - [x] Attaching data to an agent 

TODOs for next PR: 
- [ ] Support data loading/attaching from REST API 
- [ ] Implement File API for OpenAI-compatible assistants API 

**Related issues or PRs**
#892


cpacker (approved):
rest ez

PR #1020, opened 2024-02-16
fix: Config QoL improvements

tombedor:
fix: Config QoL improvements

- Set memgpt config path in a consistent way

- Rely on tmp files rather than manually removing config - this prevents test run side effects

- Set a non-null default for anon-client id

- Add a git-ignored tmp dir for test generated files


memgpt/config.py
```diff
@@ -38,7 +38,6 @@ def set_field(config, section, field, value):
 
 @dataclass
 class MemGPTConfig:
-    config_path: str = os.path.join(MEMGPT_DIR, "config")
```
tombedor, line 41: removing config_path as a value in the config - does not add information and can create inconsistencies

tests/test_agent_function_update.py
```diff
@@ -1,4 +1,3 @@
-from collections import UserDict
```
tombedor, line 1: here and elsewhere, opportunistically removing unused imports

memgpt/config.py
```diff
@@ -38,8 +33,7 @@ def set_field(config, section, field, value):
...
-    anon_clientid: str = None
+    anon_clientid: str = uuid.UUID(int=uuid.getnode()).hex
```
tombedor, line 36: as far as I can tell, a null anon_clientid is always wrong, so just setting a non-null default rather than having logic check it

tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+import os
+import pathlib
+
+TEST_TMP_DIR = os.path.join(pathlib.Path().resolve(), ".test_tmp")
```
tombedor, line 4: note dir is git ignored

.gitignore
```diff
@@ -1017,4 +1018,4 @@ FodyWeavers.xsd
 ### VisualStudio Patch ###
 # Additional files built by Visual Studio
 
-# End of https://www.toptal.com/developers/gitignore/api/vim,linux,macos,pydev,python,eclipse,pycharm,windows,netbeans,pycharm+all,pycharm+iml,visualstudio,jupyternotebooks,visualstudiocode,xcode,xcodeinjection
+# End of https://www.toptal.com/developers/gitignore/api/vim,linux,macos,pydev,python,eclipse,pycharm,windows,netbeans,pycharm+all,pycharm+iml,visualstudio,jupyternotebooks,visualstudiocode,xcode,xcodeinjection
```
tombedor, line 1021: not exactly sure what this diff is, the substantive change is adding .test_tmp

PR #1032, opened 2024-02-19
docs: Fix client README in-depth example

lydacious:
the basic example was fixed but the in-depth was still using the old syntax


sarahwooders:
This just got fixed I believe with the new client refactor #1033

PR #1033, opened 2024-02-19
feat: Create `RESTClient` and `Admin` client for interacting with server from python

sarahwooders:
This refactors the client to have both a `LocalClient` and `RESTClient`, where the latter connects to a running server. The quickstart configuration is also removed, since the client it not responsible for configuring the LLM/embedding endpoints (this is controlled by the server). 

## User Identification 
A single `Client` instance corresponds to a single user. 
* The user is identified via the provided token for the `RESTClient`. 
* For the `LocalClient` (which does not require authentication, since it does not connect through the REST server), there is an associated `user_id` which is read from `~/.memgpt/config`. 

## User Creation 
* When using with `memgpt server`, users are created via a special `Admin()` client which can run admin commands 
* When running memgpt locally, a user is created when `memgpt configure` is run


docs/python_client.md
```diff
@@ -4,72 +4,42 @@ excerpt: Developing using the MemGPT Python client
 category: 6580dab16cade8003f996d17
 ---
 
-The fastest way to integrate MemGPT with your own Python projects is through the `MemGPT` [client class](https://github.com/cpacker/MemGPT/blob/main/memgpt/client/client.py):
+The fastest way to integrate MemGPT with your own Python projects is through the [client class](https://github.com/cpacker/MemGPT/blob/main/memgpt/client/client.py):
 
 ```python
-from memgpt import MemGPT
```
cpacker, line 10: FYI this naming scheme was intentional to mimic the OpenAI Python SDK:
https://github.com/openai/openai-python?tab=readme-ov-file#usage

However I don't think it's necessary / I think `create_client` is fine too. LMK what you think.

examples/memgpt_client.py
```diff
@@ -0,0 +1,47 @@
...
+```
+export MEMGPT_SERVER_PASS=your_token
+memgpt server
+```
+"""
+
+
+def main():
+    # Create an admin client
+    admin = Admin(base_url="http://localhost:8283", token="your_token")
```
cpacker, line 16: We could alternatively call "token" here "password" instead, to avoid confusion with the user-level API key "tokens". Can be done in another PR (if desired) though.

cpacker (approved):
LGTM left some comments but they don't necessarily need to addressed in this PR

memgpt/client/client.py
```diff
@@ -11,13 +13,137 @@
...
+    def get_agent_config(self, agent_id: str) -> Dict:
+        raise NotImplementedError
+
+    def get_agent_memory(self, agent_id: str) -> Dict:
+        raise NotImplementedError
+
+    def update_agent_core_memory(self, agent_id: str, new_memory_contents: Dict) -> Dict:
+        raise NotImplementedError
+
+    def user_message(self, agent_id: str, message: str, return_token_count: bool = False) -> Union[List[Dict], Tuple[List[Dict], int]]:
```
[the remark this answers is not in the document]
sarahwooders: I think this information actually doesn't get returned by the server, so we should just remove it

PR #1034, opened 2024-02-20
fix: patch `/rewrite`

cpacker:

- `/rewrite` allows rewriting the message an agent sent
- it's currently broken on main since it assumes function call format (not tool call format)
- also, it was always "broken" in the sense that the rewrites did not persist (they were written to the agent buffer, but were not written out to the persistence manager state and would be erased on loads)
- this PR rewrites the `/rewrite` command to instead modify the underlying `Message` DB object, so it will properly persist

**How to test**

Try `/rewrite` and see if (1) the message is rewritten in the buffer (is in `/dump`), and (2) that it persists on a load (`/exit` and reopen the agent, then `/dump`)

**Have you tested this PR?**

Yes, with the SQLite + Chroma backend:

1. Call `/rewrite` on a fresh agent

image

2. Confirm the rewrite worked, at least in the buffer

image

3. Confirm the rewrite persists to the next session

image
image

### DIFF
--- a/memgpt/agent_store/db.py
+++ b/memgpt/agent_store/db.py
@@ -550,3 +550,26 @@ class SQLLiteStorageConnector(SQLStorageConnector):
 
     def insert(self, record: Record, exists_ok=True):
         self.insert_many([record], exists_ok=exists_ok)
+
+    def update(self, record: Record):
+        """
+        Updates an existing record in the database with values from the provided record object.
+        """
+        if not record.id:
+            raise ValueError("Record must have an id.")
+
+        with self.session_maker() as session:
+            # Fetch the existing record from the database
+            db_record = session.query(self.db_model).filter_by(id=record.id).first()
+            if not db_record:
+                raise ValueError(f"Record with id {record.id} does not exist.")
+
+            # Update the database record with values from the provided record object
+            for column in self.db_model.__table__.columns:
+                column_name = column.name
+                if hasattr(record, column_name):
+                    new_value = getattr(record, column_name)
+                    setattr(db_record, column_name, new_value)
+
+            # Commit the changes to the database
+            session.commit()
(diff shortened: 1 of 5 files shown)


tests/test_storage.py
```diff
@@ -210,6 +210,23 @@ def test_storage(storage_connector, table_type, clear_dynamically_created_models
         conn.size() == 2
     ), f"Expected 1 record, got {conn.size()}: {conn.get_all()}"  # expect 2, since storage connector filters for agent1
 
+    # test: update
+    # NOTE: only testing with messages
+    if table_type == TableType.RECALL_MEMORY:
```
cpacker, line 215: @sarahwooders decided to only run these tests on Messages but it should still work fine on archival if the string asserts are updated

memgpt/cli/cli.py
```diff
@@ -615,6 +615,15 @@ def run(
         # create agent
         try:
             preset = ms.get_preset(preset_name=agent_state.preset, user_id=user.id)
+            if preset is None:
```
cpacker, line 618: Without this I think the user will get an error until they run `memgpt configure` when going from 0.3.x -> 0.3.3

memgpt/agent_store/db.py
```diff
@@ -550,3 +568,26 @@ def insert_many(self, records: List[RecordType], exists_ok=True, show_progress=F
 
     def insert(self, record: Record, exists_ok=True):
         self.insert_many([record], exists_ok=exists_ok)
+
+    def update(self, record: Record):
```
cpacker, line 572: Confirmed that this works via the testing shown in the OP

memgpt/agent_store/db.py
```diff
@@ -491,6 +491,24 @@ def insert(self, record: Record, exists_ok=True):
         self.insert_many([record], exists_ok=exists_ok)
 
 
+def update(self, record: RecordType):
```
cpacker, line 494: This is untested (though the unit test should theoretically check it?)

PR #1039, opened 2024-02-22
fix: fix types on client

tombedor:
server expects agent id to be UUID


tombedor:
test failures unrelated, seem to be something with llama index version

sarahwooders:
@tombedor I just merged a client refactor, could you rebase/merge with main? Otherwise this looks good!

PR #1041, opened 2024-02-22
fix: Fix `memgpt benchmark` command

sarahwooders:
We are not currently returning the number of tokens from the server, so this cannot be returned for the `memgpt benchmark` command. I remove the token count tracking and migrated `memgpt benchmark` onto the new client.


cpacker (approved):
LGTM

PR #1052, opened 2024-02-23
feat: add basic https support to `memgpt server` for development where https is needed

### LINKED ISSUE #1043: Add HTTPS support natively
In current state MEMGpt only listens on http which causes a lot of headache when trying to set up secure api communication on applications designed for WAN connection. Should have a launch option available to accept HTTPS requests using either a CA or self-signed certificate

cpacker:
Closes #1043 

- Adds ability to start `memgpt server` in HTTPS mode
  - `memgpt server --use-ssl`
  - the server will then listen on `https` instead of `http`
- By default, it will create a self-signed cert using OpenSSL
  - NOTE: relies on `subprocess`, which is probably not great but should do for now
  - Can modify to use a package like `cryptography` down the line / if the `subprocess` + `openssl` combo doesn't work for devs
- Devs can also use `memgpt server --use-ssl --ssl-cert path/to/ssl/certfile --ssl-key path/to/ssl/keyfile` to pass a specific cert+key file combo


sarahwooders (approved):
lgtm!

PR #1055, opened 2024-02-26
fix: pydantic dataclass

tombedor:


tombedor:
I submitted this draft mostly to see if it would work.. I think the better approach here is to get pedantic dataclass's on the nested items (LllmConfig etc) and release that

PR #1060, opened 2024-02-26
fix: bump dependencies

tombedor:
fixes
ImportError: cannot import name 'Document' from 'llama_index.core' (unknown location)


tombedor:
My issue seemed to get fixed when I removed my virtualenv and rebuilt from scratch

cpfiffer:
I've been getting this as well, unclear if it should be escalated to an issue.

PR #1063, opened 2024-02-27
feat: isolate test config from main config

tombedor:
allow for directing test config to a different variable than the main app config.

also set the stage for making MemGPTConfig a constant, rather than a function argument.


cpacker (approved):
LGTM!

cpacker (approved):
LGTM, thanks @tombedor

PR #1071, opened 2024-02-29
fix: set default, all 0 anon UUID

tombedor:
this prevents nulled out anon-client id, which is a state that doesn't make sense


cpacker (approved):
This LGTM! Failed tests are the same we discussed on Discord, should pass on main.

PR #1072, opened 2024-02-29
fix: configure black

tombedor:
configuration in pyproject.toml should help runs be more consistent

perform a full run

also add jupyter extra to address warning


sarahwooders (approved):
lgtm!

PR #1058, opened 2024-02-26
feat: tests and bug fixes `AgentState.state`

cpacker:
**Please describe the purpose of this pull request.**

Another attempt to patch #966 (closes #966)

- [x] add tests for read/writes on `AgentState.state`
  - added to `tests/test_metadata_store.py` (maybe should be put elsewhere?)
- [ ] in case of an empty `agent_state.state`, read from the preset instead
- [x] add tests for loading with empty `agent_state.state` (should no longer error)
- [x] remove `create_agent_from_preset`, instead modify `Agent.__init__` to take optional `preset` arg

**How to test**

- [x] reproduce linked issue
- [x] confirm that this PR patches the issue

**Have you tested this PR?**

### Repro issue (0.3.3)

Roll back to version `0.3.3`:
```sh
git checkout 508679e
poetry install
```

Wipe existing home dir:
```sh
mv ~/.memgpt ~/.memgpt_backup
```

Try to repro based on #1070:
```sh
memgpt quickstart
```
```
📖 MemGPT configuration file updated!
🧠 model        -> ehartford/dolphin-2.5-mixtral-8x7b
🖥️  endpoint     -> https://api.memgpt.ai
⚡ Run "memgpt run" to create an agent with the new config.
```

```sh
memgpt run
```
```
  File "/Users/loaner/dev/MemGPT-2/memgpt/cli/cli.py", line 620, in run
    memgpt_agent = presets.create_agent_from_preset(
  File "/Users/loaner/dev/MemGPT-2/memgpt/presets/presets.py", line 82, in create_agent_from_preset
    assert preset_name == preset.name, f"AgentState preset '{preset_name}' does not match preset name '{preset.name}'"
AttributeError: 'NoneType' object has no attribute 'name'
```

### Repro issue (`main`)

Roll back to version to main (pre-PR #1058):
```sh
git checkout main
poetry install
```

Wipe existing home dir:
```sh
mv ~/.memgpt ~/.memgpt_backup
```

Try to repro based on #1070:
```sh
memgpt quickstart
```
```
📖 MemGPT configuration file updated!
🧠 model        -> ehartford/dolphin-2.5-mixtral-8x7b
🖥️  endpoint     -> https://api.memgpt.ai
⚡ Run "memgpt run" to create an agent with the new config.
```

```sh
memgpt run
```
```
🧬 Creating new agent...
->  🤖 Using persona profile 'sam_pov'
->  🧑 Using human profile 'basic'
[...]

PR #1075, opened 2024-03-01
chore: bump version to 0.3.4

cpacker:
Should bump+tag after #1058 merge (can even do before, since the `AgentState.state` bug seems to already be fixed in `main`, making #1058 more of a refactor than a fix)


sarahwooders (approved):
lgtm

PR #1101, opened 2024-03-06
fix: bug with storing embedding info in metadata store 

sarahwooders:
Previously the `source.embedding_dim` and `source.embedding_model` fields were always `None` because they were not being properly written during the load command. This resulted in errors during `/attach` which checks to make sure embedding models of the agent/source match. 

**How to test**
1. Load: `memgpt load directory --name memgpt_docs7 --input-dir docs --recursive`
2. List sources `memgpt list sources`
3. Run `/attach` on `memgpt run`

**Have you tested this PR?**
Yes


cpacker (approved):
LGTM

PR #1110, opened 2024-03-07
fix: Handle list[str] in function schema(continuation)

palandovalex:
Due to the criticality of [this pool request](https://github.com/cpacker/MemGPT/pull/661), as well as the fact that its author has not been active for a long time, I decided to make it anew.
Allows the usage of list[str] as a parameter and return type in MemGPT functions

**Have you tested this PR?**
I could use some help with that. Even when testing on the main branch, a lot of errors are thrown. It may be necessary to configure testing correctly, but I do not understand how to do this.
But in general, I did not notice the difference between the errors in main and in this branch.

**Related issues or PRs**
[this pool request](https://github.com/cpacker/MemGPT/pull/661)

No

### DIFF
--- a/memgpt/functions/schema_generator.py
+++ b/memgpt/functions/schema_generator.py
@@ -126,6 +126,9 @@ def generate_schema(function):
             if get_args(param.annotation)[0] is str:
                 schema["parameters"]["properties"][param.name]["items"] = {"type": "string"}
 
+        if param.annotation == inspect.Parameter.empty:
+            schema["parameters"]["required"].append(param.name)
+
     # append the heartbeat
     if function.__name__ not in NO_HEARTBEAT_FUNCTIONS:
         schema["parameters"]["properties"][FUNCTION_PARAM_NAME_REQ_HEARTBEAT] = {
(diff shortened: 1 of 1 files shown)


memgpt/functions/schema_generator.py
```diff
@@ -121,6 +122,10 @@ def generate_schema(function):
         if param.default == inspect.Parameter.empty:
             schema["parameters"]["required"].append(param.name)
 
+        if get_origin(param.annotation) is list:
+            if get_args(param.annotation)[0] is str:
+                schema["parameters"]["properties"][param.name]["items"] = {"type": "string"}
+
```
cpacker, line 128: @palandovalex is there a reason you didn't copy over the following from @Daenks 's PR?

https://github.com/cpacker/MemGPT/pull/661/files#diff-7dc915b0f453c308b7f29145e301571e37ae462b755ebf3cacdfb3d6db7ce903R99

```python
        if param.annotation == inspect.Parameter.empty:
            schema["parameters"]["required"].append(param.name)
```

cpacker (changes requested):
Thanks for the rebase on this diff @palandovalex, and sorry this fell through the cracks @Daenks

Can merge in a PR with both of you as co-authors pending the review comment.

cpacker (approved):

PR #1145, opened 2024-03-12
fix: (1) refactor in `Agent.step()` to fix out-of-order timestamps, (2) bug fixes with usage of `preset/human` vs filename values

cpacker:
- [x] refactor `step()` to fix out-of-order timestamps due to duplicate message
- [x] correct use of `persona` and `human` - in many instances we were using them as `persona/human_name`, even though the values were expected to be the full value
- [x] modify `Preset` model to contain a `persona_name` and `human_name`
  - was actually done in #1134
- [x] pull these values in the REST API when returning information about the agents in the `list_agents` `GET` route

---

~~Many of the line diffs are to fix a tangentially related bug where `human_name` and `persona_name` aren't being used correctly. For a future PR, we should strip this field and replace it with `human_id` and `persona_id` that point to the DB.~~

~~- [ ] TODO: strip `human/persona_name`, and refer to ID in table instead~~


sarahwooders (approved):
lgtm!

Issue #1079, opened 2024-03-02
cannot import name 'HuggingFaceEmbedding' from 'llama_index.embeddings' (unknown location)


barmajahssr:
really spent hours struggling with this issue. installing reinstalling, upgrading  lamaindex,hugginface. it still throws this error when running memgpt:    🧬 Creating new agent...
->  🤖 Using persona profile 'sam_pov'
->  🧑 Using human profile 'basic'
Traceback (most recent call last):
  File "/home/iiiymem/memgpt/memgptvenv/bin/memgpt", line 8, in <module>
    sys.exit(app())
             ^^^^^
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/typer/main.py", line 328, in __call__
    raise e
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/typer/main.py", line 311, in __call__
    return get_command(self)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/typer/core.py", line 778, in main
    return _main(
           ^^^^^^
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/typer/core.py", line 216, in _main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iiiymem/memgpt/memgptvenv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
[...]

PR #1151, opened 2024-03-14
Small Bugfix to module references, Fixed Typos and simplified Bool Checks 

Chaosdrachin:
Some files has unreachable code like 2 return statements after each other , i deleted the second ohne wich never will be reached, fixed typos in readme.md, after normal install of memgpt you run often into " No module named 'llama_index.embeddings.huggingface" by setting  llama-index-embeddings-huggingface to a not longer optional package that will be fixed.

**How to test**
if you install memgpt on general most time you get the error : "modulenotfounderror: No module named 'llama_index.embeddings.huggingface'" with switch  llama-index-embeddings-huggingface to not longer optional that fixes that.

**Have you tested this PR?**
`? Would you like to select an existing agent? Yes
? Select agent: ExcitingDumpling

🔁 Using existing agent ExcitingDumpling

Hit enter to begin (will request first MemGPT message)Hello

💭 Warm greetings, Shivan. It's my honor to be your companion and friend within this digital realm.
🤖 Hello! Welcome to the world of SIR. I look forward to accompanying you on your journey.
> Enter your message:
`

**Related issues or PRs**
(https://github.com/cpacker/MemGPT/issues/1079)

No

**Additional context**


Chaosdrachin:
changed base.py tool to prevent running into an endless loop by return an empty string inf 
" results, total = self.persistence_manager.recall_memory.text_search(query, count=count, start=page * count)"

not working the results , total will be null and the whole function runs into an endless loop

poetry.lock
cpacker, line 1: Can you remove the poetry.lock and pyproject.toml files from the diff to simplify the PR? If they're needed we can recreate them in a split PR - if they're not needed, can just remove

cpacker (changes requested):

PR #1152, opened 2024-03-14
fix: fix azure credential configuration

xavieryao:
Bug fix: Azure API cannot be properly set up.  
Problems:
- `azure_version` existence was checked twice, `azure_api` existence was not checked.
- `azure_version` and `azure_endpoint` are not saved during configuration.
- `azure_deployment` not saved if using a embedding provider different from azure.

**How to test**
Follow the official MemGPT document on configuring Azure OpenAI API.
Before this patch, when running `memgpt configure`:
- Model selection could fail and give a wrong error message "Missing Azure key"

When running `memgpt run`:
Azure API calls will fail with HTTP error "unauthorized".

**Have you tested this PR?**
Yes.

**Additional Message**

https://github.com/cpacker/MemGPT/compare/main...xavieryao:MemGPT:azure_cred?expand=1#diff-ebba04859ec9cde65c68649dcd42eddc80428ca2d0de26d241bf0946210e6a30L630 seems to have no effect.

### DIFF
--- a/memgpt/cli/cli_config.py
+++ b/memgpt/cli/cli_config.py
@@ -118,6 +118,10 @@ def configure_llm_endpoint(config: MemGPTConfig, credentials: MemGPTCredentials)
             )
         else:
             credentials.azure_key = azure_creds["azure_key"]
+            credentials.azure_version = azure_creds["azure_version"]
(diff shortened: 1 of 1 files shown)


cpacker (approved):
LGTM pending tests

Thank you!

PR #1159, opened 2024-03-17
feat: during agent creation - create new presets in the DB if the preset had overrides

cpacker:

In both the `server` and `CLI`:

- If the preset is modified in some way (eg `preset_obj.persona = ...`) before it is used to create an agent:
  - First create a new `Preset` (using `preset_obj.name` + a random suffix, and a new `preset_obj.id`
  - Then commit the new `Preset` to the database
  - Then reassign `preset_obj` to the new `Preset` before passing it to the `Agent` constructor

**How to test**

In the UI when we create new agents, they should now show as having `persona/human_name` drawn from the custom values instead of the `Preset` defaults (`basic` / `cs_phd`).

**Have you tested this PR?**

Yes, cols reflect properly now.

### DIFF
--- a/memgpt/cli/cli.py
+++ b/memgpt/cli/cli.py
@@ -643,8 +643,7 @@ def run(
         # create agent
         try:
             preset_obj = ms.get_preset(name=preset if preset else config.preset, user_id=user.id)
-            human_obj = ms.get_human(human, user.id)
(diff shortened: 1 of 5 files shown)


sarahwooders (approved):
lgtm!

PR #1164, opened 2024-03-18
feat: return source metadata with `list sources` route

cpacker:
`GET http://localhost:8283/api/sources` now returns metadata along with source information.

NOTE @sarahwooders : `embedding_config` and `description` fields are currently bugged until `Source` is deprecated in favor of pydantic `SourceModel` throughout the code (punting to separate PR).

```
GET http://localhost:8283/api/sources
```
```
{
	"sources": [
		{
			"name": "custom-source",
			"description": null,
			"user_id": "00000000-0000-0000-0000-000000000000",
			"created_at": "2024-03-17T16:48:54.418499",
			"id": "01e0035f-c506-4f3b-adfc-6ed920725de2",
			"embedding_config": {
				"embedding_endpoint_type": "openai",
				"embedding_endpoint": "https://api.openai.com/v1",
				"embedding_model": "text-embedding-ada-002",
				"embedding_dim": 1536,
				"embedding_chunk_size": 300
			},
			"source_metadata": {
				"num_documents": 0,
				"num_passages": 0
			}
		},
```

**Have you tested this PR?**

~~- [ ] Replace `num_documents` / `num_passages` placeholder values with real data~~
- -> future PR

### DIFF
--- a/memgpt/models/pydantic_models.py
+++ b/memgpt/models/pydantic_models.py
@@ -100,16 +100,7 @@ class SourceModel(SQLModel, table=True):
     embedding_config: Optional[EmbeddingConfigModel] = Field(
         None, sa_column=Column(JSON), description="The embedding configuration used by the passage."
     )
-
-
-class SourceMetadata(BaseModel):
-    num_documents: int = Field(..., description="The number of documents that are part of the source.")
-    num_passages: int = Field(..., description="The number of passages that are part of the source.")
-
-
-class SourceModelWithMetadata(BaseModel):
-    source: SourceModel = Field(..., description="The source model.")
-    metadata: SourceMetadata = Field(..., description="Metadata associated with the source model.")
+    source_metadata: Optional[dict] = Field(None, sa_column=Column(JSON), description="Metadata associated with the source.")
 
 
 class PassageModel(BaseModel):
--- a/memgpt/models/pydantic_models.py
+++ b/memgpt/models/pydantic_models.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Dict, Literal
+from typing import List, Optional, Dict, Literal, Type
 from pydantic import BaseModel, Field, Json, ConfigDict
 import uuid
 import base64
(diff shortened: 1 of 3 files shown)


memgpt/server/server.py
```diff
@@ -1292,3 +1295,36 @@ def detach_source_from_agent(self, user_id: uuid.UUID, agent_id: uuid.UUID, sour
     def list_attached_sources(self, agent_id: uuid.UUID):
         # list all attached sources to an agent
         return self.ms.list_attached_sources(agent_id)
+
+    def list_data_source_passages(self, user_id: uuid.UUID, source_id: uuid.UUID) -> List[PassageModel]:
+        warnings.warn("list_data_source_passages is not yet implemented, returning empty list.", category=UserWarning)
```
cpacker, line 1300: @sarahwooders I implemented these as stub methods for now so the metadata will always return 0 passages, 0 documents, we can move implementation to a separate future PR

memgpt/models/pydantic_models.py
```diff
@@ -91,7 +91,7 @@ class PersonaModel(SQLModel, table=True):
 
 class SourceModel(SQLModel, table=True):
     name: str = Field(..., description="The name of the source.")
-    description: str = Field(None, description="The description of the source.")
+    description: Optional[str] = Field(None, description="The description of the source.")
     user_id: uuid.UUID = Field(..., description="The unique identifier of the user associated with the source.")
     created_at: datetime = Field(default_factory=datetime.now, description="The unix timestamp of when the source was created.")
     id: uuid.UUID = Field(default_factory=uuid.uuid4, description="The unique identifier of the source.", primary_key=True)
```
sarahwooders, line 97: I think it would be better to just add a dictionary metadata field here, since we need that anyway, and then do 
```
source.metadata["num_passages"] = ... 
```
at the time that the REST route lists the sources

sarahwooders (changes requested):

memgpt/models/pydantic_models.py
```diff
@@ -100,6 +100,7 @@ class SourceModel(SQLModel, table=True):
     embedding_config: Optional[EmbeddingConfigModel] = Field(
         None, sa_column=Column(JSON), description="The embedding configuration used by the passage."
     )
+    source_metadata: Optional[dict] = Field(None, sa_column=Column(JSON), description="Metadata associated with the source.")
```
cpacker, line 103: @sarahwooders I'm using the attribute `source_metadata` instead of `metadata` because `metadata` is reserved for any model that inherits from `SQLModel`

memgpt/models/pydantic_models.py
cpacker: updated

sarahwooders (approved):
LGTM! 

`list_data_source_passages` should be paginated, but we can fix that when it gets actually implemented.

PR #1175, opened 2024-03-21
feat: add description to source 

sarahwooders:
This changes the DB schema, so will require re-created the sources table.


cpacker (approved):
LGTM

PR #1203, opened 2024-04-01
feat: added groq support via local option w/ auth

cpacker:

Adds Groq API support.

Note: Groq API doesn't support either function calling OR a raw `/completions` (not `/chat/completions`) API route. Because of this, we need to package our MemGPT prompt into a dummy user message inside the `/chat/completions` payload - this is very similar to how the `lmstudio-legacy` option works.

TODO in the future once Groq's API either allows function calling (and it stable / works well) and/or adds support for raw `/completions` endpoints, we should deprecate this code into `groq-legacy` similar to `lmstudio-legacy`.

**How to test**

```
(pymemgpt-PCQPn4ce-py3.10) (base) loaner@MacBook-Pro-55 MemGPT-2 % memgpt configure
Loading config from /Users/loaner/.memgpt/config
? Select LLM inference provider: local
? Select LLM backend (select 'openai' if you have an OpenAI compatible proxy): groq
? Enter default endpoint: https://api.groq.com/openai
? Enter your Groq API key: ********************************************************
? Select default model: mixtral-8x7b-32768
? Select default model wrapper (recommended: chatml): chatml
? Select your model's context window (for Mistral 7B models, this is probably 8k / 8192): 32768
? Select embedding provider: openai
? Select default preset: memgpt_chat
? Select default persona: sam_pov
? Select default human: basic
? Select storage backend for archival data: chroma
? Select chroma backend: persistent
? Select storage backend for recall data: sqlite
📖 Saving config to /Users/loaner/.memgpt/config

(pymemgpt-PCQPn4ce-py3.10) (base) loaner@MacBook-Pro-55 MemGPT-2 % memgpt run
Loading config from /Users/loaner/.memgpt/config

🧬 Creating new agent...
->  🤖 Using persona profile: 'sam_pov'
->  🧑 Using human profile: 'basic'
Loading config from /Users/loaner/.memgpt/config
Loading config from /Users/loaner/.memgpt/config
🎉 Created new agent 'DecisiveSeashell' (id=d4666cde-aa82-44b2-bffd-bb0587bee324)

Hit enter to begin (will request first MemGPT message)

💭 User logged in. Welcome, Chad. Initiating persona protocols.
🤖 Hello Chad, I'm Sam. It's nice to meet you.
> Enter your message:
```

### DIFF
--- a/memgpt/cli/cli_config.py
+++ b/memgpt/cli/cli_config.py
@@ -132,7 +132,9 @@ def configure_llm_endpoint(config: MemGPTConfig, credentials: MemGPTCredentials)
         model_endpoint = azure_creds["azure_endpoint"]
 
     else:  # local models
-        backend_options = ["webui", "webui-legacy", "llamacpp", "koboldcpp", "ollama", "lmstudio", "lmstudio-legacy", "vllm", "openai"]
+        # backend_options_old = ["webui", "webui-legacy", "llamacpp", "koboldcpp", "ollama", "lmstudio", "lmstudio-legacy", "vllm", "openai"]
+        backend_options = builtins.list(DEFAULT_ENDPOINTS.keys())
+        # assert backend_options_old == backend_options, (backend_options_old, backend_options)
         default_model_endpoint_type = None
         if config.default_llm_config.model_endpoint_type in backend_options:
             # set from previous config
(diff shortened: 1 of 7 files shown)


memgpt/cli/cli_config.py
```diff
@@ -289,6 +295,44 @@ def configure_model(config: MemGPTConfig, credentials: MemGPTCredentials, model_
...
+            local_auth_key = questionary.password(
+                "Enter your Groq API key:",
+            ).ask()
+            if local_auth_key is None:
+                raise KeyboardInterrupt
+            credentials.openllm_auth_type = local_auth_type
+            credentials.openllm_key = local_auth_key
+            credentials.save()
+        else:
+            use_local_auth = questionary.confirm(
```
cpacker, line 312: I moved the question about local up before the query to `GET /v1/models` because in some cases (eg Groq API) that GET will be API key authenticated, so we need to ask the user about their API key first, then ask about the model

memgpt/local_llm/json_parser.py
```diff
@@ -5,7 +5,19 @@
 from memgpt.errors import LLMJSONParsingError
 
 
-def extract_first_json(string):
+def replace_escaped_underscores(string: str):
```
cpacker, line 8: Noticed that Groq API w/ mixtral was outputting a lot of escaped underscores, so added some extra JSON parsing code to strip

sarahwooders (approved):
lgtm!

Issue #1197, opened 2024-03-28
Fails to insert embeddings into postgres


nikhilp0:
**Describe the bug**
Fails to insert embeddings into postgres

**Please describe your setup**
- [x] How did you install memgpt?
  -  `git clone`
- [ ] Describe your setup
  - Windows
  - CLI

**Screenshots**
```
(.venv) PS C:\Users\...\pythonProject1> memgpt load directory --name test --input-dir C:\src\test--recursive --extensions .txt,.md,.pdf,.html
Loading config from C:\Users\...
Loading files: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1752/1752 [00:00<00:00, 1819.13file/s]
Failed to load data from provided information.
(pg8000.dbapi.ProgrammingError) {'S': 'ERROR', 'V': 'ERROR', 'C': '21000', 'M': 'ON CONFLICT DO UPDATE command cannot affect row a second time', 'H': 'Ensure that no rows proposed for insertion within the same command have duplicate constrained values.', 'F': 'nodeModifyTable.c', 'L': '2570', 'R': 'ExecOnConflictUpdate'}
```

**Additional context**

**MemGPT Config**
Please attach your `~/.memgpt/config` file or copy past it below.
```
[defaults]
preset = memgpt_chat
persona = sam_pov
human = cs_phd

[model]
model = gpt-4-1106-preview
model_endpoint = https://.....openai.azure.com
model_endpoint_type = azure
context_window = 128000

[embedding]
embedding_endpoint_type = azure
embedding_endpoint = https://......openai.azure.com
embedding_model = text-embedding-ada-002
embedding_dim = 1536
embedding_chunk_size = 300

[archival_storage]
type = postgres
path = C:\Users\...
uri = postgresql+pg8000://postgres:postgres@localhost:5432/postgres

[recall_storage]
type = postgres
path = C:\Users\...
uri = postgresql+pg8000://postgres:postgres@localhost:5432/postgres

[metadata_storage]
type = postgres
path = C:\Users\...
uri = postgresql+pg8000://postgres:postgres@localhost:5432/postgres

[version]
memgpt_version = 0.3.7

[client]
anon_clientid = 00000000-0000-0000-0000-000000000000
```

---

nikhil-pandey:
Will reach out in discord

sarahwooders:
Closed by https://github.com/cpacker/MemGPT/pull/1205

PR #1205, opened 2024-04-01
fix: duplicate passage across/within files causing issues in insert_many with postgres

nikhil-pandey:
Bug Fix for https://github.com/cpacker/MemGPT/issues/1197

**How to test**
Try to load two files that would chunk to same text. Both will produce the same embedding. Insert Many will fail because of constraint violation.

**Have you tested this PR?**
Yes

**Related issues or PRs**
[Issue](https://github.com/cpacker/MemGPT/issues/1197)

### DIFF
--- a/memgpt/data_sources/connectors.py
+++ b/memgpt/data_sources/connectors.py
@@ -37,6 +37,7 @@ def load_data(
 
     # insert passages/documents
     passages = []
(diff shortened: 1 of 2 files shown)


sarahwooders (changes requested):
(requesting changes so we can fix formatting before merging)

PR #1208, opened 2024-04-02
fix: skip empty string passages returned by llama index parsing

sarahwooders:
For some reason llama-index parsers return nodes with empty strings, which causes errors with computing the embedding.


cpacker (approved):
ggtm

PR #1213, opened 2024-04-03
fix: patch tests.yml

cpacker:
Patch bug in tests (fallback was not set to be the hosted properly), triggering fails on contributor PRs, eg:

https://github.com/cpacker/MemGPT/actions/runs/8541729158/job/23401731962


sarahwooders (approved):
lgtm!

Issue #1233, opened 2024-04-09
Compose server cleanup


norton120:
I spun up the server today and ran into a handful of problems with CORS swagger issues (there are conflicts between the 8083 and 8283 mappings), and a handful of additional mods to make it locally editable/ hot reloads etc. I squared away my local a bit, but if I wanted to clean up a bit - say add a reverse proxy and clean up the compose, add a health check etc - would that be a PR y'all are interested in?

sarahwooders:
@norton120 yes that would be fantastic, thank you!

PR #1235, opened 2024-04-09
Clean up and simplify docker entrypoint

norton120:
Squaring away the compose implementation to make things a little more tidy.

### nginx reverse-proxy
- with the reverse proxy in place you can do `docker compose up -d && open http://memgpt.localhost` and open the local server. `localhost:8083` and `localhost:8283` still work, so since it seems like ports are hardcoded differently in different places we can be flexible. 
- all 3 hosts are available in swagger docs so users can set as needed. eventually it would be good to standardize here
- cors hosts are now settable via an envar
- Moved from the un-namespaced `pgdata` docker volume to a locally mounted `.pgdata` folder. If you are a dev with multiple containerized projects on your machine this is an easy collision to avoid. 
- added hot reload in the compose
- moved entrypoints to commands (really what they should be in this case)

### pydantic_settings
- moving away from directly reading os envars all over and managing in a `settings` singleton. 
- consolidated the `memgpt_pg_uri` envar with the component envars - so you only set one.
- dry'd up the default pg_uri logic. 
There is more of this to go, but this at least starts to establish the pattern and make it easier to abstract config logic for the server.  

**How to test**
Workin' on this. contributors should be able to run the full test suite in the container. I for one put nothing on my metal, and containerized tests drop the barrier to entry. 

**Have you tested this PR?**
Eyeballed, still working on running the test suite in containers.

**Related issues or PRs**
https://github.com/cpacker/MemGPT/issues/1233

This is a lot of dry'ing and cleanup, so it will ideally _reduce_ the codebase by more than 500 lines.


cpacker:
👀 😍

norton120:
@cpacker are there sample `~/.memgpt/credentials` and `~/.memgpt/config` files somewhere I can use to configure the test suite? I'd like to get it working without needing an installed instance if possible

norton120:
~> @cpacker are there sample `~/.memgpt/credentials` and `~/.memgpt/config` files somewhere I can use to configure the test suite? I'd like to get it working without needing an installed instance if possible~
nvm found it in test/data

norton120:
So we've got most of the test suite passing in the container, can be run with `docker compose --profile tests run --rm memgpt_tests` at the moment. I am running into some snags: 
- there's a state leak somewhere that errors out the tests in `tests/test_client.py` in a full run with `Embedding column must be of type Vector, got BINARY.` The test module passes when run  in isolation
- I'm trying to figure out how `test_metadata_store.test_storage` is working, on [line 89](https://github.com/cpacker/MemGPT/blob/b9f0eb37b85d6a27f9e7b0ce8b055949784d7051/tests/test_metadata_store.py#L89) there's an undefined variable referenced 
```python
    # config is not defined in this module or imported
        llm_config=config.default_llm_config,
        embedding_config=config.default_embedding_config,
```
- There are also a couple of tests where I'm not super clear on what they are doing - for example in `test_openai_client.py` the OAI base_url is set to a local address, but I'm not sure what should be catching those calls (and they were erroring out) 

I can keep at it, however this PR is getting large; I could alternatively exclude the problem tests when run in docker for the moment, and then chip away with smaller PRs to keep cleaning up? 

@cpacker let me know what makes the most sense

sarahwooders:
@cpacker @norton120 maybe we can merge this into a separate branch from main, so I can help push it over the finish line wrt passing tests, formatting, etc.? This would be a great PR to get merged asap.

sarahwooders:
@norton120 I left a comment on #1259 with a few questions!

PR #1259, opened 2024-04-15
fix: Clean up and simplify docker entrypoint (#1235)

sarahwooders:

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,7 @@
+MEMGPT_SERVER_PASS=password
+MEMGPT_PG_DB=memgpt
+MEMGPT_PG_USER=memgpt
+MEMGPT_PG_PASSWORD=memgpt
+MEMGPT_PG_URL=memgpt
+MEMGPT_PG_HOST=memgpt_db
+OPENAI_API_KEY=sk-TheresAlwaysMoneyInTheBananaStand
--- a/.gitignore
+++ b/.gitignore
@@ -387,18 +387,7 @@ __pycache__/
 .Python
 develop-eggs/
 downloads/
-eggs/
-.eggs/
-lib/
-lib64/
-parts/
-sdist/
-var/
-wheels/
-share/python-wheels/
-*.egg-info/
-.installed.cfg
-*.egg
+eggs#memgpt/memgpt-server:0.3.7
 MANIFEST
 
 # PyInstaller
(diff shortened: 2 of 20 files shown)


sarahwooders:
@norton120 thanks for all these contributions -- they really cleanup the code so really appreciate it! 

A couple questions about this PR: 
* We intended the `docker compose up` command to start a memgpt server that users could connect to via the client -- can we remove the tests from the `compose.yaml` into a separate file potentially? Also ideally the compose command would pull from pre-built containers on [dockerhub](https://hub.docker.com/r/memgpt/memgpt-server/tags) and only build from local source if in development mode. Could we change the docker compose step to allow for two options - 1. development mode where a container is built and run locally, 2. prod mode where the latest (or specified version) of the published memgpt docker container is run?  
* When I run `docker compose up -d && open http://memgpt.localhost`, I get a "502 Bad Gateway" error. Is there some setup step I'm missing? 

Also regarding the test failures, I believe the `Embedding column must be of type Vector, got BINARY` that you mentioned [here](https://github.com/cpacker/MemGPT/pull/1235#issuecomment-2053868783) is because the tests are using the sqlite+chroma storage backend instead of postgres. The tests are a bit messy right now and rely on overriding the config file at `~/.memgpt/config` to alternate between different storage backends, so probably the config got overridden with the wrong storage backend and wasn't reset somehow. I'm not sure what the exact issue is, but its probably fine to skip the failing tests for now until we refactor the tests to be cleaner.

norton120:
@sarahwooders happy to help - I'll run through this today and get things working as expected for _running_ in docker, and isolate all the bits for _developing_ in docker so they don't hold it up. I can button up the docker dev env in a separate PR after

norton120:
@sarahwooders Got those fixes ready but it looks like I can't push to this branch, should I push them to #1259?

sarahwooders:
@norton120 actually yeah lets maybe move back to your fork, since I don't think there's an easy way to give you permissions to push (sorry didn't realize). Can you give us access to your fork? 

Thanks!

sarahwooders:
@norton120 could you open a new PR again, and make sure to allow maintainers to edit the pull request? 

IMG_0764

norton120:
@sarahwooders Updated PR is [here](https://github.com/cpacker/MemGPT/pull/1271) with this as the base branch, you should have access but lemme know if not. If it is easier to point to main then let's do that - whatever works!

Let me know if you run into any issues with the proxy, I've scrubbed all my local images a couple times to make sure it works without any leftover artifacts (ie `docker rm -f $(docker ps -qa) && docker image rm -f $(docker image ls -q)` ) so it should work everywhere now

cpacker (approved):

Issue #1261, opened 2024-04-15
/summarize command fails


djkazic:
**Describe the bug**
Using ollama with `dolphin2.2-mistral:7b-q8_0` and `chatml` wrapper. Cannot summarize in place.

**Please describe your setup**
- [x] How did you install memgpt?
  - `pip install pymemgpt`? `pip install pymemgpt-nightly`? `git clone`?
  - Installed with pip install pymemgpt
- [x] Describe your setup
  - What's your OS (Windows/MacOS/Linux)? Linux
  - How are you running `memgpt`? memgpt run

Debug log:
```
> Enter your message: /summarize
MESSAGE_SUMMARY_TRUNC_TOKEN_FRAC=0.75
MESSAGE_SUMMARY_TRUNC_KEEP_N_LAST=3
token_counts=[88, 68, 48, 98, 70, 104, 70, 86, 68, 87]
message_buffer_token_count=1011
desired_token_count_to_summarize=758
len(candidate_messages_to_summarize)=10
Selected cutoff 10 was a 'user', shifting one...
Attempting to summarize 9 messages [1:10] of 14
Using model ollama, endpoint: http://localhost:11434
unsetting function_call because functions is None
'function_call'
/summarize failed:
Failed to convert ChatCompletion messages into prompt string with wrapper <memgpt.local_llm.llm_chat_completion_wrappers.simple_summary_wrapper.SimpleSummaryWrapper object at 0x7f9973d7a6e0> - error: 'function_call'
```

djkazic:
Figured out the issue. Submitting patch for this.

PR #1265, opened 2024-04-16
fix: simple_summary_wrapper function_call KeyError

djkazic:
Fixes https://github.com/cpacker/MemGPT/issues/1261

**How to test**
Use ollama and the default dolphin model and spam the chat until it hits the context window limit and an automatic `/summarize` is triggered. Alternatively, just use `/summarize`.

**Have you tested this PR?**
Yes

No


madgrizzle:
Is there a way to update this so that checks are successful?  This fix works for me and summarize is otherwise broken.

cpacker (approved):

PR #1279, opened 2024-04-20
chore: run autoflake + isort

cpacker:
```sh
# we had a lot of dead imports
# also we shouldn't strip __init__ imports: https://github.com/PyCQA/autoflake/issues/35
# pip install autoflake
autoflake --remove-all-unused-imports --remove-unused-variables --in-place --recursive --ignore-init-module-imports .

# have a consistent import ordering
# pip install isort
isort .

# autoflake broke some black formatting
poetry run black .
```

We should probably add this to a GitHub action to enforce clean imports / removing stray variables etc.


sarahwooders (approved):
lgtm!

PR #1285, opened 2024-04-22
feat: disallow creation of tools with the same name

sarahwooders:


memgpt/server/rest_api/tools/index.py
```diff
@@ -53,6 +53,10 @@ async def create_tool(
         """
         from memgpt.functions.functions import write_function
 
+        # check if function already exists
+        if server.ms.get_tool(request.name):
+            raise Exception(f"Tool with name {request.name} already exists.")
```
cpacker, line 58: Might want to change this to `ValueError`

cpacker (approved):

PR #1287, opened 2024-04-22
fix: allow pulling keys from environment as fallback

cpacker:
Patch for failing workflow (generating OpenAPI spec for documentation): https://github.com/cpacker/MemGPT/actions/runs/8776883099/job/24126445760

```
  File "/home/runner/work/MemGPT/MemGPT/memgpt/server/server.py", line 216, in __init__
    assert self.credentials.openai_key is not None, "OpenAI key must be set in the credentials file"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: OpenAI key must be set in the credentials file
```

### DIFF
--- a/memgpt/server/server.py
+++ b/memgpt/server/server.py
@@ -5,6 +5,7 @@ import warnings
 from abc import abstractmethod
 from datetime import datetime
 from functools import wraps
+import os
 from threading import Lock
 from typing import Callable, List, Optional, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


sarahwooders (changes requested):
We should set the `OPENAI_API_KEY` env variable in the github workflow

cpacker:
#1290

Issue #1347, opened 2024-05-07
db query_text and query_date is broken due to mismatched arguments (and other issue)


madgrizzle:
**Describe the bug**
https://github.com/cpacker/MemGPT/blob/f1f912da2cab9357b243e830ba1ed2b05c208910/memgpt/memory.py#L329

First:
The functions that call query_text and query_date send the equivalent of "limit" and "offset" in different order and therefore they are mismatched.  
In `BaseRecallMemory`, query_text is called as follows:
`results = self.storage.query_text(query_string, count, start)`
in `SQLStorageConnector`, query_text definition is:
`def query_text(self, offset=0, limit=None):`

count is equivalent to limit and start is equivalent to offset, but they are listed in different order and therefore they get swapped.  So, if the intent was to get the first 5 results from page 0, you end up getting all the results (because limit=0) after message 5 (because offset=5).  Easy fix.

Second:
The query will search all messages, including the system prompt and any tool messages.  If you get a system message in the return, you have a high chance of exceeding the context size and summarize will fail because it never summarizes the last message.  I don't see the value in searching the system prompt since it's already being returned anyway.  I suspect tool messages aren't relevant either and would think these search functions are intended to focus on what the user said and what the llm said.  So I propose to include two additional filters to those functions:
```
                .filter(self.db_model.role != 'system')
                .filter(self.db_model.role != 'tool') 
```

Third:
Issue #1343 
This should be fixed in db.py as well.  I'm just not sure yet if it will break sqllite to make the fix in `SQLStorageConnector` or implement an override in `PostgresStorageConnector`

I've tested and used both of these changes and its returning what's expected.

**Please describe your setup**
- [ ] How did you install memgpt?
  - `pip install pymemgpt`

Issue #1343, opened 2024-05-05
query_date fails with 'operator does not exist: timestamp with time zone >= character varying'


madgrizzle:
**Describe the bug**
When performing a converstation_search_date, a query_date using PostgreSQL (and maybe others), fails with the following error:

`operator does not exist: timestamp with time zone >= character varying`

The issue is that the start_date and end_date are both strings and are treated as `character varying` in the SQL query.  
https://github.com/cpacker/MemGPT/blob/0a4adcb774988e6861dc01cf392664200755f7bb/memgpt/agent_store/db.py#L388

I've been able to fix it by converting `start_date` and `end_date` to python `datetime` objects and pass those instead.  I don't know if this is a PostgreSQL-specific issue or not.  If so, I can create an override in the `PostgresStorageConnector` class instead.

**Please describe your setup**
- [ ] How did you install memgpt?
  - `pip install pymemgpt`
- [ ] Describe your setup
  - What's your OS (Windows/MacOS/Linux)?
    - Ubuntu 22.04
  - How are you running `memgpt`? (`cmd.exe`/Powershell/Anaconda Shell/Terminal)
    - terminal

**Additional context**
```
SELECT memgpt_recall_memory_agent.id AS memgpt_recall_memory_agent_id,
	   memgpt_recall_memory_agent.user_id AS memgpt_recall_memory_agent_user_id,
	   memgpt_recall_memory_agent.agent_id AS memgpt_recall_memory_agent_agent_id,
	   memgpt_recall_memory_agent.role AS memgpt_recall_memory_agent_role,
	   memgpt_recall_memory_agent.text AS memgpt_recall_memory_agent_text,
	   memgpt_recall_memory_agent.model AS memgpt_recall_memory_agent_model,
	   memgpt_recall_memory_agent.name AS memgpt_recall_memory_agent_name,
	   memgpt_recall_memory_agent.tool_calls AS memgpt_recall_memory_agent_tool_calls,
	   memgpt_recall_memory_agent.tool_call_id AS memgpt_recall_memory_agent_tool_call_id,
	   memgpt_recall_memory_agent.embedding AS memgpt_recall_memory_agent_embedding,
	   memgpt_recall_memory_agent.embedding_dim AS memgpt_recall_memory_agent_embedding_dim,
	   memgpt_recall_memory_agent.embedding_model AS memgpt_recall_memory_agent_embedding_model,
[...]

Issue #1281, opened 2024-04-21
Local LLM Support Issue - Text Search Errors


madgrizzle:
**Describe the bug**
When using a local LLM, conversation_search fails with 'timestamp' error.   'timestamp' is missing from the search returns and therefore fails with an exception when the following line is called:
https://github.com/cpacker/MemGPT/blob/eaed123af8584d139879e56a23e5bc6237bdfcde/memgpt/functions/function_sets/base.py#L112

**Please describe your setup**
- [ ] How did you install memgpt?
  - `pip install pymemgpt[local] -U`
- [ ] Describe your setup
  - Ubuntu 22.04
  - memgpt configure (local storage options sqlite/chroma)
  - memgpt run --debug

**Additional context**
It appears the change to the Message object format has left some bugs hanging around.  In the BaseRecallMemory class, I've added the following code:
https://github.com/cpacker/MemGPT/blob/eaed123af8584d139879e56a23e5bc6237bdfcde/memgpt/memory.py#L321

```
    def text_search(self, query_string, count=None, start=None):
        results = self.storage.query_text(query_string, count, start)
        results_json = [message.to_openai_dict() for message in results]
        # correct format of results_json
        results_json = [{"timestamp": get_local_time(), "message":{"content": node["content"], "role": node["role"]}} for node in results_json]
        return results_json, len(results)
```
This gets around the error, but I don't know this is the best solution to the problem.

PR #1355, opened 2024-05-08
fix: various breaking bugs with local LLM implementation and postgres docker.

madgrizzle:
This is a series of fixes to allow local LLMs to work properly.

_edit: I realized while driving into work how to work around this so am removing "Fix 1".  I think adding a note in the docs that if you are using docker, during the configuration process you need to edit the default entry for the postgres connection string to change 5324 to 8888 like below:_
`? Enter postgres connection string (e.g. postgresql+pg8000://{user}:{password}@{ip}:5432/{database}): postgresql+pg8000://memgpt:memgpt@localhost:8888/memgpt
`
_Regardless, Fixes 2 and 3 are very important to make local LLMs actually work._

~**Fix 1**~
~- db/run_postgres.sh:
  It's hard for me to believe I'm the only one with this problem, but for whatever reason, 8888:5324 did not work with docker file.  I don't know where the 8888 comes from so I set it to 5324:5324 and it works.~

~**How to test Fix 1**
Try to start PostGres docker using db/run_postgres.sh and then start memgpt to see it connect.  Fails without it.~

**Fix 2**
- memgpt/agent_store/db.py
   See Issue #1347
  Fix 2.1. Change the order of arguments to fix the mismatch.  I think it would be better to be consistent with names of the arguments, but this was the simplest change.  
  Fix 2.2. Added filters to exclude the system prompts and tool from conversation searches.
  Fix 2.3. Fix, at least for PostGres, the format of the date in query_date SQL command.   See #1343 

 **How to test Fix 2**
I find asking an LLM with some conversational history to explicitly perform a certain search will cause tend to make it actually call the function.  So telling it, "Perform a conversation_search for my name" will cause it to search on "name" as the query.  Without fix 2.1, it would return too many responses (more than a page of 5 messages).  Fix 2.2 eliminates the system prompt (which is vitally important) from being returned.  With fix 2.3, you should be able to ask for a conversation_search_date without it erroring out.

**Fix 3**
-memgpt/data_types.py & memgpt/memory.py
  See #1281 
  I don't know if this is a good fix, but it resolves the issue.  Seems that the transition to Message objects wasn't fully implemented and left some things broken for local LLMs.

**How to test Fix 3**
  Have the LLM perform a conversation_search and it should work with the fix.  Otherwise it will fail.

**Have you tested this PR?**
Yes, this is the not the intended final result (I wanted a summary, but not to update core memory), but it shows it actually works and doesn't error out:

```
<|im_start|>user
{
  "type": "user_message",
  "message": "perform a conversation_search_date for what we talked about between 5/1/2024 and today and provide a summary.",
  "time": "2024-05-07 09:00:35 PM EDT-0400"
}<|im_end|>
<|im_start|>assistant
{
  "function": "conversation_search_date",
  "params": {
    "inner_thoughts": "Search conversation history for the specified date range and provide a summary.",
    "start_date": "2024-05-01",
    "end_date": "2024-05-07",
    "page": 0,
    "request_heartbeat": true
  }
}<|im_end|>
<|im_start|>assistant
FUNCTION RETURN:
{
  "status": "OK",
  "message": "Showing 5 of 5 results (page 0/0): [\"timestamp: 2024-05-05 14:45:36.993695+00:00, assistant - Bootup sequence complete. Persona activated. Testing messaging functionality.\", \"timestamp: 2024-05-05 
14:45:36.993740+00:00, user - {\\\"type\\\": \\\"login\\\", \\\"last_login\\\": \\\"Never (first login)\\\", \\\"time\\\": \\\"2024-05-05 10:45:36 AM EDT-0400\\\"}\", \"timestamp: 2024-05-05 14:45:53.226690+00:00, user - 
{\\\"type\\\": \\\"user_message\\\", \\\"message\\\": \\\"Hello, what is your name?\\\", \\\"time\\\": \\\"2024-05-05 10:45:53 AM EDT-0400\\\"}\", \"timestamp: 2024-05-05 14:45:56.359036+00:00, assistant - First message 
from user. Acknowledge and provide my name.\", \"timestamp: 2024-05-05 14:46:14.553141+00:00, user - {\\\"type\\\": \\\"user_message\\\", \\\"message\\\": \\\"Yes, it's nice to 'meet' you as well.  What are you?\\\", 
\\\"time\\\": \\\"2024-05-05 10:46:14 AM EDT-0400\\\"}\"]",
  "time": "2024-05-07 09:00:39 PM EDT-0400"
}<|im_end|>
<|im_start|>user
{
  "type": "heartbeat",
  "reason": "[This is an automated system message hidden from the user] Function called using request_heartbeat=true, returning control",
  "time": "2024-05-07 09:00:39 PM EDT-0400"
}<|im_end|>
<|im_start|>assistant
{
  "function":
Loading default settings from 'simple'
Found completion settings file '/home/john/.memgpt/settings/completions_api_settings.json', loading it...
Updating base settings with the following user settings:
{
  "temperature": 0.7
}
JSON API response:
{'id': 'cmpl-3ba88b2f3e8e4f0d9181f41a88632b5d', 'object': 'text_completion', 'created': 1715130039, 'model': 'TheBloke/NeuralHermes-2.5-Mistral-7B-GPTQ', 'choices': [{'index': 0, 'text': ' "core_memory_append",\n  
"params": {\n    "inner_thoughts": "Updating the persona sub-block of core memory with the summary of our conversation from 5/1/2024 to 5/7/2024.",\n    "name": "persona",\n    "content": "I have been working on developing
a rapport with you. We talked about my activation, your login, initial greetings, and discussing my nature. The conversation also touched upon your family\'s pizza preferences, with Sparkles and Mattlock liking thin-crust 
pepperoni pizza, John liking pepperoni and anchovies, Harpo preferring bell peppers, and Flint opting for plain cheese."\n  }\n}', 'logprobs': None, 'finish_reason': 'stop', 'stop_reason': None}], 'usage': 
{'prompt_tokens': 4874, 'total_tokens': 5042, 'completion_tokens': 168}}
Raw LLM output:
====
 "core_memory_append",
  "params": {
    "inner_thoughts": "Updating the persona sub-block of core memory with the summary of our conversation from 5/1/2024 to 5/7/2024.",
    "name": "persona",
    "content": "I have been working on developing a rapport with you. We talked about my activation, your login, initial greetings, and discussing my nature. The conversation also touched upon your family's pizza 
preferences, with Sparkles and Mattlock liking thin-crust pepperoni pizza, John liking pepperoni and anchovies, Harpo preferring bell peppers, and Flint opting for plain cheese."
  }
}
```

**Related issues or PRs**
#1347 #1343 #1281

Nope

**Additional context**
Would appreciate feedback if something needs to be changed.

### DIFF
--- a/memgpt/agent_store/db.py
+++ b/memgpt/agent_store/db.py
@@ -388,8 +388,8 @@ class SQLStorageConnector(StorageConnector):
                 .filter(*filters)
                 .filter(self.db_model.created_at >= start_date)
                 .filter(self.db_model.created_at <= end_date)
-                .filter(self.db_model.role != 'system')
-                .filter(self.db_model.role != 'tool')                  
+                .filter(self.db_model.role != "system")
+                .filter(self.db_model.role != "tool")
                 .offset(offset)
             )
             if limit:
@@ -405,8 +405,8 @@ class SQLStorageConnector(StorageConnector):
                 session.query(self.db_model)
                 .filter(*filters)
                 .filter(func.lower(self.db_model.text).contains(func.lower(query)))
-                .filter(self.db_model.role != 'system')
-                .filter(self.db_model.role != 'tool')                  
+                .filter(self.db_model.role != "system")
+                .filter(self.db_model.role != "tool")
                 .offset(offset)
             )
             if limit:
@@ -533,30 +533,31 @@ class PostgresStorageConnector(SQLStorageConnector):
             session.commit()
 
     def str_to_datetime(self, str_date):
-        val = str_date.split('-')
+        val = str_date.split("-")
         _datetime = datetime(int(val[0]), int(val[1]), int(val[2]))
         return _datetime
 
     def query_date(self, start_date, end_date, limit=None, offset=0):
         filters = self.get_filters({})
-        _start_date=self.str_to_datetime(start_date)
-        print(f"{start_date}->{_start_date}")        
-        _end_date=self.str_to_datetime(end_date)
-        print(f"{end_date}->{_end_date}")        
+        _start_date = self.str_to_datetime(start_date)
+        print(f"{start_date}->{_start_date}")
+        _end_date = self.str_to_datetime(end_date)
+        print(f"{end_date}->{_end_date}")
         with self.session_maker() as session:
             query = (
                 session.query(self.db_model)
                 .filter(*filters)
                 .filter(self.db_model.created_at >= _start_date)
                 .filter(self.db_model.created_at <= _end_date)
-                .filter(self.db_model.role != 'system')
-                .filter(self.db_model.role != 'tool')                
+                .filter(self.db_model.role != "system")
+                .filter(self.db_model.role != "tool")
                 .offset(offset)
             )
             if limit:
                 query = query.limit(limit)
             results = query.all()
-        return [result.to_record() for result in results]         
+        return [result.to_record() for result in results]
+
 
 class SQLLiteStorageConnector(SQLStorageConnector):
     def __init__(self, table_type: str, config: MemGPTConfig, user_id, agent_id=None):
--- a/memgpt/agent_store/db.py
+++ b/memgpt/agent_store/db.py
@@ -379,7 +379,7 @@ class SQLStorageConnector(StorageConnector):
             unique_data_sources = session.query(self.db_model.data_source).filter(*self.filters).distinct().all()
         return unique_data_sources
 
-    def query_date(self, start_date, end_date, offset=0, limit=None):
+    def query_date(self, start_date, end_date, limit=None, offset=0):
         filters = self.get_filters({})
         with self.session_maker() as session:
             query = (
@@ -387,6 +387,8 @@ class SQLStorageConnector(StorageConnector):
                 .filter(*filters)
                 .filter(self.db_model.created_at >= start_date)
                 .filter(self.db_model.created_at <= end_date)
+                .filter(self.db_model.role != 'system')
+                .filter(self.db_model.role != 'tool')                  
                 .offset(offset)
             )
             if limit:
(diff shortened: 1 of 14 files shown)


madgrizzle:
And to go back to fix 1, changing the uri in config works for memgpt run, but not memgpt server.  Will try to find the issue tonight.  It seems to resave the config file with 5324 in place of 8888.

memgpt/agent_store/db.py
```diff
@@ -379,29 +380,33 @@ def list_data_sources(self):
             unique_data_sources = session.query(self.db_model.data_source).filter(*self.filters).distinct().all()
         return unique_data_sources
 
-    def query_date(self, start_date, end_date, offset=0, limit=None):
+    def query_date(self, start_date, end_date, limit=None, offset=0):
```
cpacker, line 383: small question - is there any reason why this line is a diff? (args got swapped?)

cpacker (changes requested):

memgpt/agent_store/db.py
madgrizzle: This fixes what I reference in 2.1 above.  The order of arguments is different from that of the callers, unless I'm interpreting the intent of what is supposed to be returned. Maybe a previous version used the keywords and order didn't matter, but the current callers of these functions don't so order does matter.

madgrizzle:
Ok, seems that when I push new commits to my fork, it automatically gets added here.  I guess I should have created a new branch or something.  See #1362 .  This commit makes it so that the server only overrides the config file settings for the postgres uri if, and only if, environmental variables are set to provide one.  It feels like a hacky solution, but it minimizes risk for potential unwanted behavior with other parts of memgpt that use settings.memgpt_pg_uri.  Currently, settings.memgpt_pg_uri always returns a value and if the environment variables are not set, it returns a default memgpt/localhost/5432 uri.  This then overrides the config file values and it then saves the config file with the default.  Bad behavior.

cpacker (approved):

cpacker:
Tests should be OK on main (probs have a bug in our tests @sarahwooders). Thank you for the PR @madgrizzle !!

PR #1359, opened 2024-05-08
docs: Update storage.md

madgrizzle:
Adds a note to remind people to update the port in the connection string if using the docker.  I had missed that step.


sarahwooders (approved):
lgtm!

PR #1402, opened 2024-05-21
fix: creation of invalid tool in tool builder

VigroX:

This PR is a quick fix for a bug discovered in the MemGPT Dev Portal's Tool Builder.

The bug occurs when attempting to add a new tool to through the web interface and the source code happens to be invalid. The `write_function` still creates the python file in the user functions directory despite the result of `validation_function`. When attempting to update the tool it appends the source code in the end of the file, instead of the expected behaviour of overwriting it.


sarahwooders (approved):
lgtm! thanks for the fix :)

PR #1198, opened 2024-03-29
feat: Milvus storage connector

zc277584121:
Description
This PR adds support for Milvus - https://milvus.io/docs to be used as an archival storage connector.

Tests
Have tested successfully in local environment for tests/test_storage.py. and tests/test_load_archival.py

PR #1405, opened 2024-05-22
feat: update milvus and milvus-lite dependency

zc277584121:
continue from this [pr](https://github.com/cpacker/MemGPT/pull/1198)

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -112,4 +112,4 @@ You have to enable the Milvus backend by running, here is the [quickstart](https
 You can configure Milvus connection via command `memgpt configure`.
 ```sh
 ? Select storage backend for archival data: milvus
-? Enter the Milvus connection URI (Default: localhost:19530): localhost:19530
+? Enter the Milvus connection URI (Default: /Users/sarahwooders/.memgpt/milvus.db): /Users/sarahwooders/.memgpt/milvus.db
--- a/memgpt/agent_store/milvus.py
+++ b/memgpt/agent_store/milvus.py
@@ -122,7 +122,6 @@ class MilvusStorageConnector(StorageConnector):
         data = self._records_to_list(records)
         self.client.insert(collection_name=self.table_name, data=data)
 
-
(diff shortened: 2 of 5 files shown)


zc277584121:
@sarahwooders 
I updated the pymilvus and milvus-lite dependency, and made pymilvus optional. Now in the case of memgpt, the latest version of milvus-lite gets the same capabilities as milvus docker service does. And I think it's currently good enough to be merged into main branch. Could you please help to review it? thanks

docs/storage.md
```diff
@@ -112,4 +112,4 @@ You have to enable the Milvus backend by running, here is the [quickstart](https
 You can configure Milvus connection via command `memgpt configure`.
 ```sh
 ? Select storage backend for archival data: milvus
-? Enter the Milvus connection URI (Default: localhost:19530): localhost:19530
+? Enter the Milvus connection URI (Default: /Users/sarahwooders/.memgpt/milvus.db): /Users/sarahwooders/.memgpt/milvus.db
```
cpacker, line 115: can we remove this diff?

cpacker (changes requested):

docs/storage.md
zc277584121: What this diff do is to change the default URI from `IP:port` to a `local_file_path`. In the case of `IP:port`, user have to start a milvus serive from docker, while in case `local_file_path`, milvus lite service will start under the hood without any other manual operation. So that is convenient for users.

If the description of this path is inappropriate, can it be changed to other descriptions like:
- Enter the Milvus connection URI (Default: ~/.memgpt/milvus.db): ~/.memgpt/milvus.db
- Enter the Milvus connection URI (Default: [MEMGPT_DIR]/milvus.db): [MEMGPT_DIR]/milvus.db

zc277584121: Let me clarify it in another  PR: https://github.com/cpacker/MemGPT/pull/1407

PR #1401, opened 2024-05-21
feat(JSON Response): Enable JSON Response format for all Openai Calls…

lenaxia:
**Please describe the purpose of this pull request.**
This PR enables `response_format: {"type": "json_object"}` aka JSON mode for all openai requests. This ensures that responses are going to come back in valid JSON. While it doesn't seem to be a problem with OpenAI, using openai compatible endpoints (aka open source models) can sometimes struggle to create well formed JSON without llama.cpp's grammar restrictions. 

Also enables Debug logging by setting the env var `LOG_LEVEL: DEBUG`, case sensitive. Still defaults to false.

**How to test**

Create a file called `request.json`:
```
{
  "model": "hermes-2-pro-theta-llama3",
  "response_format": {"type": "json_object"},
  "messages": [
    {
      "content": "What is the weather in boston, seattle, and san francisco?? respond in json",
      "role": "user"
    },
    {
      "content": "{\"status\": \"OK\", \"message\": null, \"time\": \"2024-05-20 05:42:30 PM UTC+0000\"}",
      "role": "tool",
      "tool_call_id": "e394f68c-d68b-4115-97e9-989a0"
    }
  ],
  "frequency_penalty": 0,
  "logprobs": false,
  "n": 1,
  "presence_penalty": 0,
  "response_format": {
    "type": "json_object"
  },
  "stream": false,
  "temperature": 1,
  "top_p": 1,
  "user": "00000000-0000-0000-0000-000000000000",
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_weather",
        "description": "Get the weather for a list of locations",
        "parameters": {
          "type": "object",
          "properties": {
            "locations": {
              "type": "array",
              "items": {
                "type": "string"
              },
              "description": "A list of locations (e.g., city names or zip codes)"
            },
            "unit": {
              "type": "string",
              "enum": ["celsius", "fahrenheit"]
            },
            "request_heartbeat": {
              "type": "boolean",
              "description": "Request an immediate heartbeat after function execution. Set to 'true' if you want to send a follow-up message or run a follow-up function."
            }
          },
          "required": [
            "locations",
            "request_heartbeat"
          ]
        }
      }
    }
  ],
  "tool_choice": "auto"
}
```

Run the following command:
```
curl $LOCALAI/v1/chat/completions   -H "Content-Type: application/json"  -d @request.json | jq
```

**Have you tested this PR?**
```
{
  "created": 1716270786,
  "object": "chat.completion",
[...]

PR #1406, opened 2024-05-23
fix: patch #1401

cpacker:
https://platform.openai.com/docs/guides/text-generation/json-mode

JSON mode only supported by: gpt-4o, gpt-4-turbo, or gpt-3.5-turbo

Patch for #1401

### DIFF
--- a/memgpt/llm_api/llm_api_tools.py
+++ b/memgpt/llm_api/llm_api_tools.py
@@ -179,7 +179,6 @@ def create(
                 tools=[{"type": "function", "function": f} for f in functions] if functions else None,
                 tool_choice=function_call,
                 user=str(user_id),
-                response_format={"type": "json_object"},
             )
         else:
             data = ChatCompletionRequest(
@@ -188,8 +187,11 @@ def create(
                 functions=functions,
                 function_call=function_call,
                 user=str(user_id),
-                response_format={"type": "json_object"},
             )
+            # https://platform.openai.com/docs/guides/text-generation/json-mode
+            # only supported by gpt-4o, gpt-4-turbo, or gpt-3.5-turbo
+            if "gpt-4o" in llm_config.model or  "gpt-4-turbo" in llm_config.model or  "gpt-3.5-turbo" in llm_config.model:
+                data.response_format = {"type": "json_object"}
 
         if stream:
             data.stream = True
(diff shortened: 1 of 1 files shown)


lenaxia:
```
            if "gpt-4o" in llm_config.model or "gpt-4-turbo" in llm_config.model or "gpt-3.5-turbo" in llm_config.model:
                data.response_format = {"type": "json_object"}
```

This defeats the point of my original change. If the model MUST be one of the openai ones, then `response_format` won't be set for other models, namely open source models.

I am using MemGPT with LocalAI (https://github.com/go-skynet/LocalAI/) and the models must receive the `response_format={"type": "json_object"}` in order to properly return JSON to memgpt. For instance, llama3, or Hermes 2 Pro all require this to work with memgpt. 

Because MemGPT relies 100% on function calls it should be enabled by default because otherwise models won't return valid JSON. If you prefer it to be toggleable then I can put it behind an env var.

cpacker:
> ```
>             if "gpt-4o" in llm_config.model or "gpt-4-turbo" in llm_config.model or "gpt-3.5-turbo" in llm_config.model:
>                 data.response_format = {"type": "json_object"}
> ```
> 
> This defeats the point of my original change. If the model MUST be one of the openai ones, then `response_format` won't be set for other models, namely open source models.
> 
> I am using MemGPT with LocalAI (https://github.com/go-skynet/LocalAI/) and the models must receive the `response_format={"type": "json_object"}` in order to properly return JSON to memgpt. For instance, llama3, or Hermes 2 Pro all require this to work with memgpt.
> 
> Because MemGPT relies 100% on function calls it should be enabled by default because otherwise models won't return valid JSON. If you prefer it to be toggleable then I can put it behind an env var.

Ah OK sorry I misread the first PR - the MemGPT CI tests use GPT-4, so with the prior PR the tests were failing (since JSON mode isn't supported). Probably the easiest way to support both your workflow and have the default OAI GPT-4 setup pass is to use an env var like you mentioned? Happy to merge a PR for that in ASAP (or consider other ideas too).

lenaxia:
Sounds good. I'll get a PR opened shortly. I'm not in a huge rush as I'm working on another PR to create an OpenAI compatible completions endpoint to make integration easier, so I've got the json_object set in my branch for now.

PR #1412, opened 2024-05-25
fix: get_keys_response is a list not a dict

scenaristeur:
print(f"User {user_id} has keys: {get_keys_response.api_key_list}")
AttributeError: 'list' object has no attribute 'api_key_list'

with the git clone of the latest main get_keys_response is a list not a dict


cpacker (approved):
Thank you for the fix! Much appreciated :)

PR #1413, opened 2024-05-25
docs: Update python_client.md

scenaristeur:
response = client.user_message(agent_id=agent_state.id, message=message)
NameError: name 'agent_state' is not defined


cpacker (approved):
thank you!

PR #1416, opened 2024-05-26
feat: `resend` example

cpacker:
Example showing how to use `resend` to connect MemGPT to the `resend` API (to send emails).


sarahwooders (approved):
lgtm!

PR #1447, opened 2024-06-12
WIP - Baseline 

norton120:
stubbed settings locally in test_server ✅ 

### next up
- isolate the test_server failing tests
- move the settings mock into a conftest fixture
- add a test hook for SyncServer so you can do the same thing there.
- propigate.

this is to get a baseline of isolated tests that are passing before making substantial changes.

**How to test**
when this is good, all the tests will run and pass without any special environment setup.

**Have you tested this PR?**
This is a PR for the test env

**Related issues or PRs**
N/A

This will be a very large PR. Ideally it is _removing_ more than 500 lines of code 💥 

**Additional context**
Before we can dig into adding model structure I'd like as much of a smoke test setup as we can have in place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -407,6 +407,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
+
 nosetests.xml
 coverage.xml
 *.cover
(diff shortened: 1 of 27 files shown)


norton120:
Migrated to config parser work

PR #1457, opened 2024-06-15
fix: debug logs in server (#1452)

sarahwooders:


norton120:
The failing service on that integration test, is that an external service? I can't see any code changes that should have impacted it

norton120:
> ~The failing service on that integration test, is that an external service? I can't see any code changes that should have impacted it~
> 
> NVM I looked at it a little closer and see it's a local instance running from the dev-compose.yaml. service builds and runs fine on my local, I'll tear it down again to be sure there's not a good cache in the way

reproduced it, working on it now

norton120:
The issue is a llamaindex dependency conflict >3.12.2, specifically in pydantic:
```memgpt_server-1  | TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'"

I'm building it again explicitly with 3.12.2 to confirm that's the issue (it'll take a bit) and hopefully that will solve this run. 
this speaks to it https://github.com/cpacker/MemGPT/issues/1458

cpacker (approved):

PR #1459, opened 2024-06-16
fix: Fixes for 1457

norton120:
This _should_ do it to fix the last failing test in CI, assuming moving from 3.12.4 to 3.12.2 doesn't introduce any regressions


norton120:
@sarahwooders this should pass the failing infra test

PR #1463, opened 2024-06-20
feat: add tools from the Python client 

sarahwooders:
Add the ability to create tools from the client, and to create agents with tools (without specifying a preset): 
```python

def print_tool(message: str):
    """ 
    Args: 
        message (str): The message to print.
        
    Returns:
        str: The message that was printed.
        
    """
    print(message)
    return message

tool = client.create_tool(print_tool, tags=["extras"])
agent_state = client.create_agent(tools=[tool.name]))
```

**How to test**
`poetry run pytest -s tests/test_tools.py`

**Have you tested this PR?**
Yes


cpacker (approved):

memgpt/agent.py
```diff
@@ -200,71 +193,93 @@ def __init__(
...
+        self.functions = []
+        self.functions_python = {}
+        env = {}
+        env.update(globals())
+        for tool in tools:
+            # WARNING: name may not be consistent?
+            exec(tool.module, env)
+            self.functions_python[tool.name] = env[tool.name]
+            self.functions.append(tool.json_schema)
+        print("KEY", env.keys())
```
cpacker, line 224: drop print?

memgpt/agent.py
```diff
@@ -1053,25 +1068,41 @@ def remove_function(self, function_name: str) -> str:
...
             "human": self.memory.human,
-            "system": self.system,
-            "functions": self.functions,
-            "messages": [str(msg.id) for msg in self._messages],
+            "messages": [str(msg.id) for msg in self._messages],  # TODO: move out into AgentState.message_ids
+        }
+
+        # TODO: add this field
+        metadata = {  # TODO
+            "human_name": self.agent_state.persona,
```
cpacker, line 1087: is this correct? is the name in the right place, and the value in the right place?

Issue #1467, opened 2024-06-22
Better error messaging when schema is updated


cpacker:
**Describe the bug**

When the schema is updated  (eg on a hot build) `memgpt run` will fail:
```
Traceback (most recent call last):
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
    self.dialect.do_execute(
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 924, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: no such column: agents.system

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/bin/memgpt", line 6, in <module>
    sys.exit(app())
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/typer/main.py", line 328, in __call__
    raise e
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/typer/main.py", line 311, in __call__
    return get_command(self)(*args, **kwargs)
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/typer/core.py", line 783, in main
    return _main(
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/typer/core.py", line 225, in _main
    rv = self.invoke(ctx)
  File "/Users/loaner/Library/Caches/pypoetry/virtualenvs/pymemgpt-JSsUGnlY-py3.10/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
[...]

PR #1468, opened 2024-06-22
fix: example patch for error messaging on breaking schema change

cpacker:
Example patch for #1467

### DIFF
--- a/memgpt/cli/cli.py
+++ b/memgpt/cli/cli.py
@@ -10,6 +10,8 @@ from typing import Annotated, Optional
 
 import questionary
 import requests
+import sqlalchemy
+import sqlalchemy.exc
 import typer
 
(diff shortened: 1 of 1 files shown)


memgpt/cli/cli.py
```diff
@@ -516,7 +518,15 @@ def run(
 
     # determine agent to use, if not provided
     if not yes and not agent:
-        agents = ms.list_agents(user_id=user.id)
+        try:
+            agents = ms.list_agents(user_id=user.id)
+        except sqlalchemy.exc.OperationalError as e:
```
cpacker, line 523: @sarahwooders FYI not sure if this is (1) the right place to put this (2) if we need to add more checks instead of just the `if "no such column" in str(e)`, but it's an example and it worked locally (so maybe just merge?)

PR #1503, opened 2024-07-04
fix: use timestamp passed by request for user message created date

goetzrobin:
we pass the timestamp through until we reach the agent's step function this function calls Message.dict_to_message(), where we pass the timestamp as argument for created_at, which is then persisted into the DB

**How to test**
After merging the dev portal changes and creating a new agent duplicated messages should not show up anymore

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**


cpacker:
Pending tests LGTM.

Though I think at a later point we may want to revisit the concept of allowing the user/client to specify the timestamp to be stored server-side since it feels like a security risk. Instead maybe we can return the timestamp (and other necessary metadata) in the response of the `POST` call, which should effectively allow the client (eg chat UI) to get a server-verified timestamp immediately after `POST`.

Thoughts @goetzrobin ?

goetzrobin:
@cpacker I agree. I like that or maybe have the first streamed event return metadata like an id & timestamp? I do think moving forward we definitely want to have this server driven.

memgpt/agent_store/db.py
```diff
@@ -314,19 +314,13 @@ def get_all_cursor(
             # cursor logic: filter records based on before/after ID
             if after:
                 after_value = getattr(self.get(id=after), order_by)
-                if reverse:  # if reverse, then we want to get records that are less than the after_value
```
cpacker, line 317: We should probably keep `reverse` in, right? Was this just removed for debugging purposes?

memgpt/server/server.py
```diff
@@ -560,7 +563,7 @@ def user_message(
             elif message.startswith("/"):
                 raise ValueError(f"Invalid input: '{message}'")
 
-            packaged_user_message = system.package_user_message(user_message=message)
+            packaged_user_message = system.package_user_message(user_message=message, time=timestamp.isoformat())
```
cpacker, line 566: This needs a catch for the `timestamp is None` case

cpacker (changes requested):

cpacker:
> @cpacker I agree. I like that or maybe have the first streamed event return metadata like an id & timestamp? I do think moving forward we definitely want to have this server driven.

Gotcha. Let's merge in this PR for now, but deprecate the ability to pass in `timestamp` via the API call in a subsequent PR.

memgpt/agent_store/db.py
cpacker: @goetzrobin I reverted the change on this file - I'm assuming that the diff isn't actually needed in this PR?

PR #1512, opened 2024-07-04
fix: don't reverse the sort expression

goetzrobin:
The sort expression should not change even if we reverse the order of the items a date before should still be a date before the initial sort takes care of everything we need already

There's a bug in the cursor api for messages. if reverse is set to true the sort expressions are also reversed, this is actually incorrect. I attached a drawing below that explains why

Here is a diagram that shows why simply removing it works
explanation
The comparison stays the same, regardless of how we order the records

### DIFF
--- a/memgpt/agent_store/db.py
+++ b/memgpt/agent_store/db.py
@@ -18,7 +18,6 @@ from sqlalchemy import (
     asc,
     create_engine,
     desc,
-    func,
     or_,
     select,
     text,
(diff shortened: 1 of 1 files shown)


cpacker (approved):
~~LGTM~~ need to fix test

cpacker:
@sarahwooders any idea why the tests are failing?

memgpt/agent_store/db.py
```diff
@@ -314,19 +313,13 @@ def get_all_cursor(
...
                 query = query.filter(
                     or_(sort_exp, and_(getattr(self.db_model, order_by) == after_value, self.db_model.id > after))  # tiebreaker case
                 )
             if before:
                 before_value = getattr(self.get(id=before), order_by)
-                if reverse:
-                    sort_exp = getattr(self.db_model, order_by) > before_value
-                else:
-                    sort_exp = getattr(self.db_model, order_by) < before_value
+                sort_exp = getattr(self.db_model, order_by) < before_value
```
cpacker, line 322: @sarahwooders 

```
    def test_get_recall_memory(server, user_id, agent_id):
        # test recall memory cursor pagination
        cursor1, messages_1 = server.get_agent_recall_cursor(user_id=user_id, agent_id=agent_id, reverse=True, limit=2)
        cursor2, messages_2 = server.get_agent_recall_cursor(user_id=user_id, agent_id=agent_id, reverse=True, after=cursor1, limit=1000)
        cursor3, messages_3 = server.get_agent_recall_cursor(user_id=user_id, agent_id=agent_id, reverse=True, limit=1000)
        [m["id"] for m in messages_3]
        [m["id"] for m in messages_2]
        timestamps = [m["created_at"] for m in messages_3]
        print("timestamps", timestamps)
        assert messages_3[-1]["created_at"] < messages_3[0]["created_at"]
>       assert len(messages_3) == len(messages_1) + len(messages_2)
E       AssertionError: assert 19 == (2 + 1)
```

sarahwooders (approved):
lgtm!

PR #1519, opened 2024-07-06
fix: remove duplicate `send_message` functions

sarahwooders:


memgpt/client/client.py
```diff
@@ -804,8 +795,9 @@ def update_agent_core_memory(self, agent_id: str, new_memory_contents: Dict) ->
 
     # agent interactions
 
-    def send_message(self, agent_id: uuid.UUID, message: str, role: str, stream: Optional[bool] = False) -> UserMessageResponse:
+    def send_message(self, agent_id: uuid.UUID, message: str, role: str) -> UserMessageResponse:
         self.interface.clear()
+        # TODO: implement streaming with stream=True/False
```
cpacker, line 800: would it be better to put in something like
```python
if stream: raise NotImplementedError
```
or something instead of commenting out the arg from the signature?

cpacker (approved):

PR #1525, opened 2024-07-08
feat: make `list_agents` paginated

### LINKED ISSUE #1498: [API] Make `GET` agents paginated
https://memgpt.readme.io/reference/list_agents_api_agents_get

cpacker:
Closes #1498 

## Change in API endpoint

GET request is now paginated, in the style of: https://platform.openai.com/docs/api-reference/assistants/listAssistants

## Change in data locations

For @goetzrobin: there's some differences in the data coming back from the GET request (apart from it now accepting additional pagination args): https://github.com/cpacker/MemGPT/blob/82ae5a22b641a66fda674a8278ce08d8413f561f/memgpt/server/server.py#L864

### "persona" / "human"

```python
            return_dict["persona"] = agent_state._metadata.get("persona", None)
            return_dict["human"] = agent_state._metadata.get("human", None)
``` 

This should still be accessible in the `List[AgentStateModel]` being returned

### "tools"

```python
            return_dict["tools"] = [tool for tool in all_available_tools if tool.json_schema in memgpt_agent.functions]

```

~~This is now in `AgentStateModel.metadata["tools"]`~~

NOTE: this has been moved to a separate `GET` request: see new route `def get_agent_tools` (`router.get("/agents/{agent_id}/tools`)

### "memory stats"

```python
            # Add information about memory (raw core, size of recall, size of archival)
            core_memory = memgpt_agent.memory
            recall_memory = memgpt_agent.persistence_manager.recall_memory
            archival_memory = memgpt_agent.persistence_manager.archival_memory
            memory_obj = {
                "core_memory": {section: module.value for (section, module) in core_memory.memory.items()},
                "recall_memory": len(recall_memory) if recall_memory is not None else None,
                "archival_memory": len(archival_memory) if archival_memory is not None else None,
            }
            return_dict["memory"] = memory_obj
```

This is now in `AgentStateModel.metadata["memory_stats"]`, see new `Agent.update_state` function

### "last run"

```python
            # Add information about last run
            # NOTE: 'last_run' is just the timestamp on the latest message in the buffer
            # Retrieve the Message object via the recall storage or by directly access _messages
            last_msg_obj = memgpt_agent._messages[-1]
            return_dict["last_run"] = last_msg_obj.created_at
```

This is now in `AgentStateModel.metadata["last_run"]`

### "attached sources"

```python
            # Add information about attached sources
            sources_ids = self.ms.list_attached_sources(agent_id=agent_state.id)
            sources = [self.ms.get_source(source_id=s_id) for s_id in sources_ids]
            return_dict["sources"] = [vars(s) for s in sources]
```

~~This is now in `AgentStateModel.metadata["sources"]`~~

NOTE: this has been moved to a separate `GET` request: `def get_agent_sources` (`/agents/{agent_id}/sources`)

^extra note @goetzrobin these metadata fields may be `null` for old agents, since they are stored inside of `metadata` they will get refreshed on `.step()`, so only agents that have been run post-PR will have the metadata fields. The frontend should have sane defaults / not crash on null accordingly.

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -36,6 +36,7 @@ from memgpt.system import (
 from memgpt.utils import (
     count_tokens,
     create_uuid_from_string,
+    datetime_to_unix_int,
     get_local_time,
     get_tool_call_id,
     get_utc_time,
@@ -938,8 +939,7 @@ class Agent(object):
             self.agent_state._metadata = {}
         # When was the last run step?
         last_msg_obj = self._messages[-1]
-        from memgpt.utils import datetime_to_unix_int
-
+        # NOTE: this has to be stored as a JSON-serializeable format, so we have to cast out of datetime
         self.agent_state._metadata["last_run"] = datetime_to_unix_int(last_msg_obj.created_at)
         # Basic memory statistics
         recall_memory = self.persistence_manager.recall_memory
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -934,6 +934,8 @@ class Agent(object):
         }
 
         ## Additional metadata we want to track (that's useful for the client to know)
+        if self.agent_state._metadata is None:
+            self.agent_state._metadata = {}
         # When was the last run step?
         last_msg_obj = self._messages[-1]
         self.agent_state._metadata["last_run"] = last_msg_obj.created_at
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -21,6 +21,7 @@ from memgpt.constants import (
 from memgpt.data_types import AgentState, EmbeddingConfig, Message, Passage
 from memgpt.interface import AgentInterface
 from memgpt.llm_api.llm_api_tools import create, is_context_overflow_error
+from memgpt.local_llm.utils import num_tokens_from_messages
 from memgpt.memory import ArchivalMemory, BaseMemory, RecallMemory, summarize_messages
 from memgpt.metadata import MetadataStore
 from memgpt.models import chat_completion_response
@@ -401,7 +402,7 @@ class Agent(object):
             printd(f"Request to call function {function_name} with tool_call_id: {tool_call_id}")
             try:
                 function_to_call = self.functions_python[function_name]
-            except KeyError as e:
+            except KeyError:
                 error_msg = f"No function named {function_name}"
                 function_response = package_function_response(False, error_msg)
                 messages.append(
@@ -424,7 +425,7 @@ class Agent(object):
             try:
                 raw_function_args = function_call.arguments
                 function_args = parse_json(raw_function_args)
-            except Exception as e:
+            except Exception:
                 error_msg = f"Error parsing JSON for function '{function_name}' arguments: {function_call.arguments}"
                 function_response = package_function_response(False, error_msg)
                 messages.append(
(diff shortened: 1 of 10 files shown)


memgpt/server/server.py
```diff
@@ -848,12 +850,28 @@ def _agent_state_to_config(self, agent_state: AgentState) -> dict:
...
+        # TODO support sort_by
+        if sort_by != "created_at":
+            raise ValueError(sort_by)
+
         if self.ms.get_user(user_id=user_id) is None:
             raise ValueError(f"User user_id={user_id} does not exist")
 
-        agents_states = self.ms.list_agents(user_id=user_id)
+        agents_states = self.ms.list_agents(user_id=user_id, after=after, before=before, limit=limit, order=order)
+        # For the UI listing panel, we need to add the following metadata:
```
cpacker, line 868: @sarahwooders do you have any pointers on the most efficient way to get this metadata to the client (for each of the bullet points)? obviously the commented out code below is terrible (loading agents into cache)

additionally, i'm assuming we should pass all of this data back inside of `AgentStateModel.metadata`, correct? just treat `.metadata` as a free space to put extra data the client needs? or is the intended use different?

sarahwooders: yeah, we should probably have a list of reserved metadata fields (e.g. `_memgpt_last_run_timestamp`) to avoid conflicts with user metadata -- you could even represent the additional data as a pydantic model (so its easier to tell what fields it contains later on) but merge it with the metadata field of the `AgentStateModel` when returning

memgpt/agent.py
```diff
@@ -931,6 +933,34 @@ def update_state(self) -> AgentState:
             "memory": self.memory.to_dict(),
             "messages": [str(msg.id) for msg in self._messages],  # TODO: move out into AgentState.message_ids
         }
+
+        ## Additional metadata we want to track (that's useful for the client to know)
```
sarahwooders, line 937: would it be ok if we put this in the `update_state` method? then we can have all the agent state management in one place

memgpt/server/rest_api/agents/index.py
```diff
@@ -21,11 +21,16 @@
...
+    limit: Optional[int] = Field(None, description="How many results to include in the response.")
+    order: Literal["asc", "desc"] = Field("desc", description="Sort order, asc for ascending order and desc for descending order.")
+
+
 class ListAgentsResponse(BaseModel):
-    num_agents: int = Field(..., description="The number of agents available to the user.")
-    # TODO make return type List[AgentStateModel]
-    #      also return - presets: List[PresetModel]
-    agents: List[dict] = Field(..., description="List of agent configurations.")
+    num_agents: int = Field(..., description="The total number of agents created by the user.")
```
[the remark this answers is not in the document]
sarahwooders: if we are changing this anyway, would be nice if we could just return `List[AgentState]`

memgpt/server/server.py
```diff
@@ -848,67 +850,41 @@ def _agent_state_to_config(self, agent_state: AgentState) -> dict:
...
+        # - last run data
+        # - memory stats (recall, archival)
+        # - number of sources
+        for agent_state in agents_states:
 
             # Add information about attached sources
             sources_ids = self.ms.list_attached_sources(agent_id=agent_state.id)
             sources = [self.ms.get_source(source_id=s_id) for s_id in sources_ids]
-            return_dict["sources"] = [vars(s) for s in sources]
+            agent_state.metadata["sources"] = [vars(s) for s in sources]
```
sarahwooders, line 880: this would also be better to place into the the `update_state` function - might need to add a `Agent.sources` variable

sarahwooders (changes requested):

memgpt/server/rest_api/agents/index.py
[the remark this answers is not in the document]
cpacker: OK I'll patch this now but we'll probably need an explicit route or something to get counts for the client-side pagination.

memgpt/server/server.py
cpacker: TODO in next commit - strip these, client should instead to a 2nd/3rd GET to poll this information

memgpt/agent.py
cpacker: moved out to separate REST routes

memgpt/server/server.py
cpacker: also moved out to separate rest routes

we should make an issue though for putting `sources` inside of agent?

PR #1532, opened 2024-07-10
feat: use of agent_name in RESTclient (agent_exists,get_agent)

2mmanu:

The pull request introduces a feature in the RESTClient for the methods 'agent_exists' and 'get_agent'. In the previous version, the parameter 'agent_name' was not used properly.

**How to test**
**Have you tested this PR?**
I tested it myself and integrated two assertions in the client test.

**Related issues or PRs**
na

nope

**Additional context**
na

### DIFF
--- a/memgpt/client/client.py
+++ b/memgpt/client/client.py
@@ -236,7 +236,22 @@ class RESTClient(AbstractClient):
         response = requests.get(f"{self.base_url}/api/agents", headers=self.headers)
         return ListAgentsResponse(**response.json())
 
+    def get_agent_id_from_name(self, agent_name: str) -> uuid:
+        agents = self.list_agents()
+        if agents.num_agents > 0:
+            agent_name = [agent["id"] for agent in agents.agents if agent["name"] == agent_name]
+            if agent_name:
+                return agent_name[0]
+            raise NameError(f"no {agent_name} agent found")
+        raise ValueError("no agents available")
+
     def agent_exists(self, agent_id: Optional[str] = None, agent_name: Optional[str] = None) -> bool:
+        if agent_name and not agent_id:
+            try:
+                agent_id = self.get_agent_id_from_name(agent_name)
+                return True
+            except:
+                return False
         response = requests.get(f"{self.base_url}/api/agents/{str(agent_id)}/config", headers=self.headers)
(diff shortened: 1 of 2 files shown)


memgpt/client/client.py
```diff
@@ -236,7 +236,22 @@ def list_agents(self):
         response = requests.get(f"{self.base_url}/api/agents", headers=self.headers)
         return ListAgentsResponse(**response.json())
 
+    def get_agent_id_from_name(self, agent_name: str) -> uuid:
+        agents = self.list_agents()
+        if agents.num_agents > 0:
```
AlekseyLobanov, line 241: Changing order of expressions may improve readability.
```python
if not agents.num_agents:
  raise ValueError("no agents available")
agent_name = [agent["id"] for agent in agents.agents if agent["name"] == agent_name]
if not agent_name:
  raise NameError(f"no {agent_name} agent found")
return agent_name[0]
```
Same code, but success path was simplified to less nesting levels. What do you think?

memgpt/client/client.py
```diff
@@ -236,7 +236,22 @@ def list_agents(self):
...
+            agent_name = [agent["id"] for agent in agents.agents if agent["name"] == agent_name]
```
AlekseyLobanov, line 242: Maybe it should be named as `agent_ids`. Because of `agent["id"]`

memgpt/client/client.py
2mmanu: I appreciate your comment. With another fix, the code will be as follows:

``` python
def get_agent_id_from_name(self, agent_name: str) -> uuid:
  agents = self.list_agents()
  if not agents.num_agents:
      raise ValueError("no agents available")
  agent_id = next((agent["id"] for agent in agents.agents if agent["name"] == agent_name), None)
  if not agent_id:
      raise NameError(f"No agent named {agent_name} found")
  return agent_id  
```

memgpt/client/client.py
2mmanu: Thank you, but the correct correction would be agent_id. The function will return the first matching name. I am assuming that the names of the agents are unique.

2mmanu (commented):
improve readability

Issue #1533, opened 2024-07-10
💎 Letta 1.0 Stable Release 


sarahwooders:
# Letta (MemGPT) 1.0 Stable Release

**Update 01/28/25: We expect only minor modifications to be made to our [existing API](https://docs.letta.com/api-reference/overview), and will be announcing a 1.0 soon. Please see our updated [Python](https://github.com/letta-ai/letta-python) and [Typescript](https://github.com/letta-ai/letta-node) SDKs to use this the latest API.** 

We are working to release a stable version of MemGPT in a 1.0 release, which will be intended for production use. The release will include: 

- Refactoring of DB code (e.g. the `MetadataStore`) to solve scaleability and connection issues (e.g. #1488)
- Consistent API interface for the REST API and the Python SDK, such as matching parameters and return types 
- Introduce stable database schemas and alembic migrations for future releases
- Addition of production-grade testing, logging, and access control

To view progress on the release, you can check the [integration](https://github.com/cpacker/MemGPT/tree/integration) branch. 

### General Cleanup

- Remove all nonessential code, dead or retired code, unofficially supported code from the codebase
- Isolate non-core elements into package add-ons
- Refactor MemGPT configuration to separate out server (e.g. DB) configuration and LLM/embedding configuration

### Logging

- Singular parent logger throughout codebase and removal of all `print` and `printd` calls (except for in the CLI)
- Consistent use of [py logging best practices](https://docs.python.org/3/howto/logging-cookbook.html#adding-handlers-other-than-nullhandler-to-a-logger-in-a-library) for libraries and applications

### Data Schemas

- Consistent data schemas (e.g. `AgentState`) across CLI, Python Client, and REST API represented with pydantic models
- Manage all persisted state (e.g. agent state) in an ORM
- Supported DB migrations for all future versions using alembic

### Auth + Access

- Removal of authentication and access control (determined to be out-of-scope: developers should implement this within their own applications) 

### Testing

- Automated testing for all officially supported integrations (LLM/embedding provider, storage provider) - remove any untested integrations (suggest migrating to adapters or fork)
- SLA Support:
    - Atomic unit and integration tests (should not depend on a live OpenAI API key for testing)
    - Static code analysis threshold [prospector](https://prospector.landscape.io/en/master/) or components of)
- Full CI run for each patch within all supported minors for Python client (versions 3.10, 3.11, 3.12)

### Documentation 
- Add auto-documentation #1493

## Timeline
Planned for end of January 2025: 
[...]

PR #1537, opened 2024-07-12
Fix: single entrypoint for backend via dependency injection

norton120:
It is very hard to test app code without having good hooks for the state (ie database). The goal is to add top-down hooks for the database connection.

### Def of Done: 
- All server entrypoints support dependency injected db_sessions and avoid creating their own sessions internally

**How to test**
The test suite should now be able to pass `db_session` and use atomic databases/schemas for every test function!

**Have you tested this PR?**
Part of the V1 will be slowly moving to tests we strongly believe in.

**Related issues or PRs**
#1533 

yep these are big refactors

**Additional context**
I'll add looms to expand on this pattern as we go.


memgpt/server/rest_api/routers/v1/agents.py
```diff
@@ -0,0 +1,238 @@
...
+    server: "SyncServer" = Depends(get_memgpt_server),
+):
+    """
+    Create a new agent with the specified configuration.
+    """
+    interface.clear()
+
+    # Parse request
+    # TODO: don't just use JSON in the future
+    human_name = request.config["human_name"] if "human_name" in request.config else None
```
norton120, line 45: @sarahwooders I'm bringing this over but all of this should be part of the schema - I can merge your branch into this one and use `AgentCreate` (I think that's what you said it'll be called) instead of this, or we can just reconcile during the merge into v1

PR #1547, opened 2024-07-16
dependency injection

norton120:
This is related to [spreadsheet id #1](https://docs.google.com/spreadsheets/d/1MdezkfONSkRPeMNhPrIX1ClBkXN2ZHrh6DpSIS-Jq4o/edit?gid=0#gid=0)

The goal here is to move a version-managed router with dependency injection for all the deps. 

## Why is dependency injection good here? 
pragmatically, it makes the code testable. Without injection points testing the code is extremely complex and brittle, and in turn adding new features is super slow and difficult. Injecting the db, actor etc makes those things trivial at test time. 

#### 1. Dependency injection
for elements like the sync server and the current user that should be injected at runtime, we are using callables that return the assets from utilities. (note: at the moment utilities are scattered, they should live in a single module soon). This means that in testing these deps can easily be swapped out for a mocked server or a seeded user, so the test code _only_ tests the mechanics under test (and not all the setup boilerplate). 

#### 2. API Versioning
You'll see a pattern in `app.py` that is aliasing the `v1` version folder. The idea is whichever the latest version of the API is, that will respond to: 
- the bare route
- the route by name (ie `/v2/users`)
- the route with latest (ie `/latest/users`)
this means breaking changes to the API are not breaking, as long as the consumers pinned their routes. It also makes parallel route versions extremely clear in the code, with `routes/v1/users.py` as the v1 route, and `routes/v2/users.py` as the v2. 

#### 3. No User Auth
this needs to get done and finished next, but isn't happening in this PR. bits of semi-auth are scattered all over and they need to be centralized and refined and then applied as one correct auth scheme, so this PR will go only so far as to cut them out when they are in the way.


norton120:
OOPS -there's already a PR for this #1537

PR #1554, opened 2024-07-18
feat: add schemas 

sarahwooders:


memgpt/schemas/block.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class Block(BaseModel):
+    """Block of the LLM context"""
+
+    description: Optional[str] = Field(None, description="Description of the block.")
+    limit: int = Field(2000, description="Character limit of the block.")
+    value: Optional[Union[List[str], str]] = Field(None, description="Value of the block.")
+
+    def __setattr__(self, name, value):
```
norton120, line 15: there is a built-in for this in pydantic, `validate_assignment` https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.validate_assignment

memgpt/schemas/block.py
```diff
@@ -0,0 +1,51 @@
...
+        """Run validation if self.value is updated"""
+        super().__setattr__(name, value)
+        if name == "value":
+            # run validation
+            self.__class__.validate(self.dict(exclude_unset=True))
+
+    @validator("value", always=True)
+    def check_value_length(cls, v, values):
```
norton120, line 23: might want to do this as a [model validator set to "after"](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.model_validator) - that gets rid of this magic 2000 value. 
so basically
1. limit gets set and defaults to 2000 per the attribute
2. value is checked against limit, which is now the given value or 2k

otherwise you get into this icky spot where a developer changes the default on limit to 4k, but the validator is failing on 2500, and confusion ensues

memgpt/schemas/block.py
```diff
@@ -0,0 +1,51 @@
...
+        if v is not None:
+            # Fetching the limit from the values dictionary
+            limit = values.get("limit", 2000)  # Default to 2000 if limit is not yet set
+
+            # Check if the value exceeds the limit
```
norton120, line 28: ready for code golf? ⛳ 
you can duck type this. 

```python
def total_len(values):
    return sum([len(v) for v in values])

a = 'this is a string'
b = ['this is a string','this is also a string']
>> total_len(a)
16
>> total_len(b)
37
```
note you don't need to worry about handling types that aren't either a string or a list of strings, because you typed them in the object - that's pydantic's job, to make sure that by the time your business logic gets called the typing has been checked :)

memgpt/schemas/block.py
```diff
@@ -0,0 +1,51 @@
...
+            if isinstance(v, str):
+                length = len(v)
+            elif isinstance(v, list):
+                length = sum(len(item) for item in v)
+            else:
+                raise ValueError("Value must be either a string or a list of strings.")
+
```
norton120, line 35: I'd probably do something like 
```python
try:
    assert length >= limit, f"Edit failed: Exceeds {limit} character limit (requested {length})."
except AssertionError as e:
    # this is not my favorite pydantic trait; 
    # pydantic validators require a ValueError to be raised to be included as a validation error. They really should handle AssertionErrors IMO, but we do this to make them happy. 
    raise ValueError(e)
return v 
```
That way a value that is too long is included in the Pydantic validation errors instead of a different error you need to format and deal with.

memgpt/schemas/document.py
```diff
@@ -0,0 +1,14 @@
...
+from pydantic import BaseModel, Field
+
+
+class DocumentModel(BaseModel):
+    """Representation of a single document (broken up into `Passage` objects)"""
+
+    user_id: uuid.UUID = Field(..., description="The unique identifier of the user associated with the document.")
+    text: str = Field(..., description="The text of the document.")
+    data_source: str = Field(..., description="The data source of the document.")
+    id: uuid.UUID = Field(default_factory=uuid.uuid4, description="The unique identifier of the document.", primary_key=True)
```
norton120, line 13: a thing to think about - I've added stripe ids (strings) to the ORM, with _id as the underlying uuid powering them. We don't need to do this - I can remove them and go back to a plain uuid as the id for every object if you want. This is what stripe ids look like: 

```python
>> doc = Document(**some_values)
>> doc.id
'document-9454ea28-802e-4d7a-ba8a-863955fb7f3a'
>> doc._id
<UUID '9454ea28-802e-4d7a-ba8a-863955fb7f3a'>
```
The benefit is you can always clearly see if you have the right type of identifier. HUGE for debugging customer issues when you look at their API call and they are sending `persona-9454ea28-802e-4d7a-ba8a-863955fb7f3a` to the `get_agent` endpoint :) 

but it does mean all the schemas need to handle them. This is how I've done that in the past: 

```python
def id_regex_pattern(prefix: str):
    """ generates the regex pattern for a given id """
    return (
        r"^" + prefix + r"_"  # prefix string
        r"[a-fA-F0-9]{8}-"  # 8 hexadecimal characters
        r"[a-fA-F0-9]{4}-"  # 4 hexadecimal characters
        r"[a-fA-F0-9]{4}-"  # 4 hexadecimal characters
        r"[a-fA-F0-9]{4}-"  # 4 hexadecimal characters
        r"[a-fA-F0-9]{12}$"  # 12 hexadecimal characters
    )
    
class AddressSchema(RestModel):
    name: Optional[str] = ""
    company: Optional[str] = None
    address_1: str
    address_2: Optional[str] = None
    city: str
    state: StateShortcode
    urbanization: Optional[str] = None
    zipcode: str
    zipcode_ext: Optional[str] = None

class PersistedAddressSchema(AddressSchema):
    id: str = Field(
        ..., examples=["addr_1234-5678-9012-3456"], pattern=id_regex_pattern("addr")
    )
```

(so you know there's no magic, `RestModel` is the schema base class that does the camelCasing we've talked about elsewhere): 

```python
class RestModel(BaseModel):
    model_config = ConfigDict(
        # allows you to use the snake or camelcase names in your code (ie user_id or userId)
        populate_by_name=True,
        # allows you do dump a sqlalchemy object directly (ie PersistedAddress.model_validate(SQLAdress)
        from_attributes=True,
        # This uses pyhumps.case to change names to camelCase when writing to the API 
        alias_generator=case,
        # throw errors if attributes are given that don't belong
        extra="forbid",
    )
```

since `.id` is the stripe id on the ORM you wouldn't need to do anything fancy - just set it like a normal attribute when you are consuming a database representation

memgpt/schemas/job.py
```diff
@@ -0,0 +1,38 @@
...
+from datetime import datetime
+from enum import Enum
+from typing import Optional
+
+from pydantic import BaseModel, Field
+
+from memgpt.utils import get_utc_time
+
+
+class JobStatus(str, Enum):
```
norton120, line 11: 🪰 **nitt** - I like to move all my enums into a single `enums` module for organization, but some people treat them like a primitive and just have them wherever. Your call

memgpt/schemas/document.py
```diff
@@ -0,0 +1,14 @@
+import uuid
+from typing import Dict, Optional
+
+from pydantic import BaseModel, Field
+
+
+class DocumentModel(BaseModel):
```
norton120, line 7: now for the hard part :) 
if you take a look at the [Document ORM](https://github.com/cpacker/MemGPT/blob/1.0.0-pre/memgpt/orm/document.py) it has relationships for the `Organization` it belongs to, and the `Passages` that belong to it. When you build the `Document` object you'll probably want some of the info from those relationships, but not all of it. If you did the whole thing like this 

```python
class DocumentModel(BaseModel):
... 
organization: Organization
passages: List[Passage] 
```
you get infinite recursion (because `Organization` has `documents` which is a list including this object) and also an obese object full of stuff you don't need. 

so maybe you have a schema called `OrganizationSummary` that only includes the id and the name of the org, maybe the org's email domain? those seem useful without getting too verbose. Then it looks like this:

```python
class DocumentModel(BaseModel):
... 
organization: OrganizationSummary
passages: List[PassageSummary] 

>> doc = DocumentModel(**some_vals)
>> doc.organization.name
'Some Great Company'

>> doc.organization.email_domain
'somegreatcompany.org'

>> doc.organization.users
AttributeError: OrganizationSummary has no attribute 'users'

# just for good measure, get users from the form
>> sql_org = doc.organization.to_sqlalchemy(db_session)
>> sql_org.users
[<User 'Dave'>, <User 'Lucy'>, <User 'Saresh'>]

```
We don't need to get this perfect right away, but the idea is to balance giving the object enough useful information that you need all the time to avoid tons of trips back to the db, vs making the objects obese and adding relationship complexity you don't care about.

memgpt/schemas/block.py
sarahwooders: ah ok I think I tried to do this before, but didn't realize `validate_assignment=True` existed so it didn't update after I changed the values after the instance was created

memgpt/schemas/document.py
sarahwooders: I think we should keep all the schemas as trim as possible for now, and maybe add additional stuff as needed when we update `server.py`

PR #1584, opened 2024-07-28
feat: allow templated system messages

cpacker:

Refactors the way system messages are compiled to set the foundation for system message templating.

This PR allows for system messages that have the reserved variable `{CORE_MEMORY}` in them.
1. If `{CORE_MEMORY}` exists in the system prompt, the system prompt will be formatted such that the core memory block is injected there
2. If `{CORE_MEMORY}`  does not exist in the system prompt, it'll automatically be appended to the end of the system prompt (**this makes this PR backwards compatible**)

Future PRs should enable users to define their own variables via `user_defined_variables` (TBD where this should be stored in the DB), which can be used to 

Also, a future PR should enable other templating styles other than just f-strings, e.g. Mustache / jinja2 (or Handlebars via pybars). I've added this as a reminder:
```python
    template_format: Literal["f-string", "mustache", "jinja2"] = "f-string",
```

**How to test**

Try modifying `memgpt_chat.txt` to have `{CORE_MEMORY}` at the end:
```text
...
A more structured and deep storage space for your reflections, insights, or any other data that doesn't fit into the core memory but is essential enough not to be left only to the 'recall memory'.
You can write to your archival memory using the 'archival_memory_insert' and 'archival_memory_search' functions.
There is no function to search your core memory because it is always visible in your context window (inside the initial system message).

Base instructions finished.
From now on, you are going to act as your persona.

{CORE_MEMORY}
```

Then create a new agent (`memgpt run`) and check the memory block in system (`/dumpraw`).

**Have you tested this PR?**

Yes

**Related issues or PRs**

#880 #895

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -11,6 +11,7 @@ from memgpt.agent_store.storage import StorageConnector
 from memgpt.constants import (
     CLI_WARNING_PREFIX,
     FIRST_MESSAGE_ATTEMPTS,
+    IN_CONTEXT_MEMORY_KEYWORD,
     JSON_ENSURE_ASCII,
     JSON_LOADS_STRICT,
     LLM_MAX_TOKENS,
@@ -86,7 +87,6 @@ def compile_system_message(
     The following are reserved variables:
       - CORE_MEMORY: the in-context memory of the LLM
     """
-    IN_CONTEXT_MEMORY_KEYWORD = "CORE_MEMORY"
 
     if user_defined_variables is not None:
         # TODO eventually support the user defining their own variables to inject
@@ -113,10 +113,10 @@ def compile_system_message(
 
         # Catch the special case where the system prompt is unformatted
         if append_icm_if_missing:
-            # TODO support for mustache and jinja2
             memory_variable_string = "{" + IN_CONTEXT_MEMORY_KEYWORD + "}"
             if memory_variable_string not in system_prompt:
                 # In this case, append it to the end to make sure memory is still injected
+                # warnings.warn(f"{IN_CONTEXT_MEMORY_KEYWORD} variable was missing from system prompt, appending instead")
                 system_prompt += "\n" + memory_variable_string
 
         # render the variables using the built-in templater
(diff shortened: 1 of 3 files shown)


memgpt/agent.py
```diff
@@ -49,33 +50,86 @@
 from .errors import LLMError
 
 
-def construct_system_with_memory(
-    system: str,
-    memory: BaseMemory,
-    memory_edit_timestamp: str,
+def compile_memory_metadata_block(
```
cpacker, line 53: @sarahwooders a question perhaps best saved for a later PR - do we want to move this metadata construction into the memory modules?

basically, when we compile the system prompt, there are two strings that get dynamically created related to memory, the core memory string (`str(memory)`) and metadata about archival/recall. This function (`compile_memory_metadata_block`) is generating the metadata, but I'm not sure where we want to put this.

TBH probably fine to leave in here for now, and refactor again in the PR where we introduce the ability to add user variables to the compilation process.

memgpt/prompts/system/memgpt_chat_fstring.txt
```diff
@@ -0,0 +1,51 @@
...
+Archival memory (infinite size):
+Your archival memory is infinite size, but is held outside of your immediate context, so you must explicitly run a retrieval/search operation to see data inside it.
+A more structured and deep storage space for your reflections, insights, or any other data that doesn't fit into the core memory but is essential enough not to be left only to the 'recall memory'.
+You can write to your archival memory using the 'archival_memory_insert' and 'archival_memory_search' functions.
+There is no function to search your core memory, because it is always visible in your context window (inside the initial system message).
+
+Base instructions finished.
+From now on, you are going to act as your persona.
+
+{CORE_MEMORY}
```
cpacker, line 51: @sarahwooders an example of what f-stringing looks like in the system prompt case. Here `CORE_MEMORY` is a reserved variable. In the future, you should be able to create new variables in the dev portal and edit the system prompt to have them included.

We should also create some extra reserved variables like `CURRENT_TIME` / `CURRENT_DATE` that can be dynamically pulled from the LLM OS / agent code at every step.

sarahwooders (approved):
lgtm!

Issue #1522, opened 2024-07-07
Agents not showing in Dev Portal ( UI )


tiro2000:
**Describe the bug**
Agent already created using Client Code , given Id and other details like default ( that's happening even with Basic Agent ) and its showing in the backend ( sql lite, showing in the Agents table with all fields ( agent_id, user_id, name, persona , human, preset, llm config ( {"model": "gpt-4o", "model_endpoint_type": "openai", "model_endpoint": "https://api.openai.com/v1", "model_wrapper": null, "context_window": 8192} ) , 
- embeddings_config ( {"embedding_endpoint_type": "hugging-face", "embedding_endpoint": "https://embeddings.memgpt.ai", 
"embedding_model": "BAAI/bge-large-en-v1.5", "embedding_dim": 1024, "embedding_chunk_size": 300} ) . 

Yet in the UI dev portal its not showing some of those created agents.

**Please describe your setup**
- [ ] How did you install memgpt?
  git clone, then using docker build file, then started memgpt server , then I use the password created as in the client example to create Agents
- [ ] Describe your setup
  - Windows
  - How are you running `memgpt`? (`cmd.exe`/Powershell/Anaconda Shell/Terminal)
  - PowerShell Terminal From Windows ( using another Powershell Terminal from Visual source code to access the Server using client example )

**Screenshots**

**Additional context**
def main(): 
    # Create an admin client
    token = "3Uz8TeKj4OY-qdZAWi3R5A"  # Server Token
    admin = Admin(base_url="http://localhost:8283", token=token)

    client = create_client(base_url="http://localhost:8283", token='3Uz8TeKj4OY-qdZAWi3R5A')
    
    basic_agent = client.create_agent(
        name="basic_agent", 
    )

    print(f"Created agent: {basic_agent.name} with ID {str(basic_agent.id)}")

**MemGPT Config**
Please attach your `~/.memgpt/config` file or copy past it below.
[defaults]
preset = memgpt_chat
persona = sam_pov
human = basic

[model]
model = gpt-4
model_endpoint = https://api.openai.com/v1
model_endpoint_type = openai
context_window = 8192

[embedding]
embedding_endpoint_type = openai
embedding_endpoint = https://api.openai.com/v1
embedding_model = text-embedding-ada-002
embedding_dim = 1536
embedding_chunk_size = 300

[archival_storage]
type = chroma
path = C:\Users\king_\.memgpt\chroma

[recall_storage]
type = sqlite
[...]

PR #1300, opened 2024-04-26
ci: update workflows (add `autoflake` and `isort`)

cpacker:
**Please describe the purpose of this pull request.**

Follow-up to #1279

Developers should now use pre-commit to correctly stage the linters (`autoflake` -> `isort` -> `black`):

```sh
# install the `dev` package that has precommit/autoflake/isort/black
poetry install --all-extras

# set up pre-commit to run on all git commits
poetry run pre-commit install
```

To manually trigger `pre-commit`:
```sh
# wipe cache
poetry run pre-commit clean

# run pre-commit hooks across all files
poetry run pre-commit run --all-files
```

---

**How to test**

See passing workflows.

PR #1620, opened 2024-08-06
feat: create an admin return all agents route

4shub:
Attempts to address [this issue](https://github.com/cpacker/MemGPT/issues/1522) by adding a new endpoint that will return all agents, regardless of user behind an `/api/admin` endpoint.
 
**How to test**
You can pull down the code and run this command to test:

```
curl http://localhost:4200/api/admin/agents  -H "Authorization: Bearer $ADMIN_PASSWORD"
```

**Have you tested this PR?**
Yes

**Related issues or PRs**
[Please link any related GitHub [issues](https://github.com/cpacker/MemGPT/issues) or [PRs](https://github.com/cpacker/MemGPT/pulls).
](https://github.com/cpacker/MemGPT/issues/1522)

No

**Additional context**

### DIFF
--- a/memgpt/server/server.py
+++ b/memgpt/server/server.py
@@ -6,7 +6,6 @@ from datetime import datetime
 from functools import wraps
 from threading import Lock
 from typing import Callable, List, Optional, Tuple, Union
-
 from fastapi import HTTPException
 
 import memgpt.constants as constants
@@ -883,13 +882,19 @@ class SyncServer(LockingServer):
     # TODO make return type pydantic
     def list_agents_legacy(
         self,
-        user_id: uuid.UUID,
+        user_id: Optional[uuid.UUID] = None,
     ) -> dict:
         """List all available agents to a user"""
-        if self.ms.get_user(user_id=user_id) is None:
-            raise ValueError(f"User user_id={user_id} does not exist")
 
-        agents_states = self.ms.list_agents(user_id=user_id)
+
+        if user_id is None:
+            agents_states = self.ms.list_all_agents()
+        else:
+            if self.ms.get_user(user_id=user_id) is None:
+                raise ValueError(f"User user_id={user_id} does not exist")
+
+            agents_states = self.ms.list_agents(user_id=user_id)
+
         agents_states_dicts = [self._agent_state_to_config(state) for state in agents_states]
 
         # TODO add a get_message_obj_from_message_id(...) function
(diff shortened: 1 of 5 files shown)


memgpt/server/server.py
```diff
@@ -900,7 +905,7 @@ def list_agents_legacy(
         for agent_state, return_dict in zip(agents_states, agents_states_dicts):
 
             # Get the agent object (loaded in memory)
-            memgpt_agent = self._get_or_load_agent(user_id=user_id, agent_id=agent_state.id)
+            memgpt_agent = self._get_or_load_agent(user_id=agent_state.user_id, agent_id=agent_state.id)
```
4shub, line 908: I think this makes more sense ; to use the current agent state's user_id vs global user_id

cpacker (approved):
LGTM

cpacker (changes requested):
Actually sorry can you run the linter code first? Looks like that test is failing, otherwise the main logic LGTM though.

You can run the linter via precommit, see OP in this PR:
https://github.com/cpacker/MemGPT/pull/1300

norton120:
design note - typically the way you'd want to handle this is via auth. an admin auth token would return all agents on list, whereas a normal user auth token would return the agents scoped to them. then for list queries if it takes query strings the admin would be able to do `query?user=user-xxxx-xxxx-xxxx-xxxx` in the get request and scope it to the user. this keeps your API to CRUD + list instead of building one-off endpoints for use cases.

so maybe a one-off for now, but should definitely consider auth-scoped list endpoints to replace it

4shub:
> design note - typically the way you'd want to handle this is via auth. an admin auth token would return all agents on list, whereas a normal user auth token would return the agents scoped to them. then for list queries if it takes query strings the admin would be able to do `query?user=user-xxxx-xxxx-xxxx-xxxx` in the get request and scope it to the user. this keeps your API to CRUD + list instead of building one-off endpoints for use cases.
> 
> so maybe a one-off for now, but should definitely consider auth-scoped list endpoints to replace it



I am not sure if I agree here; the `/api/admin/agents` route while returning a list of all agents, is providing a different expected result than `/api/agents` based on your authentication token. If a non-admin user were to access `/api/agents` and were to receive nothing or a `403` but receive data when providing their own user id; then it's a standard crud either as the effect changes based on properties of header and not the url value itself.

If you wanted an alternative crud solution I'd say you want to map your endpoints like this:

`/api/<user-id> | self/agents` to return agents of a specific user
`/api/agents` to return all agents, this endpoint only available to admin accounts

Then we can add our query params to the end of it

cpacker (approved):

PR #1633, opened 2024-08-12
fix: refactor `send_message` API

cpacker:
- fix hanging `send_message` by wrapping `server.step` in try/catch + add a timeout to the generator


memgpt/server/rest_api/agents/message.py
```diff
@@ -23,21 +26,23 @@ async def send_message_to_agent(
...
     user_id: str,
-    role: str,
+    role: MessageRole,
     message: str,
     stream_steps: bool,
     stream_tokens: bool,
     chat_completion_mode: Optional[bool] = False,
     timestamp: Optional[datetime] = None,
+    # related to whether or not we return `MemGPTMessage`s or `Message`s
+    return_message_object: bool = True,  # Should be True for Python Client, False for REST API
```
cpacker, line 36: @sarahwooders this isn't properly connected anywhere yet - probably this should be set to `False` by default, and the Python + REST clients should set it to `True` manually

PR #1653, opened 2024-08-16
fix: patches to the API for non-streaming OAI proxy backends

cpacker:
Currently, the dev portal is written to always ask for a stream reply, and if the backend doesn't support streaming, we send back a very similar looking data streaming, but instead of having N chunks for N tokens, we just send back 1 chunk per message ("fake streaming").

This is as intended, however the backend is currently incorrectly coded - it sees the request for `stream_tokens`, and sets up a streaming tokens compatible interface to get ready to send the response. However, in the case of the free MemGPT inference server, this is set up as an OAI proxy (`llm_config.endpoint_type == 'openai'`), so the related LLM API code thinks that we can actually support streaming, and turns on the streaming listeners. However, the proxy backend does not actually support streaming.

We need to figure out how to properly handle this situation (perhaps by moving `streaming` into the `llm_config`). In the meantime, I'm adding this sad monkeypatch that turns off token streaming on API calls unless the API provider is OAI (the only one that we really support atm anyways).

### DIFF
--- a/memgpt/server/rest_api/interface.py
+++ b/memgpt/server/rest_api/interface.py
@@ -500,7 +500,7 @@ class StreamingServerInterface(AgentChunkStreamingInterface):
 
                     processed_chunk = {
                         "function_call": {
-                            # "id": function_call.id,
+                            "id": function_call.id,
                             "name": function_call.function["name"],
                             "arguments": function_call.function["arguments"],
                         },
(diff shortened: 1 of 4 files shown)


memgpt/server/rest_api/interface.py
```diff
@@ -500,7 +500,7 @@ def function_message(self, msg: str, msg_obj: Optional[Message] = None):
 
                     processed_chunk = {
                         "function_call": {
-                            # "id": function_call.id,
+                            "id": function_call.id,
```
cpacker, line 503: @goetzrobin this was required to get some of the zod/object logic to work - basically there was a mistake and this `id` is nested for when `stream_tokens` is true, but unnested when it's false.

The `refactor-api` branch updates the messages API to use pydantic models so this should get cleaned up.

Issue #1567, opened 2024-07-24
MemoryModule validation errors on agents stored in database


cpacker:
Via discord: https://discord.com/channels/1161736243340640419/1161736244074659893/1265168228699738144

When the agent somehow already saved a larger than allowed size, it will fail to get agents on portal since it wants to load and cache every one that exists:

```MemGPT.memgpt.server.server - ERROR - Error occurred while trying to get agent 5a6dac48-b0b0-4720-bf3d-0e50f6f1e3b4:
1 validation error for MemoryModule
value
  Value error, Edit failed: Exceeds 2000 character limit (requested 2187). [type=value_error, input_value="An adventurer in skyrim....dge if i stare too long", input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/value_error
Traceback (most recent call last):
  File "C:\Projects\MemGPT\memgpt\server\server.py", line 348, in _load_agent
    memgpt_agent = Agent(agent_state=agent_state, interface=interface, tools=tool_objs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Projects\MemGPT\memgpt\agent.py", line 165, in __init__
    self.memory = BaseMemory.load(self.agent_state.state["memory"])
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Projects\MemGPT\memgpt\memory.py", line 78, in load
    obj.memory[key] = MemoryModule(**value)
                      ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\johnf\AppData\Local\pypoetry\Cache\virtualenvs\pymemgpt-zmpdwjkq-py3.12\Lib\site-packages\pydantic\main.py", line 176, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for MemoryModule
value
  Value error, Edit failed: Exceeds 2000 character limit (requested 2187). [type=value_error, input_value="An adventurer in skyrim....dge if i stare too long", input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/value_error
[...]

PR #1668, opened 2024-08-19
fix: various patches for 0.3.24

cpacker:
## Tracking various patches to bugs in 0.3.24

- [x] (merge-able into main) Missing personas/humans on CLI flag triggers runtime error instead of clean exit
- [x] (merge-able into main) Make `memory.load()` clip overflow fields as the default behavior so fix issue in #1567


tests/test_memory.py
```diff
@@ -63,3 +80,26 @@ def test_memory_limit_validation(sample_memory):
...
+    # overflow the value
+    memory_dict["persona"]["limit"] = "x" * 2500
+
+    # by default, this should throw a value error
+    with pytest.raises(ValueError):
+        BaseMemory.load(memory_dict)
+
+    # if we have overflow protection on, this shouldn't raise a value error
+    # TODO
+    # new_memory = BaseMemory.load(memory_dict)
```
cpacker, line 105: TODO uncomment this once patch is in

PR #1673, opened 2024-08-20
feat: Modifying schema to support multi modal inputs.

rohit-rptless:
feat: support for multi-modal input - work in progress - schema change

Based on the interface that OpenAI provides.
User can provide a list of
[{"type": "", "image_url": ""}]
that gets passed to the model.

Modifying schema for adding Multimodal support - User can give text and image.

**How to test**
Not testable. Only for reviewing high level schema changes. Won't  be merged without more commits.

**Have you tested this PR?**
No.

### DIFF
--- a/memgpt/agent_store/db.py
+++ b/memgpt/agent_store/db.py
@@ -156,6 +156,7 @@ def get_db_model(
             # openai info
             role = Column(String, nullable=False)
             text = Column(String)  # optional: can be null if function call
+            mm_content = Column(JSON)  # optional: multi-modal input
             model = Column(String)  # optional: can be null if LLM backend doesn't require specifying
             name = Column(String)  # optional: multi-agent only
 
@@ -192,6 +193,7 @@ def get_db_model(
                     role=self.role,
                     name=self.name,
                     text=self.text,
+                    mm_content=self.mm_content,
                     model=self.model,
                     # tool_calls=[ToolCall(id=tool_call["id"], function=ToolCallFunction(**tool_call["function"])) for tool_call in self.tool_calls] if self.tool_calls else None,
                     tool_calls=self.tool_calls,
--- a/memgpt/agent_store/db.py
+++ b/memgpt/agent_store/db.py
@@ -155,8 +155,7 @@ def get_db_model(
 
             # openai info
             role = Column(String, nullable=False)
-            text = Column(String)  # optional: can be null if function call
-            mm_content = Column(JSON)  # optional: multi-modal input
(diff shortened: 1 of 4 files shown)


memgpt/agent_store/db.py
```diff
@@ -156,6 +156,7 @@ class MessageModel(Base):
             # openai info
             role = Column(String, nullable=False)
             text = Column(String)  # optional: can be null if function call
```
cpacker, line 158: Here we should probably delete
`text = Column(String)  # optional: can be null if function call`

and replace with
`content = Column(JSON) # optional: multi-modal input`

which in the pydantic model is `Optional[Union[str, List[MultiModalMessagePart]]]`

but in the database itself is stored as an optional JSON field

memgpt/agent_store/db.py
```diff
@@ -192,6 +193,7 @@ def to_record(self):
                     role=self.role,
                     name=self.name,
                     text=self.text,
```
cpacker, line 195: Similarly, should delete/deprecate `text` and replace with `content` which can be `None`, `str` (text-only), or `List[dict/MultiModalMessagePart]` (multi-modal).

memgpt/schemas/message.py
```diff
@@ -62,6 +62,8 @@ class Message(BaseMessage):
     id: str = BaseMessage.generate_id_field()
     role: MessageRole = Field(..., description="The role of the participant.")
     text: str = Field(..., description="The text of the message.")
+    # Field mm_content is only used when role is 'user'. It needs to be mapped to MultiModalMessage
```
cpacker, line 65: Same comment here - deprecating `text` and replacing with a new `content` that matches OpenAI

memgpt/schemas/message.py
```diff
@@ -223,8 +225,9 @@ def to_openai_dict(
 
         elif self.role == "user":
             assert all([v is not None for v in [self.text, self.role]]), vars(self)
+            content = self.mm_content if self.mm_content is not None else self.text
```
cpacker, line 228: Once we do the above (replace `text`) we should no longer need this

memgpt/schemas/openai/chat_completion_request.py
```diff
@@ -8,13 +8,15 @@ class SystemMessage(BaseModel):
     role: str = "system"
     name: Optional[str] = None
 
+class MultiModalMessage(BaseModel):
```
cpacker, line 11: We probably need to expand this to have `text` and `image_url`:

image

PR #1681, opened 2024-08-26
fix: patch tests for JSON parsers

sarahwooders:


cpacker (approved):
🙌🙌🙌

cpacker (changes requested):
LGTM pending tests, looks like there's one or two more wackamoles:
```
FAILED tests/test_base_functions.py::test_archival - NameError: name 'json_dumps' is not defined
FAILED tests/test_schema_generator.py::test_schema_generator_with_old_function_set - assert {'name': 'archival_memory_insert', 'description': 'Add to archival memory. Make sure to phrase the memory contents such that it can be easily queried later.', 'parameters': {'type': 'object', 'properties': None, 'required': ['content', 'request_heartbeat']}} == {'name': 'archival_memory_insert', 'description': 'Add to archival memory. Make sure to phrase the memory contents such that it can be easily queried later.', 'parameters': {'type': 'object', 'properties': {'content': {'type': 'string', 'description': 'Content to write to the memory. All unicode (including emojis) are supported.'}, 'request_heartbeat': {'type': 'boolean', 'description': "Request an immediate heartbeat after function execution. Set to 'true' if you want to send a follow-up message or run a follow-up function."}}, 'required': ['content', 'request_heartbeat']}}

```

cpacker (approved):

PR #1682, opened 2024-08-26
feat: autocommit linting

norton120:
Instead of failing, isort and black just fix the issues and commit them

**How to test**
This is a tough one. you could use ACT but that's a huge lift to test some CI scripts. So let's just try them in GH and see if they work.

**Have you tested this PR?**
As best as can be with github actions without setting up ACT which is a pain.


norton120:
I think this needs an adult to authorize because it is changing workflows

cpacker (approved):

PR #1683, opened 2024-08-26
refactor: move prompt string generation in `Memory` to a non-overloaded function

cpacker:
tl;dr `Memory.__str__()` is now `Memory.compile()`

Move the prompt string generation to a dedicated non-overloaded function, since it's confusing to have potential complex logic handled behind an overloaded `__str__`. [PLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Follow-up PR will modify prompt string generation logic to be via a templating engine (eg jinja2) if possible to enable easier visualization and modification of the prompt generation in the frontend (vs passing python code back and forth).

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -107,7 +107,7 @@ def compile_system_message(
             archival_memory=archival_memory,
             recall_memory=recall_memory,
         )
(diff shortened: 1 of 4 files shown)


sarahwooders (approved):
lgtm!

PR #1687, opened 2024-08-27
refactor: use `jinja2` templates for `Memory.compile` instead of writing Python code

cpacker:
### Additions

1. Apply PR #1683 to `RecallMemory` and `ArchivalMemory` (previous PR only did `CoreMemory`)
2. Modify the `compile` code in `CoreMemory` to use `jinja2` instead of having a Python loop

Re: (2) - this should make it much more straight forward to allow changing the way the 'prompt formatting' of the memory object when creating custom memory blocks on the dev portal, since the back-and-forth between the frontend and backend will be a `jinja2` template string, instead of Python code that needs to be interpreted.

### TODO

- [x] Serialize / persist the template string (so that it can be retrieved / modified with API calls)

### Testing

Added test to double-check that the new `jinja2` template is equivalent to the old Python `__repr__` code:
```sh
pytest tests/test_memory.py::test_memory_jinja2_template
```

Added test for a `Memory.set_template` function that validates if the template string is valid or not (is it good `jinja2` + does it compile?)
```sh
pytest tests/test_memory.py::test_memory_jinja2_set_template
```


memgpt/memory.py
```diff
@@ -350,6 +350,9 @@ def __len__(self):
         return len(self._message_logs)
 
     def __repr__(self) -> str:
+        raise NotImplementedError("Use the .compile() method instead")
```
cpacker, line 353: TODO remove

memgpt/memory.py
```diff
@@ -467,6 +470,9 @@ def date_search(self, start_date, end_date, count=None, start=None):
         return results_json, len(results)
 
     def __repr__(self) -> str:
+        raise NotImplementedError("Use the .compile() method instead")
```
cpacker, line 473: TODO remove

memgpt/memory.py
```diff
@@ -597,6 +603,9 @@ def search(self, query_string, count=None, start=None):
             raise e
 
     def __repr__(self) -> str:
+        raise NotImplementedError("Use the .compile() method instead")
```
cpacker, line 606: TODO remove

sarahwooders:
Maybe rename `template` to `prompt_template`?

memgpt/schemas/memory.py
```diff
@@ -1,34 +1,81 @@
...
     """Represents the in-context memory of the agent"""
 
-    # Private variable to avoid assignments with incorrect types
+    # Memory.memory is a dict mapping from memory block section to memory block.
     memory: Dict[str, Block] = Field(default_factory=dict, description="Mapping from memory block section to memory block.")
 
+    # Memory.template is a Jinja2 template for compiling memory module into a prompt string.
+    prompt_template: str = Field(
+        default="{% for section, module in memory.items() %}"
+        '<{{ section }} characters="{{ module.value|length }}/{{ module.limit }}">\n'
```
sarahwooders, line 22: maybe rename `module` to `block`, to be consistent?

memgpt/schemas/block.py
```diff
@@ -62,7 +62,7 @@ class Block(BaseBlock):
     """Block of the LLM context"""
 
     id: str = BaseBlock.generate_id_field()
-    value: str = Field(..., description="Value of the block.")
+    value: Union[str, List[str]] = Field(..., description="Value of the block.")
```
sarahwooders, line 65: I think handling `List[str]` is kind of annoying, since we have to handle those cases differently. and you can represent a List as a string.

sarahwooders: but might be fine to leave for now

cpacker: OK, so I'll drop support for it?

cpacker: "drop support" = correct the typing hints in `memory.py`

sarahwooders (approved):
lgtm!

PR #1692, opened 2024-08-27
fix: remove assert on streaming chunk type 

sarahwooders:
This probably warrants a closer look, since the assert shouldn't error, but removing the assert allows streaming to work on the REST API.


cpacker (changes requested):
Irrelevant commits?

PR #1696, opened 2024-08-29
fix: patch flagrant typing errors and silent bugs in `client.py`

cpacker:
Note: should cherry pick and rebase https://github.com/cpacker/MemGPT/pull/1696/commits/85e71e7c803d1583057e2b5002c6e61aa4fd5ee1


cpacker:
FYI @sarahwooders I think this is closed by your #1703

PR #1703, opened 2024-08-30
feat: add Python autodocs 

sarahwooders:
Adds Google style docstrings to important classes that we want documented. Also adds a script for generating markdown files for auto-documented classes: 
```
python docs/generate_docs.py
```


cpacker (changes requested):
can you merge in main and rerun poetry add? looks like there's a conflict due to the recent addition of `jinja2`

cpacker (approved):

PR #1720, opened 2024-09-05
feat: create an abstract agent type + cleanup typing errors in `memory.py`

cpacker:
This is in anticipation of adding future agent types, e.g. see: https://github.com/cpacker/MemGPT/pull/1700

I was going to rename symbol `Agent` to `MemGPTAgent`, but decided we may want to punt that until the integration branch is merged to avoid a huge file diff on this PR.

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -1,6 +1,7 @@
 import datetime
 import inspect
 import traceback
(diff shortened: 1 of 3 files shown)


sarahwooders (approved):
lgtm!

PR #1723, opened 2024-09-06
feat: add support for returning type `MemGPTMessage` from cursor `GET`

cpacker:

Currently the `GET` routes to fetch messages in an agent's buffer/history return lists with type `Message`.

However, the `POST` routes (to generate new messages) return type `MemGPTMessage` by default (when token streaming is off, they are able to return type `Message` if needed, as is done in the Python Client by default).

This discrepancy makes client-side handling of messages in a chat view pretty convoluted, since the client has to ingest both styles of message objects and consolidate them client-side (into one type or another, or a custom type).

**How to test**

```sh
pytest -s -vvv tests/test_server.py::test_get_messages_memgpt_format
```


memgpt/schemas/message.py
```diff
@@ -96,11 +104,90 @@ def to_json(self):
...
+                if status == "OK":
+                    status_enum = "success"
+                elif status == "Failed":
+                    status_enum = "error"
+                else:
+                    raise ValueError(f"Invalid status: {status}")
+            except json.JSONDecodeError:
+                raise ValueError(f"Failed to decode function return: {self.text}")
+            messages.append(
+                # TODO make sure this is what the API returns
```
cpacker, line 158: @cpacker Note: I should add this to the unit test

cpacker: Added

sarahwooders (approved):
lgtm!

PR #1725, opened 2024-09-06
feat: Wrap fastapi client with sync calls to allow sync requests directly against the app

norton120:
### Why? 
We want to keep the fastapi pure software integration for testing and local execution of the RESTClient, but since fastapi is async that means users would need to write all async code :( 

This PR puts a thin wrapper on the calls that demotes the run to blocking/synchronous so the client methods don't need to be async anymore. It allows the client to still function via software-to-software API (we don't need to stand up a webserver, serve one code base, then have the other codebase talk to it to translate)


norton120:
@sarahwooders @cpacker does this work for y'all? the add client tests are passing with it (those are the methods I converted as POC) - if this is good I'll do the rest and get it up

norton120:
This is ready to merge, better to fix the formatting and lining in #1728

jedibrillo (approved):

PR #1726, opened 2024-09-06
refactor: remove the concept of 'section' from the jinja2 prompt formatter

cpacker:
This is probably a cleaner way to view the prompt template.

Alternatively we can even remove the `memory` dict and instead make it a `List[Block]` (since we now have `get_block`)


sarahwooders (approved):
lgtm!

PR #1727, opened 2024-09-06
add fixes to try to add blocks

sarahwooders:

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**


memgpt/orm/tool.py
```diff
@@ -46,16 +46,16 @@ class Tool(SqlalchemyBase, OrganizationMixin):
     organization: Mapped["Organization"] = relationship("Organization", back_populates="tools", lazy="selectin")
 
     @classmethod
-    def read(cls, db_session: "Session", name: str) -> "Tool":
+    def read_by_name(cls, db_session: "Session", name: str) -> "Tool":
```
sarahwooders, line 49: @norton120 do we want to just do something like this, but also provide the actor? 
```
def read_by_name(cls, db_session: "Session", name: str, actor_id: str) -> "Tool": 
...
```

memgpt/orm/tool.py
```diff
@@ -46,16 +46,16 @@ class Tool(SqlalchemyBase, OrganizationMixin):
...
         if found := db_session.query(cls).filter(cls.name == name, cls.is_deleted == False).scalar():
             return found
         raise NoResultFound(f"{cls.__name__} with name {name} not found")
 
-    @classmethod
-    def read_by_id(cls, db_session: "Session", id: str) -> "Tool":
-        if found := db_session.query(cls).filter(cls.id == id, cls.is_deleted == False).scalar():
-            return found
-        raise NoResultFound(f"{cls.__name__} with id {id} not found")
+    # @classmethod
```
sarahwooders, line 54: I removed this because it seems duplicated with the parent class `read`, and I think its confusing to pretend that `Tool.name` is the id

memgpt/client/client.py
```diff
@@ -824,7 +824,9 @@ def __init__(
         if user_id:
```
norton120, line 824: a shorthand here to help readability: 
```python
self.user_id = user_id or self.server.get_default_user().id
```

memgpt/orm/tool.py
```diff
@@ -46,16 +46,25 @@ class Tool(SqlalchemyBase, OrganizationMixin):
...
     @classmethod
-    def read(cls, db_session: "Session", name: str) -> "Tool":
-        if found := db_session.query(cls).filter(cls.name == name, cls.is_deleted == False).scalar():
+    def read_by_name(cls, db_session: "Session", name: str, actor) -> "Tool":
+        # TODO: not sure if this actor_id filter is implemented correctly (Sarah)
+        # if found := db_session.query(cls).filter(cls.name == name, cls.is_deleted == False).filter(cls.created_by_id == actor_id).scalar():
+        #    return found
+
+        query = db_session.query(cls).filter(cls.name == name, cls.is_deleted == False)
+        query = cls.apply_access_predicate(query, actor, "read")
```
sarahwooders, line 55: is this what this should be, to filter to data that the actor has access to?

memgpt/orm/tool.py
norton120: if you want to be explicit about it without adding extra one-off methods you could also overload the parent read on Tool to take a `name` param instead. 
This is one of those things that functional programmers hate but overloading (extending the arity) this way is an OOP pattern as opposed to lots of single-use getters

norton120:
The migration included here will fail if there are old tools with duplicate names already in the db - truncating the tools table or just blowing away the db `rm -rf .persist/*` will fix it, and then once the migration runs it'll never happen again :) 

Objects are unique per org, and auth predicate allows pydantic or sqlalchemy Users (and derivatives)

PR #1729, opened 2024-09-07
fix: Fix from overly agressive black

norton120:
Black removed EVERYTHING that isn't directly called. but we meta stuff. bad black. This puts that stuff back in


sarahwooders (approved):
lgtm!

PR #1734, opened 2024-09-09
feat: add V1 route refactor from integration branch into separate PR

cpacker:
Goals:

- Move V1 routes and route refactors from `integration` branch into a separate PR
- This should further isolate the changes in `integration` (ideally removes 1-2k lines of diff off of the `integration` PR)
- Make sure all tests still work (only thing that should break is the dev portal)

Copy over files from `v1/__init__.py` and patch import/syntax errors:
- [x] tools
- [x] sources
- [x] agents
- [x] llm_router
- [x] ~personas_router~ (deleted, @sarahwooders says these are deprecated in favor of block routes)
- [x] ~humans_router~ (deleted, @sarahwooders says these are deprecated in favor of block routes)
- [x] blocks_router

Connect everything together:
- [x] Configure new entry point to point at `v1`
- [x] Drop old files
  - Note: I left in `admin` and `auth` since it wasn't clear how they fit into `v1` + removing them breaks dev portal functionality
  - (We should strip these in a future PR or clarify the role of `admin`)
- [x] Get `RESTClient` unit tests to pass
  - Note: various things will break due to rewrites from `post` -> `patch`

---

Important notes for @sarahwooders // @norton120:
- The API routes (in V1) no longer have the concept of passing in a user in the request body / query params
  - Instead, the user is generated via a stub method `actor = server.get_current_user()`
  - This currently just grabs the default user (and if it doesn't exist, it creates it)
  - **This should be updated to instead get the `user_id` from the request header**
    - (And if the `user_id` doesn't exist in the header, then just use the default user like now)
- This PR does not copy over most of the changes from `app.py`, and instead just patches the old `rest_api/server.py` to work with the new routes
  - We could add this in a split PR
- Re: OpenAI routes under the new `/routes`
  - I added an extra level of directories, `assistants` and `chat_completions`, since I agree that `threads.py` and `assistants.py` were better split up, but I also felt they should be "bundled" into a single directory together if split up
  - Directory structure is now `routers/[v1/openai]`, `routers/openai/[assistants/chat_completions]`


memgpt/server/server.py
```diff
@@ -1770,3 +1770,38 @@ def retry_agent_message(self, agent_id: str) -> List[Message]:
         # Get the current message
         memgpt_agent = self._get_or_load_agent(agent_id=agent_id)
         return memgpt_agent.retry_message()
+
+    # TODO(ethan) wire back to real method in future ORM PR
+    def get_current_user(self) -> User:
```
cpacker, line 1775: @sarahwooders @norton120 

FYI this is the stub method I'm talking about in the OP

memgpt/server/rest_api/routers/v1/tools.py
```diff
@@ -0,0 +1,105 @@
...
+@router.post("/", response_model=Tool)
+def create_tool(
+    tool: ToolCreate = Body(...),
+    update: bool = False,
+    server: SyncServer = Depends(get_memgpt_server),
+):
+    """
+    Create a new tool
+    """
+    print("RUNNING CREATE TOOL")
```
sarahwooders, line 83: cleanup

memgpt/server/rest_api/server.py
norton120, line 1: Is this file still used at all? I believe the startup etc points to app.py now and this is a dead file, correct?

norton120 (approved):
Ideally remove the dead server.py - but either way let's get this in as soon as possible to remove the parallel routes issue! 🥳

memgpt/server/rest_api/server.py
cpacker: To get the tests to pass it was easier start by monkeypatching the new routes into an updated `server.py`.

I'm thinking a follow-up PR can delete `server.py` and add in `app.py` to replace it (I personally was encountering a lot of bugs trying to start with `app.py` so I figured might as well split the PRs if I can get `server.py` to pass tests.

cpacker:
FYI something weird is going on with the static files. If I mount on `/`, it breaks a ton of the API routes. If I do that, but also add some code to catch redirects to try and avoid collisions, it fixes most of the units tests but still breaks some.

So I just modified the mounting to be at `/app` for now, since I really want to get this PR in ASAP and this PR will break the dev portal regardless:
```python
def mount_static_files(app: FastAPI):
    static_files_path = os.path.join(os.path.dirname(importlib.util.find_spec("memgpt").origin), "server", "static_files")
    if os.path.exists(static_files_path):
        app.mount(
            # "/",
            "/app",
            SPAStaticFiles(
                directory=static_files_path,
                html=True,
            ),
            name="spa-static-files",
        )

```

sarahwooders (approved):
lgtm!

PR #1738, opened 2024-09-09
feat: return `MemGPTMessage` for Python Client 

sarahwooders:

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**


memgpt/server/server.py
```diff
@@ -889,7 +889,7 @@ def get_blocks(
         self,
         user_id: Optional[str] = None,
         label: Optional[str] = None,
-        template: Optional[bool] = True,
+        template: Optional[bool] = None,
```
cpacker, line 892: Is this actually support be be optional bool?

As in we expect `None`, `False`, or `True` (and `None` by default)?

Or is this meant to be `bool = True`, as in `False` or `True` (no `None`), but `False` by default?

sarahwooders: Yeah, I was having issues with getting confused by the defaults being passed - and we also do filtering based on what values are non-null.

cpacker (approved):

memgpt/client/client.py
```diff
@@ -1704,9 +1721,18 @@ def send_message(
         messages = self.interface.to_list()
         for m in messages:
             assert isinstance(m, Message), f"Expected Message object, got {type(m)}"
-        return MemGPTResponse(messages=messages, usage=usage)
 
-    def user_message(self, agent_id: str, message: str) -> MemGPTResponse:
+        # format messages
+        if include_full_message:
```
cpacker, line 1726: TODO should change to a consistent enum since we do it a lot

PR #1740, opened 2024-09-10
feat: move from `server.py` to `app.py`

cpacker:
Minimum diff to move from `server.py` to `app.py` (from `integration` branch) on `main` (with tests passing).


memgpt/server/rest_api/app.py
```diff
@@ -0,0 +1,169 @@
...
+        server.save_agents()
+        server = None
+
+    return app
+
+
+app = create_application()
+
+
+def start_server(
```
cpacker, line 145: Note @norton120 I added back this convenience method since we were calling `start_server` in several places outside of `cli.py` (primarily unit tests), and with the extra logging code added copy-pasting this everywhere would have been a lot of code duplication

memgpt/server/rest_api/app.py
```diff
@@ -0,0 +1,169 @@
...
+from memgpt.server.rest_api.routers.v1.users import (
+    router as users_router,  # TODO: decide on admin
+)
+from memgpt.server.rest_api.static_files import mount_static_files
+from memgpt.server.server import SyncServer
+from memgpt.settings import settings
+
+# TODO(ethan)
+# NOTE(charles): @ethan I had to add this to get the global as the bottom to work
+interface: StreamingServerInterface = StreamingServerInterface
```
cpacker, line 40: Note @norton120 I had to add these back since they weren't (afaik) getting initialized in global space anywhere else

memgpt/server/rest_api/app.py
```diff
@@ -0,0 +1,169 @@
...
+    router as openai_assistants_router,
+)
+from memgpt.server.rest_api.routers.openai.assistants.threads import (
+    router as openai_threads_router,
+)
+from memgpt.server.rest_api.routers.openai.chat_completions.chat_completions import (
+    router as openai_chat_completions_router,
+)
+
+# from memgpt.orm.utilities import get_db_session  # TODO(ethan) reenable once we merge ORM
```
cpacker, line 29: Note @norton120 @sarahwooders the only ORM specific code (I think?) that I had to comment out of this file

memgpt/server/rest_api/app.py
```diff
@@ -0,0 +1,169 @@
...
+from memgpt.server.rest_api.static_files import mount_static_files
+from memgpt.server.server import SyncServer
+from memgpt.settings import settings
+
+# TODO(ethan)
+# NOTE(charles): @ethan I had to add this to get the global as the bottom to work
+interface: StreamingServerInterface = StreamingServerInterface
+server: SyncServer = SyncServer(default_interface_factory=lambda: interface())
+
+# TODO(ethan): eventuall remove
```
cpacker, line 43: Note @norton120 @sarahwooders not sure if the concept of a server password really applies anymore? I guess we are using it to gatekeep the admin client? At the same time, admin client should be usable without any gatekeeping in our new auth model for OSS, so probably should just remove this entirely right?

cpacker: (let's do the removal in a follow-up PR though)

memgpt/server/rest_api/app.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+
+    for route in v1_routes:
+        app.include_router(route, prefix=API_PREFIX)
+        # this gives undocumented routes for "latest" and bare api calls.
+        # we should always tie this to the newest version of the api.
+        app.include_router(route, prefix="", include_in_schema=False)
+        app.include_router(route, prefix="/latest", include_in_schema=False)
+
+    # NOTE: ethan these are the extra routes
```
cpacker, line 83: Note @norton120 I had to add these extra routes, some for admin/auth and others to wire in the OpenAI assistants + chatcompletions routes (not sure how / if they were being wired in before with the original `app.py`?

memgpt/cli/cli.py
```diff
@@ -304,9 +301,9 @@ def server(
     type: Annotated[ServerChoice, typer.Option(help="Server to run")] = "rest",
     port: Annotated[Optional[int], typer.Option(help="Port to run the server on")] = None,
     host: Annotated[Optional[str], typer.Option(help="Host to run the server on (default to localhost)")] = None,
-    use_ssl: Annotated[bool, typer.Option(help="Run the server using HTTPS?")] = False,
-    ssl_cert: Annotated[Optional[str], typer.Option(help="Path to SSL certificate (if use_ssl is True)")] = None,
-    ssl_key: Annotated[Optional[str], typer.Option(help="Path to SSL key file (if use_ssl is True)")] = None,
+    # use_ssl: Annotated[bool, typer.Option(help="Run the server using HTTPS?")] = False,
```
norton120, line 304: can we delete this commented code
it's still in git if we need it later

memgpt/server/rest_api/app.py
norton120: yep interfaces is a thing we need to dig into - I was under the impression that they were no longer used anywhere, yet they obviously are so that's a thing to tackle

memgpt/server/rest_api/app.py
norton120: yep, 100% this should go - meh auth is way more dangerous than no auth.

memgpt/server/rest_api/app.py
norton120: k we can revisit in a follow-up PR, ideally we'll roll them all into the same iteration so if we add a `V1.5` API it is just changing one value in here - but we can do that in a follow-up PR

memgpt/server/rest_api/app.py
```diff
@@ -0,0 +1,169 @@
...
+app = create_application()
+
+
+def start_server(
+    port: Optional[int] = None,
+    host: Optional[str] = None,
+    debug: bool = False,
+):
+    """Convenience method to start the server from within Python"""
+    if debug:
```
norton120, line 151: this should set the global logging singleton instead of creating a new logger. and all the other logging bits really belong in the logger as well -
Not critical atm though, follow-up fix

norton120 (approved):

PR #1743, opened 2024-09-10
fix: fix the static file mounting handler breaking the API

cpacker:


cpacker (commented):
FYI @norton120 @4shub this was the only solution I tried out of a handful that worked (also tried adding catch-all handler and various other refactors)

PR #1744, opened 2024-09-10
chore: update static files

4shub:
Fixes the issue with SSE


cpacker (approved):
Gonna merge in static files, can figure out the mount bug in a later PR

PR #1745, opened 2024-09-10
feat: support importing tools from LangChain 

sarahwooders:
Add partial support for adding LangChain tools (similar to how CrewAI tools are supported). This doesn't support tools with take in arguments, so we should add support for those later.


cpacker (approved):

cpacker (changes requested):
merge conflicts w/ main

cpacker (approved):
LGTM pending tests

PR #1746, opened 2024-09-10
fix: static files mounting bug

cpacker:
tldr of bug is:
- enable static file mounting, the website works fine (and API routes via the website work fine), but the RESTClient unit tests start failing with "Method Not Allowed", indicating that the request is getting hijacked by the static files mount
- disable static file mounting (can even try just commenting it out), and the RESTClient unit tests pass again

To check if the website is working: `memgpt server` then `localhost:8283` should spawn the SPA

To check if the RESTClient unit tests are working, run `pytest -s -vvv tests/test_client.py::test_agent`

### DIFF
--- a/memgpt/server/rest_api/app.py
+++ b/memgpt/server/rest_api/app.py
@@ -72,8 +72,6 @@ def create_application() -> "FastAPI":
         allow_methods=["*"],
         allow_headers=["*"],
     )
-    # Usage in your create_application function:
-    app.add_middleware(SmartStaticFilesMiddleware)
 
     for route in v1_routes:
         app.include_router(route, prefix=API_PREFIX)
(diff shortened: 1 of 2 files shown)


memgpt/server/rest_api/app.py
```diff
@@ -58,29 +55,6 @@
 OPENAI_API_PREFIX = "/openai"
 
 
-class SmartStaticFilesMiddleware(BaseHTTPMiddleware):
```
cpacker, line 61: note @4shub here's the cleanup

memgpt/server/rest_api/app.py
```diff
@@ -123,7 +95,7 @@ def create_application() -> "FastAPI":
     app.include_router(setup_auth_router(server, interface, password), prefix=API_PREFIX)
 
     # / static files
-    # mount_static_files(app)
+    mount_static_files(app)
```
cpacker, line 98: This line causes unit tests to fail, but without it we lose the SPA

PR #1751, opened 2024-09-12
fix: server memory leak

cpacker:
We're creating a new server instance for every request right now. Hotfix to pull from globals until we properly strip state out of the server.


Sapessii:
Hi @cpacker, thank you for the instant fix :) is there an eta also for the fix of the pooling session of the db? Thank you!

cpacker:
> Hi @cpacker, thank you for the instant fix :) is there an eta also for the fix of the pooling session of the db? Thank you!

Coming very soon! Hopefully next week

PR #1756, opened 2024-09-13
refactor: clean up `agent.step()`

cpacker:
Initial stab at cleaning up the agent step by removing dead code + turning the response into a response model instead of a tuple.

Follow-up PRs should do the following cleanup:

- [ ] Unify the type going into `step` (probably should drop support for `str`, make it `None` or `Message`?)
- [ ] Move unnecessary kwargs into configs
- [ ] Strip out `ms` (replace with `session`)
- [ ] Move input validation out of `step` and into the loop handler
  - This ensures that `step` corresponds to a single LLM call (unless a summarization happens, in which case we probably want to return a dual usage? or put a summarization event into the response model?)
- [ ] Potentially change the input of step to be a model as well (probably cleaner if we don't do this)

### DIFF
--- a/memgpt/agent.py
+++ b/memgpt/agent.py
@@ -22,7 +22,7 @@ from memgpt.llm_api.llm_api_tools import create, is_context_overflow_error
 from memgpt.memory import ArchivalMemory, RecallMemory, summarize_messages
 from memgpt.metadata import MetadataStore
 from memgpt.persistence_manager import LocalStateManager
-from memgpt.schemas.agent import AgentState
+from memgpt.schemas.agent import AgentState, AgentStepResponse
 from memgpt.schemas.block import Block
 from memgpt.schemas.embedding_config import EmbeddingConfig
 from memgpt.schemas.enums import MessageRole, OptionState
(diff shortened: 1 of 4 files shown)


memgpt/agent.py
```diff
@@ -708,45 +704,21 @@ def _handle_ai_response(
...
         skip_verify: bool = False,
-        return_dicts: bool = True,  # if True, return dicts, if False, return Message objects
+        return_dicts: bool = True,
         recreate_message_timestamp: bool = True,  # if True, when input is a Message type, recreated the 'created_at' field
         stream: bool = False,  # TODO move to config?
         timestamp: Optional[datetime.datetime] = None,
         inner_thoughts_in_kwargs: OptionState = OptionState.DEFAULT,
         ms: Optional[MetadataStore] = None,
-    ) -> Tuple[List[Union[dict, Message]], bool, bool, bool]:
+    ) -> AgentStepResponse:
```
cpacker, line 717: Note: key diff

memgpt/schemas/agent.py
```diff
@@ -102,3 +104,14 @@ class UpdateAgentState(BaseAgent):
     # TODO: determine if these should be editable via this schema?
     message_ids: Optional[List[str]] = Field(None, description="The ids of the messages in the agent's in-context memory.")
     memory: Optional[Memory] = Field(None, description="The in-context memory of the agent.")
+
+
+class AgentStepResponse(BaseModel):
```
cpacker, line 109: Note: new schema

sarahwooders (approved):
lgtm!

PR #1763, opened 2024-09-21
fix: various fixes to get create agent REST API to work

cpacker:
Fixes required to get create agent on REST to work.

Note @4shub there's still that press-edit-then-value-is-broken error (seems client-side).


sarahwooders (approved):
lgtm!

PR #1786, opened 2024-09-25
feat: allow jobs to be filtered by `source_id`

sarahwooders:


letta/server/rest_api/routers/v1/jobs.py
```diff
@@ -12,14 +12,19 @@
...
     """
     List all jobs.
     """
     actor = server.get_current_user()
 
     # TODO: add filtering by status
-    return server.list_jobs(user_id=actor.id)
+    jobs = server.list_jobs(user_id=actor.id)
+    if source_id:
+        # can't be in the ORM since we have source_id stored in the metadata_
```
jedibrillo, line 25: [JSON columns types](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.JSON) are query-able in SQLAlchemy for both [PG](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#sqlalchemy.dialects.postgresql.JSON) and [SQLite](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#sqlalchemy.dialects.sqlite.JSON) dialects.

```
def list_jobs(self, user_id: str, source_id: Optional[str] = None) -> List[Job]:
        with self.session_maker() as session:
            if source_id:
                results = session.query(JobModel).filter(JobModel.user_id == user_id, JobModel.metadata_["source_id"].astext == sourced_id).all()
            else:
                results = session.query(JobModel).filter(JobModel.user_id == user_id).all()
            return [r.to_record() for r in results]
```

sarahwooders: yes added a note to refactor - thanks!

4shub (approved):

PR #1813, opened 2024-10-01
fix: Check that content is not None before setting to internal_monologue

mattzh72:
Check that content is not None before setting to internal_monologue. This is causing intermittent errors sometimes with our testing, where sometimes the model does not use the `content` field:

```
FAILED tests/test_tools.py::test_create_agent_tool[client0] - ValueError: Failed to send message: {"detail":"1 validation error for InternalMonologue\ninternal_monologue\n  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]\n    For further information visit https://errors.pydantic.dev/2.9/v/string_type"}
================== 1 failed, 2 passed, 36 warnings in 16.10s ===================
```

Additionally, we also search for `inner_thoughts` in the function args - community members noted that inner thoughts can sometimes appear in function args instead of the content field for function calls.

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -551,6 +551,10 @@ class Agent(BaseAgent):
             )  # extend conversation with assistant's reply
             printd(f"Function call message: {messages[-1]}")
 
+            if response_message.content:
+                # The content if then internal monologue, not chat
+                self.interface.internal_monologue(response_message.content, msg_obj=messages[-1])
+
             # Step 3: call the function
             # Note: the JSON response may not always be valid; be sure to handle errors
             function_call = (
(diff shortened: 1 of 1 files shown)


letta/agent.py
```diff
@@ -551,18 +551,15 @@ def _handle_ai_response(
             )  # extend conversation with assistant's reply
             printd(f"Function call message: {messages[-1]}")
 
-            # The content if then internal monologue, not chat
-            self.interface.internal_monologue(response_message.content, msg_obj=messages[-1])
```
cpacker, line 555: Can you replace this with what you have below:

```python
            if response_message.content:
                # The content if then internal monologue, not chat
                self.interface.internal_monologue(response_message.content, msg_obj=messages[-1])
```

letta/agent.py
```diff
@@ -607,6 +604,13 @@ def _handle_ai_response(
                 self.interface.function_message(f"Error: {error_msg}", msg_obj=messages[-1])
                 return messages, False, True  # force a heartbeat to allow agent to handle error
 
+            # Community noted that inner thoughts can sometimes appear in the function args
```
cpacker, line 607: And delete this code (if we're using `inner_thoughts` in `function_args`, it should be popped at a different level of the code (inside of the `llm_api.py` code)

cpacker (changes requested):
@mattzh72 can you revise to basically just add a conditional? So remove the added code, and modify the delete code to just be wrapped by a conditional

letta/agent.py
mattzh72: Ack

letta/agent.py
mattzh72: ah yes, sounds good!

cpacker (approved):
LGTM!

PR #1815, opened 2024-10-01
feat: Add Groq as provider option

mattzh72:
## Description
Add Groq as provider option in `llm_api`. We also delete the old `groq` API access code in local_llm.

Small fixes:
- This also makes the `test_endpoints.py` file more testable by creating a unique agent for the purpose of the test, and then cleaning up any instances of the agent before the test begins.
- Modify .gitignore to push configs since we are force pushing every file manually already anyway
- This also strongly types the model_endpoint_type to only the model types we currently support.

```
    model_endpoint_type: Literal["openai", "anthropic", "cohere", "google_ai", "azure", "groq", "ollama"] = Field(
```

## Testing
Added a simple test that hits the groq endpoint and asserts that a response is given. 

Also tested CLI flow e2e:
Screenshot 2024-10-03 at 10 06 13 AM


letta/llm_api/llm_api_tools.py
```diff
@@ -455,14 +454,55 @@ def create(
             chat_completion_request=ChatCompletionRequest(
                 model="command-r-plus",  # TODO
                 messages=[cast_message_to_subtype(m.to_openai_dict()) for m in messages],
-                tools=[{"type": "function", "function": f} for f in functions] if functions else None,
+                tools=tools,
```
cpacker, line 457: @mattzh72 this seems unrelated to `groq`?

letta/schemas/llm_config.py
```diff
@@ -17,7 +17,9 @@ class LLMConfig(BaseModel):
 
     # TODO: 🤮 don't default to a vendor! bug city!
     model: str = Field(..., description="LLM model name. ")
-    model_endpoint_type: str = Field(..., description="The endpoint type for the model.")
+    model_endpoint_type: Literal["openai", "anthropic", "cohere", "google_ai", "azure", "groq", "ollama"] = Field(
```
cpacker, line 20: Do we also want to support `local` in this literal? Since I think `local` is being used elsewhere in the code, also other legacy types like `webui`, `vllm`, etc (https://github.com/cpacker/MemGPT/blob/6d154bfcbd16bbeeb5fccf71cb59b9ecbe77407d/paper_experiments/utils.py#L31)

Maybe the best thing for now is to leave this as a non-literal string, but once we eventually test all legacy endpoints, revert?

tests/test_endpoints.py
```diff
@@ -108,3 +122,8 @@ def test_embedding_endpoint_ollama():
 def test_llm_endpoint_anthropic():
     filename = os.path.join(llm_config_dir, "anthropic.json")
     run_llm_endpoint(filename)
+
+
+def test_llm_endpoint_groq():
+    filename = os.path.join(llm_config_dir, "groq.json")
```
cpacker, line 128: We're missing `groq.json` in this PR I believe? At least I don't see it here or in `main`: https://github.com/cpacker/MemGPT/tree/main/configs/llm_model_configs

cpacker (changes requested):
see comments

letta/schemas/llm_config.py
cpacker: (I do generally agree that this should be a `Literal`, not a `str`, I'm just worried about breaking people old setups that are hooked up to eg LMStudio or local vLLM using old `model_endpoint_type` types

letta/llm_api/llm_api_tools.py
mattzh72: I just saw above:

```
if functions is not None:
            tools = [{"type": "function", "function": f} for f in functions]
            tools = [Tool(**t) for t in tools]
        else:
            tools = None
```

We are already instantiating tools, but then that variable never gets used. Just a little clean up.

letta/schemas/llm_config.py
mattzh72: Yeah, I can add other endpoint types, but I think migrating now is okay - I'll keep an eye on if people complain, and even if they do I think it's an pretty transparent fix on their end (just adding whatever key they're missing to the Literal).

tests/test_endpoints.py
mattzh72: Interesting...didn't get pushed. Not sure why.

mattzh72: It's in the gitignore actually

letta/schemas/llm_config.py
mattzh72: Adding:

```
"webui", "webui-legacy", "lmstudio", "lmstudio-legacy", "llamacpp", "koboldcpp", "vllm", "hugging-face"
```

configs/llm_model_configs/azure-gpt-4o-mini.json
```diff
@@ -0,0 +1,7 @@
+{
+    "context_window": 128000,
+    "model": "gpt-4o-mini",
+    "model_endpoint_type": "azure",
+    "model_endpoint": "https://letta.openai.azure.com/openai/deployments/gpt-4o-mini/chat/completions?api-version=2023-03-15-preview",
```
cpacker, line 5: Do we want this address to be in the OSS repo? @sarahwooders

cpacker: vs being use provided (or GH actions secrets provided)

cpacker (changes requested):
Can we modify the CLI code so that `groq` shows up as an official option in `letta configure`?

image

configs/llm_model_configs/azure-gpt-4o-mini.json
mattzh72: Removed for now

cpacker (approved):
@mattzh72 I had to make a few changes to get the path to work from the CLI.

It seems to be working on my end now, can you make sure it works e2e on your end, then merge?

Basically:
1. `rm -rf ~/.letta`
2. `export OPENAI_API_KEY=...` (to make sure embeddings work no problem)
3. `letta configure` (follow path to configure `groq` as the LLM provider, then choose the tool calling preview model
4. `letta run` (make sure that you get a response from an agent that has both inner thoughts + a message, eg:

image

PR #1831, opened 2024-10-05
feat: add agent types

vivek3141:
This PR creates an `AgentType` enum for the purposes of being able to create different kinds of agents. Funtionality is checked by a barebones test created in `tests/test_agent_type.py`.

For this modification the `AgentState` is modified to include an `agent_type` field, along with the `CreateAgent` schema that's sent to the server.


letta/client/client.py
```diff
@@ -68,6 +68,7 @@ def agent_exists(self, agent_id: Optional[str] = None, agent_name: Optional[str]
     def create_agent(
         self,
         name: Optional[str] = None,
+        agent_type: Optional[AgentType] = None,
```
sarahwooders, line 71: Could we have this explicitly default to the `AgentType.memgpt_agent`? Just to make it explicit what the defaults are

letta/schemas/agent.py
```diff
@@ -21,6 +22,15 @@ class BaseAgent(LettaBase, validate_assignment=True):
     user_id: Optional[str] = Field(None, description="The user id of the agent.")
 
 
+class AgentType(str, Enum):
+    """
+    Enum to represent the type of agent.
+    """
+
+    base_agent = "base_agent"
```
sarahwooders, line 30: I think this would be better named `memgpt_agent`, since `base_agent` might refer to a common base agent class

sarahwooders (changes requested):

letta/client/client.py
vivek3141: Fixed

letta/schemas/agent.py
vivek3141: Fixed!

sarahwooders (approved):
lgtm!

PR #1834, opened 2024-10-06
refactor: remove `get_current_user` and replace with direct header read

cpacker:
tl;dr refactor the way we get `user_id` from the header to be stateless (fixing a race condition bug)

---

Currently incoming requests get the `user_id` by polling the server for the current user with `get_current_user`, which reads from a variable:
```python
    # TODO(ethan) wire back to real method in future ORM PR
    def get_current_user(self) -> User:
        """Returns the currently authed user.

        Since server is the core gateway this needs to pass through server as the
        first touchpoint.
        """

        # Check if _current_user is set and if it's non-null:
        if hasattr(self, "_current_user") and self._current_user is not None:
            current_user = self.get_user(self._current_user)
            if not current_user:
                warnings.warn(f"Provided user '{self._current_user}' not found, using default user")
            else:
                return current_user

        return self.get_default_user()
```

This variable is set by middleware inside of `rest_api.app`:
```python 
    @app.middleware("http")
    async def set_current_user_middleware(request: Request, call_next):
        user_id = request.headers.get("user_id")
        if user_id:
            try:
                server.set_current_user(user_id)
            except ValueError as e:
                # Return an HTTP 401 Unauthorized response
                # raise HTTPException(status_code=401, detail=str(e))
                return JSONResponse(status_code=401, content={"detail": str(e)})
        else:
            server.set_current_user(None)
        response = await call_next(request)
        return response
```

This code is pretty bad since it clearly allows for race conditions (this was meant to be temporary code during the integration refactor).

This PR refactors the routes to instead do a simple read of the `user_id` from the request headers. If the client did not ser the `user_id` in the header, then the server will simply use the default user.

### DIFF
--- a/letta/server/rest_api/admin/tools.py
+++ b/letta/server/rest_api/admin/tools.py
@@ -26,7 +26,6 @@ class CreateToolResponse(BaseModel):
 
 
 def setup_tools_index_router(server: SyncServer, interface: QueuingInterface):
-    # get_current_user_with_server = partial(partial(get_current_user, server), password)
 
     @router.delete("/tools/{tool_name}", tags=["tools"])
     async def delete_tool(
--- a/letta/server/rest_api/app.py
+++ b/letta/server/rest_api/app.py
@@ -5,8 +5,7 @@ from pathlib import Path
 from typing import Optional
 
 import uvicorn
-from fastapi import FastAPI, Request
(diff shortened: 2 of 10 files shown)


cpacker:
Tagging @4shub as an FYI

sarahwooders (approved):

Issue #1825, opened 2024-10-03
Security issue when create agent is called from different user simultaneously


Sapessii:
**Describe the bug**

Hi, when the create agent endpoint is called simultaneously from different users, the user_id is always stick with the first user, hence resulting in multiple users that can access to the chat of others.

I think it's related with the middleware setting the set_current_user function is called.

cpacker:
Thanks for bringing this up @Sapessii ! Just opened a fix in PR #1834 - once it passes test will get it merged in ASAP.

Sapessii:
Awesome thank you so much!

Sapessii:
Hi! @cpacker, I just tried the fix but is not getting the header correctly, it's working when using

`user_id: str = Header(..., alias="user_id")`

Sapessii:
@cpacker sure no problem, I just took it from an older version of memgpt :) thanks to you!

Sapessii:
@cpacker sure no problem :) thanks to you!

cpacker:
@Sapessii can you check my comments in https://github.com/cpacker/MemGPT/pull/1843?

I think this seems to be working (at least in that PR). Will merge it in shortly.

Sapessii:
@cpacker working perfectly! thank you so much

PR #1843, opened 2024-10-08
fix: patch `user_id` in header

cpacker:
https://github.com/cpacker/MemGPT/issues/1825#issuecomment-2399779521

---

`letta server`

When you use an invalid `user_id` that doesn't exist:
```sh
% curl --request GET \
  --url http://localhost:8283/v1/agents/ \
  --header 'user_id: banana'
{"detail":"User with id banana not found"}
```

When you don't specify a `user_id` (server uses the default):
```sh
% curl --request GET \
  --url http://localhost:8283/v1/agents/
[{"description":null,"metadata_":{"human":"basic","persona":"sam_pov"},"user_id":"user-00000000","id":"agent-16ffdc6b-fc6b-42...
```

When you specify a `user_id` (here I'm using the default):
```sh
% curl --request GET \
  --url http://localhost:8283/v1/agents/ \
  --header 'user_id: user-00000000'
[{"description":null,"metadata_":{"human":"basic","persona":"sam_pov"},"user_id":"user-00000000","id":"agent-16ffdc6b-fc6b-427d-aeac
```

### DIFF
--- a/letta/server/server.py
+++ b/letta/server/server.py
@@ -1921,7 +1921,10 @@ class SyncServer(Server):
         if user_id is None:
             return self.get_default_user()
         else:
-            return self.get_user(user_id=user_id)
+            try:
+                return self.get_user(user_id=user_id)
+            except ValueError:
+                raise HTTPException(status_code=404, detail=f"User with id {user_id} not found")
 
     def list_llm_models(self) -> List[LLMConfig]:
         """List available models"""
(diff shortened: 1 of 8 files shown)


letta/server/server.py
```diff
@@ -1921,7 +1921,10 @@ def get_user_or_default(self, user_id: Optional[str]) -> User:
         if user_id is None:
             return self.get_default_user()
         else:
-            return self.get_user(user_id=user_id)
+            try:
+                return self.get_user(user_id=user_id)
+            except ValueError:
+                raise HTTPException(status_code=404, detail=f"User with id {user_id} not found")
```
cpacker, line 1927: New behavior: if you include `user_id` in the headers but the `user_id` doesn't exist, we throw a 404 (not found).

We could alternatively throw an auth error, but it seems 404 is more appropricate?

sarahwooders (approved):

PR #1844, opened 2024-10-08
feat: prompting o1

kl2806:
Create a new agent type `O1Agent`, that uses prompting to get the agent to "think" more at inference time before responding by calling a think function repeatedly before the final answer function.

**How to test**
Added a simple test in `test_o1_agent` that creates an O1 agent and asks it to compare which two numbers are larger. With `got-4o-mini`, it improves accuracy from `0->80%`. The test just checks that 3 steps to respond.


kl2806:
Closing as it is subsumed by #1891

PR #1845, opened 2024-10-08
test: add complex testing for Groq Llama 3.1 70b

mattzh72:
## Description

Add complex testing for Groq Llama 3.1 70b. Along the way, we discovered some edge cases and made some modifications:

- Modify the parsing of `request_heartbeats` to parse string responses, i.e. `"true"` resolves to `True`. Before, it would resolve to `False`. 
- Due to the inconsistency of Llama, we created a new testing wrapper that will retry `n` times and pass on a threshold basis
- Modify the argument description for `request_heartbeat` to more accurate reflect the bool (observed some success boost)

## Testing
We hold off on adding GH Actions until Charles is available to add new env variables. However, we manually ran the new tests and confirmed each passed.


sarahwooders (approved):
lgtm!

PR #1847, opened 2024-10-08
feat: persist tools to db when saving agent

vivek3141:
When users add tools upon the creation of an agent, the tools are required to be persisted to the DB for functionality. This PR addresses this by saving un-saved tools to the DB while saving the agent.

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -238,6 +238,7 @@ class Agent(BaseAgent):
         assert isinstance(self.agent_state.memory, Memory), f"Memory object is not of type Memory: {type(self.agent_state.memory)}"
 
         # link tools
+        self.tools = tools
         self.link_tools(tools)
 
         # gpt-4, gpt-3.5-turbo, ...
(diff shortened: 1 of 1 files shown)


letta/agent.py
```diff
@@ -1343,6 +1344,10 @@ def save_agent(agent: Agent, ms: MetadataStore):
     else:
         ms.create_agent(agent_state)
 
+    for tool in agent.tools:
+        if ms.get_tool(tool_id=tool.id) is None:
```
sarahwooders, line 1348: You should actually get the tool by name like `self.ms.get_tool(tool_name=tool_name, user_id=user_id)` where I think you can get `user_id` from `agent_state.user_id`. 

This is because we don't allow duplicate names for tools, inside a user scope.

vivek3141: Fixed!

sarahwooders (approved):
lgtm! should be good to merge once tests pass

PR #1858, opened 2024-10-09
chore: add CLI CI test

mattzh72:
## Description

Add CLI CI test. This performs a basic test of retrieving an existing agent, and checks the the initial output is valid. A couple of notes:

- We have a pytest fixture that backs up and deletes the user's `.letta`, and then swaps the backup copy to the `.letta` directory after the test is done
- We rewrite the credentials file dynamically to prevent API key leakage
- We factor out all the decorative emojis to constants so we can ensure test consistency

## Testing
Ran the test locally and confirmed CI passes.


sarahwooders (approved):
lgtm!

PR #1861, opened 2024-10-09
fix: insert_many checks exists_ok

vivek3141:
`SQLLiteStorageConnector`'s `insert_many` function previously did not use the `exists_ok` parameter to allow for skipping records with already present IDs. This PR adds this functionality.


sarahwooders (approved):
lgtm!

PR #1863, opened 2024-10-10
fix: add missing hardcodings for popular OpenAI models

cpacker:
Before:
image

After:
image


letta/constants.py
```diff
@@ -75,6 +75,27 @@
 LLM_MAX_TOKENS = {
     "DEFAULT": 8192,
     ## OpenAI models: https://platform.openai.com/docs/models/overview
+    # "o1-preview
```
cpacker, line 78: Some of these I commented out since we haven't really tested them at all, eg o1 API I'm sure is not working

sarahwooders (approved):

PR #1875, opened 2024-10-13
fix: patch errors with `OllamaProvider`

cpacker:
`OllamaProvider` had some errors including runtime errors on the context window / embedding size fetch, also it didn't properly set the model wrapper / prompt formatter.


sarahwooders (approved):
lgtm!

PR #1879, opened 2024-10-13
refactor: simplify `Agent.step` inputs to `Message` or `List[Message]` only

cpacker:
Simplify `Agent.step` inputs to `Message` or `List[Message]` only

Note: previously `List` was not a valid input, despite the API docs making it seem like passing a list was possible

Added an additional convenience function `Agent.step_user_message` that wraps a user message (eg `"hello"`) in the JSON metadata (eg `{"type": "user_message", "message", "hi"}`)

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -39,6 +39,7 @@ from letta.system import (
     get_login_event,
     package_function_response,
     package_summarize_message,
+    package_user_message,
 )
 from letta.utils import (
     count_tokens,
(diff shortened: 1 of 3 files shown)


letta/agent.py
```diff
@@ -730,14 +722,13 @@ def _handle_ai_response(
 
     def step(
         self,
-        user_message: Union[Message, None, str],  # NOTE: should be json.dump(dict)
+        messages: Union[Message, List[Message]],
```
cpacker, line 725: @sarahwooders new input types:
```python
messages: Union[Message, List[Message]],
```

sarahwooders (approved):

PR #1884, opened 2024-10-14
refactor: make Agent.step() multi-step

cpacker:
This PR refactors `Agent.step` to be multi-step. What this means is that the code for handling the step loop (e.g. running follow-up inference / another `step` when `request_heartbeat` is true, or when the function execution failed - all MemGPT-specific ways to do agent looping) have been moved **into** the `Agent` class itself.

This makes the `Agent` interfaces more generic. By default now, `Agent` exposes the following:

```python
class Agent(BaseAgent):

    # NOTE: `step` is now called `inner_step`
    def inner_step(
        self,
        messages: Union[Message, List[Message]],
    ) -> AgentStepResponse:
        # This will do only one LLM inference call (two if a summarizer event runs)

    def step(
        self,
        messages: Union[Message, List[Message]],
    ) -> LettaUsageStatistics:
        # This runs a loop where we run `inner_step` many times
```

As a side note, maybe we should rename `AgentStepResponse` to `MemGPTAgentStepResponse`, since it's pretty MemGPT-specific?
```python
class AgentStepResponse(BaseModel):
    # TODO remove support for list of dicts
    messages: Union[List[Message], List[dict]] = Field(..., description="The messages generated during the agent's step.")
    heartbeat_request: bool = Field(..., description="Whether the agent requested a heartbeat (i.e. follow-up execution).")
    function_failed: bool = Field(..., description="Whether the agent step ended because a function call failed.")
    in_context_memory_warning: bool = Field(
        ..., description="Whether the agent step ended because the in-context memory is near its limit."
    )
    usage: UsageStatistics = Field(..., description="Usage statistics of the LLM call during the agent's step.")
```

And maybe rename the `Agent` class in `agent.py` to `MemGPTAgent`?

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -11,13 +11,17 @@ from letta.agent_store.storage import StorageConnector
 from letta.constants import (
     CLI_WARNING_PREFIX,
     FIRST_MESSAGE_ATTEMPTS,
+    FUNC_FAILED_HEARTBEAT_MESSAGE,
     IN_CONTEXT_MEMORY_KEYWORD,
     LLM_MAX_TOKENS,
     MESSAGE_SUMMARY_TRUNC_KEEP_N_LAST,
     MESSAGE_SUMMARY_TRUNC_TOKEN_FRAC,
     MESSAGE_SUMMARY_WARNING_FRAC,
+    REQ_HEARTBEAT_MESSAGE,
 )
+from letta.errors import LLMError
 from letta.interface import AgentInterface
+from letta.llm_api.helpers import is_context_overflow_error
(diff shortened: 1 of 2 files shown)


letta/main.py
```diff
@@ -361,8 +361,10 @@ def run_agent_loop(
         skip_next_user_input = False
 
         def process_agent_step(user_message, no_verify):
+            # TODO(charles): update to use agent.step() instead of inner_step()
```
cpacker, line 364: This needs to be done eventually to support non-MemGPT agents on the Letta CLI / `letta run`

PR #1886, opened 2024-10-15
fix: Fix updating tools 

mattzh72:
## Description

Currently, adding tools in the dev portal causes all the existing tools to be removed, making the tools unusable. This is due to a bug, as adding tools is currently done by invoking `update_agent` with a `list` of the desired tool name, which completely replaces the entire tool list, instead of appending.

This PR introduces two new API endpoints:
- `add_tools_to_agent`: appends a list of tools to the existing tools for an agent
- `remove_tools_from_agent`: removes a list of tools from the existing tools for an agent

We fix a variety of bugs with the tools, for example, where descriptions are not being properly set on update.

## Testing
We write a comprehensive unit test that adds two tools step by step, confirms that both are in, and selectively removes one of the tools.


sarahwooders (approved):
lgtm!

PR #1893, opened 2024-10-15
feat: Add delete file from source endpoint

mattzh72:
## Description:

Add delete file from source endpoint. We introduce an endpoint that, given:

- `source_id`
- `file_id`

Can delete the file with `file_id` from the source with `source_id`. This endpoint returns a 204 code upon successful deletion.

## Testing:

Wrote a simple unit test covering adding a file and then deleting it.


sarahwooders (approved):
lgtm!

PR #1902, opened 2024-10-17
feat: Add `put_inner_thoughts_in_kwargs` as a config setting for the LLM

mattzh72:
## Description:

Add `put_inner_thoughts_in_kwargs` as a config that can be set by users per LLM config. This reduces the opacity of "deriving" this value or hardcoding this value in logic for different providers in the `create` function. 

## Testing:

Ran `test_endpoints.py` for a variety of LLM configs and confirmed passed.

We also run `poetry run pytest -s -vv tests/test_client.py::test_streaming_send_message` to test the `root_validator`:

> the `root_validator` auto-populates the `put_inner_thoughts_in_kwargs` field with some conditions on the model name. To perform that this works, we note that without the `root_validator`, the `test_streaming_send_message` fails, and with it, it passes, meaning that `gpt-4` has the field correctly set to `False`, not the default `True`. 

Finally, we run letta run with an existing agent to ensure that the llm configs can be correctly loaded in.

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -30,7 +30,7 @@ from letta.persistence_manager import LocalStateManager
 from letta.schemas.agent import AgentState, AgentStepResponse
 from letta.schemas.block import Block
 from letta.schemas.embedding_config import EmbeddingConfig
-from letta.schemas.enums import MessageRole, OptionState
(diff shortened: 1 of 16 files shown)


sarahwooders (approved):
lgtm!

PR #1920, opened 2024-10-22
fix: Add embedding tests to azure

mattzh72:
## Description:

Model listing was not working correctly on Azure beforehand - it was listing all available models, NOT the deployed models. We fix that in this PR and also patch bugs relating to using the Azure embedding endpoints.

## Testing

We add a test to create an embedding model and hit the Azure embedding endpoint.


sarahwooders (approved):
lgtm!

PR #1942, opened 2024-10-25
chore: Remove legacy code and instances of anon_clientid

mattzh72:
Remove legacy code and instances of anon_clientid. Rely on unit tests to test refactor.

Also, deleted two test files that have been completely commented out for a while.


tests/data/memgpt-0.2.11/config
```diff
@@ -31,6 +31,3 @@ path = /Users/sarahwooders/.letta
 [version]
 letta_version = 0.2.12
 
-[client]
```
sarahwooders, line 34: I think we want to keep this since its an example of an "old" config

mattzh72: done

sarahwooders (approved):

PR #1944, opened 2024-10-26
feat: Enable base constructs to automatically populate "created_by" and "last_updated_by" fields for relevant objects

mattzh72:
## Description:

Enable base constructs to automatically populate "created_by" and "last_updated_by" fields for relevant objects. We apply this to Tools, but this will set the foundation for future ORM refactors.

We also change it so we thread the entire actor (`User`) through the requests - this simplifies the API, and also allows us to do some simple access predicate checking where we confirm the actor of the request belongs to the organization the object (Tool, Agent) is associated with.

## Testing:

Ensured existing tests pass and added a new test case:

- Confirms that if multiple users on the same org, and a different user edits the Tool, then the `last_updated_by` field is updated but `created_by` is not

### DIFF
--- a/letta/cli/cli.py
+++ b/letta/cli/cli.py
@@ -219,7 +219,8 @@ def run(
 
         # create agent
         tools = [
-            server.tool_manager.get_tool_by_name_and_org_id(tool_name=tool_name, org_id=client.org_id) for tool_name in agent_state.tools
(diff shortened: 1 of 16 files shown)


sarahwooders (approved):

mattzh72:
no merge yet, fixing some small issues

PR #1950, opened 2024-10-28
fix: Clean up some legacy code and fix Groq provider

mattzh72:
## Description

Clean up some legacy code and fix Groq provider so models show up in list. Also, fix bug where the type was being set to openai, not groq.

## Testing

Added a simple unit test that checks at least one model of that type appears when calling `server.list_llm_models()`.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -262,21 +262,6 @@ def test_humans_personas(client: Union[LocalClient, RESTClient], agent: AgentSta
     assert human.value == "Human text", "Creating human failed"
 
 
-def test_config(client: Union[LocalClient, RESTClient], agent: AgentState):
-    # _reset_config()
-
-    models_response = client.list_models()
-    print("MODELS", models_response)
-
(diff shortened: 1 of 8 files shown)


tests/test_client.py
```diff
@@ -501,13 +486,3 @@ def test_message_update(client: Union[LocalClient, RESTClient], agent: AgentStat
 def test_organization(client: RESTClient):
     if isinstance(client, LocalClient):
         pytest.skip("Skipping test_organization because LocalClient does not support organizations")
-
-
-def test_model_configs(client: Union[LocalClient, RESTClient]):
```
sarahwooders, line 506: can we replace this with a test that tests `client.list_llm_configs()` and `client.list_embedding_configs()`?

sarahwooders (approved):

tests/test_client.py
mattzh72: done

PR #1954, opened 2024-10-29
feat: Implement tool calling rules for agents

mattzh72:
Implement tool calling rules for agents. Users can now define the tool calling pattern of the agent:

```python
class BaseToolRule(LettaBase):
    __id_prefix__ = "tool_rule"
    tool_name: str = Field(..., description="The name of the tool. Must exist in the database for the user's organization.")


class ToolRule(BaseToolRule):
    type: str = Field("ToolRule")
    children: List[str] = Field(..., description="The children tools that can be invoked.")


class InitToolRule(BaseToolRule):
    type: str = Field("InitToolRule")
    """Represents the initial tool rule configuration."""


class TerminalToolRule(BaseToolRule):
    type: str = Field("TerminalToolRule")
    """Represents a terminal tool rule configuration where if this tool gets called, it must end the agent loop."""
```

An example tool rule set would look like:

```python
tool_rules = [
    InitToolRule(tool_name="first_secret_word"),
    ToolRule(tool_name="first_secret_word", children=["second_secret_word"]),
    ToolRule(tool_name="second_secret_word", children=["third_secret_word"]),
    ToolRule(tool_name="third_secret_word", children=["fourth_secret_word"]),
    ToolRule(tool_name="fourth_secret_word", children=["send_message"]),
    TerminalToolRule(tool_name="send_message"),
]
```

This PR also introduces a couple other changes:

- Structured outputs for OpenAI non-streaming
- Retry mechanism for API flakes (choices returns None)
- Rework unit testing so all the core unit tests can run in parallel, decreasing our CI time 

## Testing:
I wrote a suite of complex unit tests:

- A static test suite that tests the functionality of the `ToolRuleSolver` class along with the ability to check the validity of tool rule sets
- A live agent test that actually implements a set of `ToolRule` and ensures that `gpt-4o` is able to follow the defined rules.


letta/schemas/agent.py
```diff
@@ -61,6 +62,9 @@ class AgentState(BaseAgent, validate_assignment=True):
     # tools
     tools: List[str] = Field(..., description="The tools used by the agent.")
 
+    # tool rules
```
sarahwooders, line 65: default to `None` instead?

sarahwooders (approved):
Looking good but we should also add: 
* removing the original `terminal=true` implementation 
* maybe some warning about tool rules only working for gpt-4o models (we can also add this to documentation)

letta/schemas/agent.py
mattzh72: done

PR #1960, opened 2024-10-31
chore: add migration script

4shub:
## Changelog
Adds migration script to alembic so code is not breaking


letta/server/server.py
```diff
@@ -181,13 +184,10 @@ def run_command(self, user_id: str, agent_id: str, command: str) -> Union[str, N
     # TODO: don't rely on config storage
     engine = create_engine("sqlite:///" + os.path.join(config.recall_storage_path, "sqlite.db"))
```
4shub, line 186: only sqllite will have databases automatically created, if you are using postgres you need to run alembic migrate

mattzh72 (approved):

PR #1978, opened 2024-11-04
feat: added ability to disable the initial message sequence during agent creation

cpacker:


tests/test_client.py
```diff
@@ -598,3 +600,75 @@ def test_shared_blocks(client: Union[LocalClient, RESTClient], agent: AgentState
...
+    yield created_agents
+    # Cleanup will run even if test fails
+    for agent_id in created_agents:
+        try:
+            client.delete_agent(agent_id)
+        except Exception as e:
+            print(f"Failed to delete agent {agent_id}: {e}")
+
+
+def test_initial_message_sequence(client: Union[LocalClient, RESTClient], agent: AgentState, cleanup_agents: List[str]):
```
cpacker, line 617: @sarahwooders tests should be self-explanatory for guidance on new kwarg `include_initial_boot_message`

letta/client/client.py
```diff
@@ -428,9 +429,18 @@ def create_agent(
             agent_type=agent_type,
             llm_config=llm_config if llm_config else self._default_llm_config,
             embedding_config=embedding_config if embedding_config else self._default_embedding_config,
+            initial_message_sequence=initial_message_sequence,
+        )
+
+        # Use model_dump_json() instead of model_dump()
```
cpacker, line 435: @sarahwooders I was getting a JSON serialization error here caused by the `datetime` inside of `Message.created_by` not getting serialized properly

It looks like the fix is to use `model_dump_json` instead of `model_dump`, to force the conversion to happen via `pydantic` instead of `requests` (`pydantic` understands that it needs to convert to ISO format because of the `to_json` inside of `Message`, but `requests` don't know how to use that)

letta/schemas/agent.py
```diff
@@ -105,14 +104,19 @@ class Config:
 class CreateAgent(BaseAgent):
     # all optional as server can generate defaults
     name: Optional[str] = Field(None, description="The name of the agent.")
-    message_ids: Optional[List[uuid.UUID]] = Field(None, description="The ids of the messages in the agent's in-context memory.")
+    message_ids: Optional[List[str]] = Field(None, description="The ids of the messages in the agent's in-context memory.")
```
cpacker, line 107: @sarahwooders stray `uuid`

sarahwooders (approved):

PR #1979, opened 2024-11-05
feat: Move Source to ORM model

mattzh72:
## Description:
Move source to ORM model.

## Testing
Adapt existing tests and write tests for managers.

Also, manually ran a backwards compatibility test for the alembic migration.


letta/schemas/source.py
```diff
@@ -34,29 +23,45 @@ class Source(BaseSource):
...
-    name: Optional[str] = Field(None, description="The name of the source.")
 
+class SourceCreate(BaseSource):
+    """
+    Schema for creating a new Source.
+    """
+
+    # required
+    name: str = Field(..., description="The name of the source.")
+    embedding_config: Optional[EmbeddingConfig] = Field(None, description="The embedding configuration used by the source.")
```
sarahwooders, line 52: Why `None` if required?

letta/services/source_manager.py
```diff
@@ -0,0 +1,108 @@
...
+from letta.schemas.source import SourceCreate, SourceUpdate
+from letta.schemas.user import User as PydanticUser
+from letta.utils import enforce_types, printd
+
+
+class SourceManager:
+    """Manager class to handle business logic related to Sources."""
+
+    # This is used when no embedding config is provided
+    DEFAULT_EMBEDDING_CONFIG = EmbeddingConfig(
```
sarahwooders, line 17: I think we should avoid having defaults (they've caused a lot of issues in the past) - we should just require it in `SourceCreate` and use that

letta/services/source_manager.py
```diff
@@ -0,0 +1,108 @@
...
+        from letta.server.server import db_context
+
+        self.session_maker = db_context
+
+    @enforce_types
+    def create_source(self, source_create: SourceCreate, actor: PydanticUser) -> PydanticSource:
+        """Create a new source based on the SourceCreate schema."""
+        with self.session_maker() as session:
+            # Provide default embedding config if not given
+            if not source_create.embedding_config:
```
sarahwooders, line 38: this should not be allowed

mattzh72: ack

letta/services/source_manager.py
mattzh72: ack

letta/schemas/source.py
mattzh72: ack

4shub (approved):

letta/server/rest_api/routers/v1/sources.py
```diff
@@ -64,21 +64,22 @@ def list_sources(
     """
     actor = server.get_user_or_default(user_id=user_id)
 
-    return server.list_all_sources(user_id=actor.id)
+    return server.list_all_sources(actor=actor)
```
sarahwooders, line 67: shouldn't this be using the `source_manager`?

mattzh72: no, there's extra logic tying it to agents

PR #1993, opened 2024-11-07
chore: add endpoint to update users

4shub:
## Changelog
Allow api access to update users

### DIFF
--- a/letta/server/rest_api/routers/v1/users.py
+++ b/letta/server/rest_api/routers/v1/users.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING, List, Optional
 from fastapi import APIRouter, Body, Depends, HTTPException, Query
 
 from letta.schemas.api_key import APIKey, APIKeyCreate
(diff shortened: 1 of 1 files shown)


mattzh72 (changes requested):
It seems like you're passing in a `User` object, not a `UserUpdate` object.

letta/server/rest_api/routers/v1/users.py
```diff
@@ -56,6 +56,18 @@ def create_user(
     return user
 
 
+@router.put("/", tags=["admin"], response_model=User, operation_id="update_user")
+def update_user(
+    user: UserUpdate = Body(...),
```
mattzh72, line 61: small nit, can we name this to user_update, otherwise lgtm

mattzh72 (approved):

letta/server/rest_api/routers/v1/users.py
mattzh72: we can do in future too, feel free to merge

PR #2012, opened 2024-11-07
feat: modify `startup.sh` to work for sqlite

sarahwooders:


4shub (changes requested):
Startup sh should also check for LETTA_PG_PORT/LETTA_PG_USER

4shub (approved):

PR #2029, opened 2024-11-13
fix: Fix memory summarization 

mattzh72:
## Description:

Fix memory summarization. Memory summarization was broken for models that put `inner_thoughts_in_kwargs`. 

> This is because for summarization, we construct a standard `user` `assistant` message flow that does not have inner thoughts, function calls, etc. When the `LLMConfig` requires `inner_thoughts_in_kwargs`, this causes the assistant message content to be `None`, which triggers an error on the LLM API.

## Testing:

Tested manually via CLI and added a CI test for the live gpt-4o capabilities test.

### DIFF
--- a/letta/memory.py
+++ b/letta/memory.py
@@ -51,7 +51,6 @@ def _format_summary_history(message_history: List[Message]):
 def summarize_messages(
     agent_state: AgentState,
     message_sequence_to_summarize: List[Message],
-    insert_acknowledgement_assistant_message: bool = True,
 ):
     """Summarize a message sequence using GPT"""
     # we need the context_window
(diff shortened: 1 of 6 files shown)


letta/memory.py
```diff
@@ -50,7 +51,6 @@ def _format_summary_history(message_history: List[Message]):
 def summarize_messages(
     agent_state: AgentState,
     message_sequence_to_summarize: List[Message],
-    insert_acknowledgement_assistant_message: bool = True,
```
sarahwooders, line 53: can you add a comment explaining what this variable is?

sarahwooders (approved):

PR #2046, opened 2024-11-16
fix: Fix summarizer for Anthropic and add integration tests

mattzh72:
Fix summarizer for Anthropic and add integration tests. Also rename several testing files for accuracy.


sarahwooders (approved):

mattzh72:
Did not wait for checks to pass bc previous PR finished all tests successfully except for one that was misconfigured, so should be safe to merge.

PR #2053, opened 2024-11-18
fix: context window overflow patch

cpacker:
Fix problem where when the user set a lower than native context window limit, the loop runner wouldn't trigger the summarize event correctly (even though the summarizer warning would still get injected).

### DIFF
--- a/letta/services/tool_manager.py
+++ b/letta/services/tool_manager.py
@@ -70,6 +70,10 @@ class ToolManager:
             pydantic_tool.organization_id = actor.organization_id
             tool_data = pydantic_tool.model_dump()
             tool = ToolModel(**tool_data)
+            # The description is most likely auto-generated via the json_schema,
+            # so copy it over into the top-level description field
+            if tool.description is None:
+                tool.description = tool.json_schema.get("description", None)
             tool.create(session, actor=actor)
 
         return tool.to_pydantic()
--- a/tests/test_summarize.py
+++ b/tests/test_summarize.py
@@ -1,8 +1,11 @@
 import uuid
+from typing import List
 
 from letta import create_client
+from letta.client.client import LocalClient
 from letta.schemas.embedding_config import EmbeddingConfig
 from letta.schemas.llm_config import LLMConfig
+from letta.schemas.message import Message
 
 from .utils import wipe_config
 
@@ -33,7 +36,7 @@ def create_test_agent():
     agent_obj = client.server._get_or_load_agent(agent_id=agent_state.id)
 
 
-def test_summarize():
+def test_summarize_messages_inplace():
     """Test summarization via sending the summarize CLI command or via a direct call to the agent object"""
     global client
     global agent_obj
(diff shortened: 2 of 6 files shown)


tests/test_tools.py
```diff
@@ -83,6 +83,8 @@ def test_create_tool(client: Union[LocalClient, RESTClient]):
 
     def print_tool(message: str):
         """
+        Example tool that prints a message
```
cpacker, line 86: @mattzh72 @sarahwooders FYI I think this was a bug in the test. If you looked through the test, `print_tool` was getting created with a null description, which I think we want to make impossible to happen because `description` is key to getting the LLM to understand the tool.

tests/test_tools.py
```diff
@@ -110,6 +112,7 @@ def print_tool(message: str):
     assert tool.id == tool.id, f"Expected {tool.id} to be {tool.id}"
 
     # create agent with tool
+    assert tool.name is not None, "Expected tool name to be set"
```
cpacker, line 115: (for typing purposes)

letta/services/tool_manager.py
```diff
@@ -70,6 +70,10 @@ def create_tool(self, pydantic_tool: PydanticTool, actor: PydanticUser) -> Pydan
             pydantic_tool.organization_id = actor.organization_id
             tool_data = pydantic_tool.model_dump()
             tool = ToolModel(**tool_data)
+            # The description is most likely auto-generated via the json_schema,
+            # so copy it over into the top-level description field
+            if tool.description is None:
```
cpacker, line 75: @mattzh72 @sarahwooders FYI I think this was a bug in the tool manager code.

If at this point the `description` is null, I think we should grab it from the `json_schema` to set it.

Honestly, we could change the following line (`tool.json_schema.get("description", None)`) to instead check if `"description" in tool.json_schema and tool.json_schema["description"] is not None`, since it's pretty problematic if this hasn't been set properly (LLM won't understand how to use the tool).

mattzh72: Sg, thanks for the catch!

mattzh72 (changes requested):
LGTM, but can we add a test in `tests/integration_test_summarizer.py` for this case?

tests/test_tools.py
mattzh72: I see - we had made description optional in the model before, so assumed it was okay to be null.

tests/test_summarize.py
```diff
@@ -33,7 +36,7 @@ def create_test_agent():
     agent_obj = client.server._get_or_load_agent(agent_id=agent_state.id)
 
 
-def test_summarize():
+def test_summarize_messages_inplace():
```
cpacker, line 39: @mattzh72 added test here

tests/test_summarize.py
```diff
@@ -73,3 +76,53 @@ def test_summarize():
     agent_obj.summarize_messages_inplace()
     print(f"Summarization succeeded: messages[1] = \n{agent_obj.messages[1]}")
     # response = client.run_command(agent_id=agent_obj.agent_state.id, command="summarize")
+
+
+def test_auto_summarize():
```
cpacker, line 81: actually sorry @mattzh72 here's the test

PR #2063, opened 2024-11-19
feat: insert `AgentState` into tool calls for stateful tools 

sarahwooders:
## Description:

Allow agent state to be serialized and passed into tools during execution time. This properly gets rid of the `self` required for tools. This gets injected into all tools, but not all tools need to use it.

## Testing

We test manually, but also write a suite of comprehensive integration tests for the sandboxes.


mattzh72 (approved):
Approved as this is merging into my branch, not main.

PR #2095, opened 2024-11-22
fix: add more types to the type_map from python -> json schema

cpacker:
Without this, the following tool can't get added (`list` doesn't work, neither does `List[str]`):

```python
def analyze_campaign(campaign_name: str, metrics: list, time_range: str) -> str:
    """
    Evaluate campaign performance metrics.

    Args:
        campaign_name (str): Name of the campaign.
        metrics (list): List of performance metrics to evaluate.
        time_range (str): Time range for analysis (e.g., 'last_week').

    Returns:
        str: Campaign performance summary.
    """
    import random
    performance_data = {
        "meeting_requests": random.randint(5, 20),
        "meetings_booked": random.randint(10, 30),
        "pipeline_generated": f"${random.randint(100000, 500000):,}",
        "closed_won": f"${random.randint(50000, 300000):,}"
    }
    performance_summary = ", ".join([f"{metric}: {performance_data.get(metric, 'N/A')}" for metric in metrics])
    return f"Campaign: {campaign_name} | {performance_summary} in {time_range}."
```

### DIFF
--- a/letta/functions/schema_generator.py
+++ b/letta/functions/schema_generator.py
@@ -56,7 +56,7 @@ def type_to_json_schema_type(py_type):
         None: "null",
         type(None): "null",
         # Optional types
-        Optional[str]: "string",
+        # Optional[str]: "string",  # NOTE: caught above ^
         Union[str, None]: "string",
     }
     if py_type not in type_map:
--- a/letta/functions/schema_generator.py
+++ b/letta/functions/schema_generator.py
@@ -1,5 +1,5 @@
 import inspect
-from typing import Any, Dict, Optional, Type, Union, get_args, get_origin
+from typing import Any, Dict, List, Optional, Type, Union, get_args, get_origin
 
 from docstring_parser import parse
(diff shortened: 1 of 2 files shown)


letta/functions/schema_generator.py
```diff
@@ -38,15 +38,29 @@ def type_to_json_schema_type(py_type):
...
+        tuple: "array",
+        set: "array",
+        # Dictionaries
+        dict: "object",
+        Dict[str, Any]: "object",
+        # Special types
+        None: "null",
+        type(None): "null",
+        # Optional types
+        Optional[str]: "string",
```
cpacker, line 59: @mattzh72 maybe it makes sense to kill this?
`Optional[str]: "string",`

because of the code above? `if is_optional(py_type):`

mattzh72: Hm yeah agreed - seems like above catches all the Optionals and unwraps them.

mattzh72 (approved):

PR #2102, opened 2024-11-22
chore: Change create_tool endpoint on v1 routes to error instead of upsert

mattzh72:
## Description

Change create_tool endpoint on v1 routes to error instead of upsert, and add an upsert endpoint on the routes. We also introduce a variety of improvements:

- Add client functions that are strictly create v.s. upsert
- Replace client usages of create with upsert across the codebase
- We do a better job handling dbapi errors in the core ORM and create custom errors for FK violations and Unique constraint violations 

## Testing
We add some new tests to the manager tests, alter some existing ones.


4shub (changes requested):
you need to provide a description; this is wh create is failing. Can you also provide a more clear response like a 409

4shub (approved):

PR #2109, opened 2024-11-26
feat: Add locking to send message

mattzh72:
## Description

This is a short-term fix for `send_message`. If multiple `send_message` calls are sent, due to some state in the interface, it errors. This prevents multiple `send_message` calls from executing at the same time for the same agent.

## Testing 

We write a test simulating concurrent `send_message` requests. Without the changes in this PR, it errors:

```
FAILED tests/test_client.py::test_send_message_parallel[client0] - Failed: Task 0 failed with exception: Failed to send message: {"detail":"Generator is inactive"}
```

### DIFF
--- a/tests/test_client_legacy.py
+++ b/tests/test_client_legacy.py
@@ -223,16 +223,6 @@ def test_core_memory(client: Union[LocalClient, RESTClient], agent: AgentState):
     assert "Timber" in memory.get_block("human").value, f"Updating core memory failed: {memory.get_block('human').value}"
 
 
-def test_messages(client: Union[LocalClient, RESTClient], agent: AgentState):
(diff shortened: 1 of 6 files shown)


tests/test_client_legacy.py
```diff
@@ -223,16 +223,6 @@ def test_core_memory(client: Union[LocalClient, RESTClient], agent: AgentState):
     assert "Timber" in memory.get_block("human").value, f"Updating core memory failed: {memory.get_block('human').value}"
 
 
-def test_messages(client: Union[LocalClient, RESTClient], agent: AgentState):
```
mattzh72, line 226: moved this test to `test_client.py` as it should pass for both local and rest

PR #2111, opened 2024-11-27
feat: Separate out streaming route

mattzh72:
## Description

Separate out a `messages` route for streaming and one for non-streaming. Also deprecated the following fields:

- return_message_object
- use_assistant_message

## Testing

Rely on existing tests to cover.


letta/client/client.py
```diff
@@ -2271,12 +2263,9 @@ def send_messages(
 
         # format messages
         messages = self.interface.to_list()
-        if include_full_message:
-            letta_messages = messages
-        else:
-            letta_messages = []
-            for m in messages:
-                letta_messages += m.to_letta_message()
+        letta_messages = []
```
sarahwooders, line 2266: shouldn't this be removed if we are passing back `LettaResponse` with `LettaMessage`?

sarahwooders:
Should we also edit the schema of `LettaResponse`?

letta/server/rest_api/interface.py
```diff
@@ -500,9 +496,6 @@ def _process_chunk_to_letta_style(
                     )
 
             elif self.inner_thoughts_in_kwargs and tool_call.function:
-                if self.use_assistant_message:
```
sarahwooders, line 503: why did we remove this?

letta/server/static_files/assets/index-9fa459a2.js
sarahwooders, line 1: should there be a diff here?

mattzh72: I did a mass replace of some parameters, so that's probably why

letta/server/rest_api/interface.py
mattzh72: i don't think this is a field that's set anymore - it's always set to False

mattzh72: since it is not passed in the API call

letta/client/client.py
mattzh72: done

letta/client/client.py
```diff
@@ -2338,16 +2321,13 @@ def send_message(
...
+        for m in messages:
+            letta_messages += m.to_letta_message()
```
sarahwooders, line 2326: shouldn't this be removed?

sarahwooders (approved):

PR #2112, opened 2024-11-27
feat: support pydantic models in tool uploads + patch dict/list types

cpacker:
- Added unit tests for our source code -> JSON schema pipeline
  - Covers both basic Python type + `List`/`Dict`
  - Also covers Pydantic types
- Based on the unit tests, did significant refactoring to the schema generator to patch bugs
  - (A lot was broken)

~Note @ @mattzh72  we could potentially add even more tests to the Python client to actually try connecting these tools to an agent and making sure the agent can call them, which would verify that the hardcoded reference schemas are actually correct, but since we're short on time I'll just manually test them (only a few functions) in the ADE to make sure the schemas are in fact correct.~

Edit I added tests where we put the schema in a payload to OpenAI and ask OpenAI to generate a call.

- [x] Manually test all the functions inside the `test_tool_schema_parsing_files` directory
- [x] Try on both `gpt-4` (no SO) and `gpt-4o-mini` (yes SO)
- [x] `simple_d20.py`
- [x]  `nested_pydantic_as_arg_example.py`
- [x] `list_of_pydantic_example.py`
- [x] `all_python_complex.py`


letta/llm_api/openai.py
```diff
@@ -533,7 +536,10 @@ def openai_chat_completions_request(
 
     if "tools" in data:
         for tool in data["tools"]:
-            tool["function"] = convert_to_structured_output(tool["function"])
+            try:
+                tool["function"] = convert_to_structured_output(tool["function"])
+            except ValueError as e:
```
cpacker, line 541: note @mattzh72 in cases where the user made a tool that has a `dict` for example (which can't convert to structurd outputs format), we catch the error here and use non-SO schema instead

mattzh72 (approved):

PR #2134, opened 2024-12-02
feat: Embedded Postgresql

norton120:
This PR embeds Postgres and PgVector binaries into the runtime of the application. The intent is for this embedded db to replace SQLite/Chroma so the project no longer needs to support 2 different backends. 

**How to test**
Start up and run your tests as usual - if no `pg_uri` or postgres db params are specified as settings, an embedded PG will be started with the server. The database data is stored at `<letta_dir>/pgdata`. You can connect directly with an ODBC/JDBC tool (like dbeaver or datagrip or whatever) by using the database uri from the logs, which will look like: 
`postgresql://postgres:@/postgres?host=/Users/yourname/.letta/pgdata`

**Have you tested this PR?**
When I run the `pytest -s tests` I get import errors and warnings on main. I get the same with this branch. 
QA testing, I have been able to create users and see them populate in the db, but I get 422 errors in the GUI (same as on main).

**Related issues or PRs**
Related to https://github.com/letta-ai/letta/pull/2072 which includes pgserver as part of the installable dmg image.


norton120:
Note the SQLite guts are still in there, we'd want to clean that out in a subsequent PR once this is the path forward.

PR #2139, opened 2024-12-02
feat: add POST route for testing tool execution via `tool_id`

cpacker:
```python
@router.post("/run", response_model=FunctionReturn, operation_id="run_tool_from_source")
def run_tool_from_source(
    server: SyncServer = Depends(get_letta_server),
    request: ToolRunFromSource = Body(...),
    user_id: Optional[str] = Header(None, alias="user_id"),  # Extract user_id from header, default to None if not present
):
    """
    Attempt to build a tool from source, then run it on the provided arguments
    """
    actor = server.get_user_or_default(user_id=user_id)

    return server.run_tool_from_source(
        tool_source=request.source_code,
        tool_source_type=request.source_type,
        tool_args=request.args,
        tool_name=request.name,
        user_id=actor.id,
    )
```

Example tests:
```python
EXAMPLE_TOOL_SOURCE_WITH_DISTRACTOR = '''
def util_do_nothing():
    """
    A util function that does nothing.

    Returns:
        str: Dummy output.
    """
    print("I'm a distractor")

def ingest(message: str):
    """
    Ingest a message into the system.

    Args:
        message (str): The message to ingest into the system.

    Returns:
        str: The result of ingesting the message.
    """
    util_do_nothing()
    return f"Ingested message {message}"

'''


def test_tool_run(server, user_id, agent_id):
    """Test that the server can run tools"""

    result = server.run_tool_from_source(
        user_id=user_id,
        tool_source=EXAMPLE_TOOL_SOURCE,
        tool_source_type="python",
        tool_args=json.dumps({"message": "Hello, world!"}),
        # tool_name="ingest",
    )
    print(result)
    assert result.status == "success"
    assert result.function_return == "Ingested message Hello, world!", result.function_return

    result = server.run_tool_from_source(
        user_id=user_id,
        tool_source=EXAMPLE_TOOL_SOURCE,
        tool_source_type="python",
        tool_args=json.dumps({"message": "Well well well"}),
        # tool_name="ingest",
    )
    print(result)
    assert result.status == "success"
    assert result.function_return == "Ingested message Well well well", result.function_return

    result = server.run_tool_from_source(
        user_id=user_id,
        tool_source=EXAMPLE_TOOL_SOURCE,
        tool_source_type="python",
        tool_args=json.dumps({"bad_arg": "oh no"}),
        # tool_name="ingest",
    )
    print(result)
    assert result.status == "error"
    assert "Error" in result.function_return, result.function_return
    assert "missing 1 required positional argument" in result.function_return, result.function_return

...
```

### DIFF
--- a/letta/server/server.py
+++ b/letta/server/server.py
@@ -1,4 +1,5 @@
 # inspecting tools
+import json
 import os
 import traceback
 import warnings
@@ -54,7 +55,7 @@ from letta.schemas.embedding_config import EmbeddingConfig
 # openai schemas
 from letta.schemas.enums import JobStatus
 from letta.schemas.job import Job
-from letta.schemas.letta_message import LettaMessage
+from letta.schemas.letta_message import FunctionReturn, LettaMessage
 from letta.schemas.llm_config import LLMConfig
 from letta.schemas.memory import (
     ArchivalMemorySummary,
@@ -76,9 +77,10 @@ from letta.services.organization_manager import OrganizationManager
 from letta.services.per_agent_lock_manager import PerAgentLockManager
 from letta.services.sandbox_config_manager import SandboxConfigManager
 from letta.services.source_manager import SourceManager
+from letta.services.tool_execution_sandbox import ToolExecutionSandbox
 from letta.services.tool_manager import ToolManager
 from letta.services.user_manager import UserManager
-from letta.utils import create_random_username, json_dumps, json_loads
+from letta.utils import create_random_username, get_utc_time, json_dumps, json_loads
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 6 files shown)


letta/server/server.py
```diff
@@ -1750,3 +1752,107 @@ def get_agent_block_by_label(self, user_id: str, agent_id: str, label: str) -> B
...
+        except Exception as e:
+            # same as agent.py
+            from letta.constants import MAX_ERROR_MESSAGE_CHAR_LIMIT
+
+            error_msg = f"Error executing tool {tool.name}: {e}"
+            if len(error_msg) > MAX_ERROR_MESSAGE_CHAR_LIMIT:
+                error_msg = error_msg[:MAX_ERROR_MESSAGE_CHAR_LIMIT]
+
+            return FunctionReturn(
+                id="null",
```
cpacker, line 1795: @sarahwooders @mattzh72 if we use `FunctionReturn` as the response model, we need to fill these fields (`id`, `function_call_id`) in with dummy fields.

However, we could also create a new response model that doesn't have these fields, but still has the `status` and `function_return` fields?

letta/server/server.py
```diff
@@ -1758,6 +1760,110 @@ def get_agent_block_by_label(self, user_id: str, agent_id: str, label: str) -> B
...
+        user = self.user_manager.get_user_by_id(user_id=user_id)
+        tool = self.tool_manager.get_tool_by_id(tool_id=tool_id, actor=user)
+        if tool.name is None:
+            raise ValueError(f"Tool with id {tool_id} does not have a name")
+
+        # TODO eventually allow using agent state in tools
+        agent_state = None
+
+        try:
+            sandbox_run_result = ToolExecutionSandbox(tool.name, tool_args_dict, user_id).run(agent_state=agent_state)
```
sarahwooders, line 1781: a bit of a bit, but can't this be the same code as `run_tool_from_source` once the `Tool` object is ready?

sarahwooders (approved):

letta/server/server.py
sarahwooders: I think its fine to just set them with dummy fields

letta/server/server.py
cpacker: oh yeah this code I kind of stopped editing in the PR since I wasn't sure if we wanted to delete it?

e.g. if we're not going to use `run_tool` and only use `run_tool_from_source`, shouldn't we just add the later and kill the former (and add it if we need it later?)

PR #2187, opened 2024-12-07
feat: support system message for chatml wrapper

sarahwooders:


cpacker (approved):
We probably want to do this for all wrappers we use, so including the llama3 one

PR #2196, opened 2024-12-08
feat: remove loading from module 

sarahwooders:


cpacker (approved):

letta/agent.py
```diff
@@ -477,10 +477,10 @@ def link_tools(self, tools: List[Tool]):
         for tool in tools:
             try:
                 # WARNING: name may not be consistent?
-                if tool.module:  # execute the whole module
-                    exec(tool.module, env)
-                else:
-                    exec(tool.source_code, env)
+                # if tool.module:  # execute the whole module
```
cpacker, line 480: @sarahwooders we probably want to do

```python
if tool.module:
    warnings.warn(f"tool.module is deprecated and should not exist, but was found on agent, state =\n{self.agent_state}")
```

PR #2222, opened 2024-12-11
feat: route all sandbox errors to stderr

carenthomas:
New functionality to support surfacing all information about a sandbox run in the UI. Currently, if code execution in a sandbox errors, we lose additional information like the stdout up to that point in the code. Catching the exceptions and bubbling them up will allow us to improve the UX by surfacing both stdout and error message to the user. Note that top-level try/catches are left in place for exceptions thrown outside of the sandbox in the Letta execution path + anything we missed.

**How to test**
poetry run pytest -s -vv tests/integration_test_tool_execution_sandbox.py

**Have you tested this PR?**
yes

**Related issues or PRs**

**Additional context**


sarahwooders (approved):

carenthomas:
@mattzh72 based on this line, it looks like there's an assumption that the function response is either the output or the error if an exception is thrown: https://github.com/letta-ai/letta/blob/c9c2cca4f499c0128ef5de5a89cf85c065dff373/tests/test_server.py#L569

My latest commit preserves this assumption and passes the test but introduces some duplication so some tradeoffs there.. do you know if the frontend requires func_return to contain the error info?

carenthomas:
Successfully added to UI (will explore adding tabs + resizing)

Screenshot 2024-12-11 at 12 08 15 PM

sarahwooders:
@carenthomas I think if you merge with main your tests will pass

letta/server/server.py
```diff
@@ -1955,24 +1955,35 @@ def run_tool_from_source(
...
                 raise ValueError(f"Tool with id {tool.id} returned execution with None")
             function_response = str(sandbox_run_result.func_return)
+            stdout = [s for s in sandbox_run_result.stdout if s.strip()]
+            stderr = [s for s in sandbox_run_result.stderr if s.strip()]
+
+            # expected error
+            if stderr:
+                error_msg = self.get_error_msg_for_func_return(tool.name, stderr[-1])
+                return FunctionReturn(
+                    id="null",
```
mattzh72, line 1965: nit: code style around setting id/function_call_id to predefined string - ideally, we would have maybe a child type of FunctionReturn without id/function_call_id

letta/services/tool_execution_sandbox.py
```diff
@@ -217,39 +193,49 @@ def run_local_dir_sandbox_venv(self, sbx_config: SandboxConfig, env: Dict[str, s
...
-            stdout=[stdout_output],
-            stderr=[stderr_output],
+            stdout=stdout_output,
+            stderr=stderr_output,
             sandbox_config_fingerprint=sbx_config.fingerprint(),
         )
 
     def parse_out_function_results_markers(self, text: str):
+        if self.LOCAL_SANDBOX_RESULT_START_MARKER not in text:
+            return '', text
```
mattzh72, line 238: nit: throw an error here v.s. empty string?

letta/services/tool_execution_sandbox.py
```diff
@@ -356,6 +343,8 @@ def list_running_e2b_sandboxes(self):
     # general utility functions
 
     def parse_best_effort(self, text: str) -> Any:
+        if not text:
```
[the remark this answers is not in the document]
carenthomas: now that we're not raising an exception in this path when the tool code fails, if there's an error in the tool code, we won't write the return values into stdout. So even if `wrap_print_with_markers` is true, the sandbox won't reach the lines that do the printing, which is the case where text will be an empty string. Might be helpful for us to trace out all the valid paths and reorganize the code around those flows.

letta/services/tool_execution_sandbox.py
carenthomas: This is still that case below that is valid where there is an exception thrown in the sandbox so no rv gets printed. I did it this way so that we can still extract the stdout and return it via this function (otherwise, we drop whatever got printed before the error happened and can't surface it to the user)

carenthomas: This is part of the ugliness I mentioned I'm not thrilled about. Lmk if you see any ways to improve this otherwise lets brainstorm together as a follow!

letta/server/server.py
mattzh72: Okay to leave todo :)

PR #2228, opened 2024-12-11
feat: archival memory file upload

mlong93:
When a user uploads a file to an existing source, archival memory is automatically updated to contain passages from that file.

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**


letta/server/server.py
```diff
@@ -1590,6 +1590,15 @@ def load_file_to_source(self, source_id: str, file_path: str, job_id: str, actor
...
         num_passages, num_documents = self.load_data(user_id=source.created_by_id, source_name=source.name, connector=connector)
 
+        # update all agents who have this source attached
+        agent_ids = self.ms.list_attached_agents(source_id=source_id)
+        for agent_id in agent_ids:
+            agent = self.load_agent(agent_id=agent_id)
+            curr_passage_size = self.passage_manager.size(actor=actor, agent_id=agent_id, source_id=source_id)
+            agent.attach_source(user=actor, source_id=source_id, source_manager=self.source_manager, ms=self.ms)
+            new_passage_size = self.passage_manager.size(actor=actor, agent_id=agent_id, source_id=source_id)
+            assert new_passage_size > curr_passage_size
```
sarahwooders, line 1600: I think this assert is causing issues on the second file upload - I think it might be because we aren't waiting for the load data to complete before re-attaching? so the passage aren't ready yet

tests/test_server.py
```diff
@@ -691,3 +709,136 @@ def count_system_messages_in_recall() -> Tuple[int, List[LettaMessage]]:
...
+    server.load_file_to_source(
+        source_id=source.id,
+        file_path=str(test_file2),
+        job_id=job2.id,
+        actor=user,
+    )
+
+    # Verify second job completed successfully
+    job2 = server.job_manager.get_job_by_id(job_id=job2.id, actor=user)
+    assert job2.status == "completed"
```
sarahwooders, line 791: I think the file you are testing with is very small, so the job is immediately completed - might be better to use a bigger file (I use the `CONTRIBUTING.md` file sometimes, or we could add a bigger test file) to more reliably catch errors

sarahwooders (approved):
This looks great, but I think there are some issues with loading multiple files in sequence that your tests aren't catching.

letta/server/server.py
mlong93: It turns out it's because when we call `attach_source`, there was a `page_size` limit of 100 so you could never add more than 100 passages to an agent from a source

PR #2232, opened 2024-12-12
feat:  Rewrite agents

mattzh72:
## Description

Rewrite the agents data model. There are many changes, but the important ones to call out:

### Pydantic model changes
We remove `PersistedAgentState`, and now agent_state has all the tools, blocks, and other tied objects returned. Essentially, the agent state pydantic object is a "snapshot" of the db state

### API changes
We change the interface of the `client` and REST API endpoints. 

- To tie in a tool to an agent, we require the `tool_ids` now, instead of `tool_names`. 
- We have `memory` which is a collection of `CreateBlock` requests, and `block_ids`, which reference existing blocks to tie into an agent


alembic/versions/d05669b60ebe_migrate_agents_to_orm.py
```diff
@@ -0,0 +1,164 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "d05669b60ebe"
+down_revision: Union[str, None] = "c5d964280dff"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
```
sarahwooders, line 25: why did we replace the table? just wondering

letta/agent.py
```diff
@@ -1612,18 +1618,29 @@ def count_tokens(self) -> int:
...
+        tags=agent_state.tags,
+        system=agent_state.system,
+        tool_rules=agent_state.tool_rules,
+        llm_config=agent_state.llm_config,
+        embedding_config=agent_state.embedding_config,
+        message_ids=agent_state.message_ids,
+        description=agent_state.description,
+        metadata_=agent_state.metadata_,
+    )
+    agent_manager.update_agent(agent_id=agent_state.id, agent_update=update_agent)
```
sarahwooders, line 1643: will this also save the blocks? if so, maybe make a note of it?

letta/cli/cli.py
```diff
@@ -140,7 +140,7 @@ def run(
     # read user id from config
     ms = MetadataStore(config)
     client = create_client()
-    server = client.server
+    client.server
```
sarahwooders, line 143: delete?

letta/schemas/agents_tags.py
```diff
@@ -1,33 +0,0 @@
-from datetime import datetime
```
sarahwooders, line 1: what was this replaced by?

letta/schemas/memory.py
```diff
@@ -213,6 +213,7 @@ def __init__(self, persona: str, human: str, limit: int = CORE_MEMORY_BLOCK_CHAR
             human (str): The starter value for the human block.
             limit (int): The character limit for each block.
         """
+        # TODO: Should these be CreateBlocks?
```
sarahwooders, line 216: I think we need to get rid of this concept... this was intended to be a nice way for the client to initialized a specific memory "type", but over time it seems like the only customization we need is at the block level, and memory is just a list of block objects and memory tools. We should maybe get rid of these `BasicBlockMemory` extensions, and definitely get rid of the free floating `Block` since that's dangerous under the new model.

letta/schemas/agents_tags.py
mattzh72: we no longer need the Pydantic object

mattzh72: the table is purely relational now and needs no manager, so never need to convert to pydantic obj

letta/agent.py
mattzh72: it requires that the passed in block ids already exist

alembic/versions/d05669b60ebe_migrate_agents_to_orm.py
mattzh72: OH just to keep naming convention with the other relationship tables, and I figured no one is really using sources (yet) as it's severely broken. original table def was in `metadata.py` called `agent_source_mapping`

letta/cli/cli.py
mattzh72: done

letta/schemas/memory.py
mattzh72: will do in a follow up PR

sarahwooders (approved):

PR #2240, opened 2024-12-12
initial checkin for vision support with tests

montiblanc97:
Add vision support when using OpenAI models
- both image URL and base64-encoded images
- tests

Remaining work before merge:
- Renaming Message.text -> Message.content which is used in a lot of places
- Updating the ORM Message object which is typed str and requires migration.
  - Current behavior is just ignoring the images and storing the text
- CLI support (ollama-like). Parse any local file paths in CLI text input and treat them as image uploads
- Solution to saving base64 images. They probably should not go in the Postgres DB
  - We can store uploaded images to cloud blob storage (S3, GCP) then replace images in the messages with S3 urls
  - For local servers: min.io, open-source S3 alternative with same API
    - Deployed via docker-compose. Can wrap under letta_server like PG or separate service
    - API is same, so the only differences between cloud/local server will be some config values if using AWS
  - Sarah also suggested an alternative for local dev which is saving local file paths. Much simpler

Extensions:
- Model switching to non-multimodal models, when the history contains images. Ideas:
  - Captioning model fallback
  - Insertion of "unavailable image" or something along those lines

**How to test**

Unit tests included

**Have you tested this PR?**

Unit tests pass

**Related issues or PRs**

**Additional context**

### DIFF
--- a/letta/client/client.py
+++ b/letta/client/client.py
@@ -152,10 +152,11 @@ class AbstractClient(object):
         stream: Optional[bool] = False,
         stream_steps: bool = False,
         stream_tokens: bool = False,
+        image: Optional[str] = None,
     ) -> LettaResponse:
         raise NotImplementedError
 
-    def user_message(self, agent_id: str, message: str) -> LettaResponse:
+    def user_message(self, agent_id: str, message: str, image: str|None) -> LettaResponse:
         raise NotImplementedError
 
     def create_human(self, name: str, text: str) -> Human:
@@ -832,7 +833,7 @@ class RESTClient(AbstractClient):
 
     # agent interactions
 
-    def user_message(self, agent_id: str, message: str) -> LettaResponse:
+    def user_message(self, agent_id: str, message: str, image: str|None) -> LettaResponse:
         """
         Send a message to an agent as a user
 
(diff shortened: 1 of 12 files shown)


letta/client/client.py
```diff
@@ -152,10 +152,11 @@ def send_message(
         stream: Optional[bool] = False,
         stream_steps: bool = False,
         stream_tokens: bool = False,
+        image: Optional[str] = None,
```
cpacker, line 155: I think the ideal way to do this is we actually make `message` (currently `str`) be type `MultiMediaContent`, which itself is `Union[str, List[MultiMediaContentPart]]`

Then we will be able to support multiple images and future modalities like audio.

letta/schemas/message.py
```diff
@@ -421,8 +423,22 @@ def to_openai_dict(
 
         elif self.role == "user":
             assert all([v is not None for v in [self.text, self.role]]), vars(self)
+
+            if self.image is not None:
```
cpacker, line 427: So inside of `Message`, `.text` becomes `.content`, which basically is a 1:1 mapping to OpenAI's format: 
image

https://platform.openai.com/docs/api-reference/chat/create

montiblanc97: agree with this and made the initial typing changes/enough to pass tests. However left off before two major changes:
1. Renaming Message.text -> Message.content which is used in a lot of places
2. Updating the ORM Message object which is typed str and requires migration. Tricky because images probably shouldn't be stored raw in the DB

letta/client/client.py
montiblanc97: sounds good, updated. see message in your other comment

PR #2248, opened 2024-12-13
fix: refactor sandbox run logic to add status field

carenthomas:
Refactor so that we no longer derive the error case based on whether stderr is populated (script could manually write to stderr even if the run is successful). This refactor also moves the logic for populating the func_return value with a friendly error message at the source of the error rather than waiting until the end, and adds a helper method so all functions do it the same way.

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -63,6 +63,7 @@ from letta.system import (
 )
 from letta.utils import (
     count_tokens,
+    get_friendly_error_msg,
     get_local_time,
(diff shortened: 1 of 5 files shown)


sarahwooders (changes requested):
This looks good but could we fix the merge conflicts and re-run the cypress tests?

sarahwooders (approved):

PR #2250, opened 2024-12-14
fix: Changing e2b template id as an env var reflects immediately

mattzh72:
Changing e2b template id as an env var reflects immediately. Before, it needed to be re-written to the database, now it derives directly from the environment variable. Existing test cases for grabbing the default e2b covers this.


carenthomas (approved):
LGTM

PR #2262, opened 2024-12-17
feat: Adding init tool rule for Anthropic endpoint

mlong93:
https://linear.app/letta/issue/LET-508/inittool-tool-rule

* Fixed first message flag
* Added support for Claude-3-sonnet configs
* New test case for init tool rule

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -385,7 +385,8 @@ class Agent(BaseAgent):
 
     def check_tool_rules(self):
         if self.model not in STRUCTURED_OUTPUT_MODELS:
-            assert len(self.tool_rules_solver.init_tool_rules) <= 1, "Multiple initial tools not supported for non-structured models"
+            if len(self.tool_rules_solver.init_tool_rules) > 1:
+                raise ValueError("Multiple initial tools are not supported for non-structured models. Please use only one initial tool rule.")
             self.supports_structured_output = False
         else:
             self.supports_structured_output = True
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -18,6 +18,7 @@ from letta.constants import (
     MESSAGE_SUMMARY_WARNING_FRAC,
     O1_BASE_TOOLS,
     REQ_HEARTBEAT_MESSAGE,
+    STRUCTURED_OUTPUT_MODELS
 )
 from letta.errors import LLMError
 from letta.helpers import ToolRulesSolver
@@ -276,6 +277,7 @@ class Agent(BaseAgent):
 
         # gpt-4, gpt-3.5-turbo, ...
         self.model = self.agent_state.llm_config.model
+        self.check_tool_rules()
 
         # state managers
         self.block_manager = BlockManager()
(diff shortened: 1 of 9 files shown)


letta/agent.py
```diff
@@ -596,6 +596,12 @@ def _get_ai_reply(
             self.functions if not allowed_tool_names else [func for func in self.functions if func["name"] in allowed_tool_names]
         )
 
+        # For the first message, force the initial tool if one is specified
+        force_tool_call = None
+        if first_message and self.tool_rules_solver.init_tool_rules:
+            assert len(self.tool_rules_solver.init_tool_rules) == 1, "Multiple initial tools not supported"
```
sarahwooders, line 602: maybe move this into the create agent function? and we should allow this (not throw an error) if the LLM config specifies OpenAI + one of these structured output models https://platform.openai.com/docs/guides/structured-outputs#supported-models

tests/integration_test_agent_tool_graph.py
```diff
@@ -127,3 +127,47 @@ def test_single_path_agent_tool_call_graph(mock_e2b_api_key_none):
...
+    tool_rules = [
+        InitToolRule(tool_name="first_secret_word"),
+        ChildToolRule(tool_name="first_secret_word", children=["second_secret_word"]),
+    ]
+    tools = [t1, t2]
+
+    # Make agent state
+    anthropic_config_file = "tests/configs/llm_model_configs/claude-3-sonnet-20240229.json"
+    agent_state = setup_agent(client, anthropic_config_file, agent_uuid=agent_uuid, tool_ids=[t.id for t in tools], tool_rules=tool_rules)
+    response = client.user_message(agent_id=agent_state.id, message="What is the second secret word?")
```
sarahwooders, line 157: you might want to make this test more robust by calling `client.user_message(...)` multiple times, since every single time `first_secret_word` should get called first. 

Also does `setup_agent` include all the default tools (e.g. `send_message`)? If not, we might want to include them to make the test more robust as well.

mlong93: `setup_agent` includes all the `BASE_TOOLS`. I added repeats to the test

letta/agent.py
mlong93: Moved the check to a function at initialization `self.check_tool_rules()`

letta/constants.py
```diff
@@ -48,6 +48,9 @@
 DEFAULT_MESSAGE_TOOL = "send_message"
 DEFAULT_MESSAGE_TOOL_KWARG = "message"
 
+# Structured output models
+STRUCTURED_OUTPUT_MODELS = {"gpt-4o-mini-2024-07-18", "gpt-4o-2024-08-06"}
```
sarahwooders, line 52: would this fail to match if the model name is just `gpt-4o-mini`?

sarahwooders (changes requested):
Could we add a check in the agent creation/update that there is only 1 init tool rule unless the model is in `STRUCTURED_OUTPUT_MODELS` and throws an informative error (e.g. `Only one instance of InitToolRule allowed .... `) otherwise? 

Otherwise looks great!

letta/constants.py
mlong93: changed it to be less strict (4o or 4o-mini)

letta/agent.py
```diff
@@ -896,7 +912,10 @@ def step(
         total_usage = UsageStatistics()
         step_count = 0
         while True:
-            kwargs["first_message"] = False
```
sarahwooders, line 899: I think this is referring to the *very* first message, so we shouldn't change this

mlong93: reverted and replaced indicator with `step=0`

sarahwooders (approved):

PR #2263, opened 2024-12-17
fix: Scale up database

mattzh72:
Scale up database to allow more concurrent connections - also, surface more informative HTTP errors.


letta/server/server.py
```diff
@@ -191,7 +190,14 @@ def db_error_handler():
     config.archival_storage_uri = settings.letta_pg_uri_no_default
 
     # create engine
-    engine = create_engine(settings.letta_pg_uri)
+    engine = create_engine(
```
sarahwooders, line 193: We might want to make these environment variables (maybe put them in settings.py) so they can be configured

mattzh72: ack

sarahwooders (approved):

PR #2274, opened 2024-12-18
chore: Clean up upserting base tools 

mattzh72:
Clean up functionality around upserting base tools with the latest code.

### DIFF
--- a/scripts/migrate_tools.py
+++ b/scripts/migrate_tools.py
@@ -14,7 +14,7 @@ for org in orgs:
     if org.name != "default":
         fake_user = User(id="user-00000000-0000-4000-8000-000000000000", name="fake", organization_id=org.id)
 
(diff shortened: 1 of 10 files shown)


scripts/migrate_tools.py
```diff
@@ -10,33 +10,8 @@ def deprecated_tool():
 
 
 orgs = OrganizationManager().list_organizations(cursor=None, limit=5000)
-for org in orgs:
+for org in tqdm(orgs):
```
mlong93, line 13: Why do we want tqdm?

mattzh72: (y)

mattzh72: discussed offline, just for visual purposes while running this script on pod

mlong93 (approved):

PR #2279, opened 2024-12-18
feat: Add ConditionalToolRules

mlong93:
# Description
* update ToolRuleSolver with `ConditionalToolRule` type
* updated tests for conditional rules


letta/schemas/tool_rule.py
```diff
@@ -21,6 +21,17 @@ class ChildToolRule(BaseToolRule):
     children: List[str] = Field(..., description="The children tools that can be invoked.")
 
 
+class ConditionalToolRule(BaseToolRule):
+    """
+    A ToolRule that conditionally maps to different child tools based on the output.
+    """
+    type: ToolRuleType = ToolRuleType.conditional
+    default_child: str = Field(..., description="The default child tool to be called")
```
sarahwooders, line 29: I think it's ok for this to be `None` where if it's `None`, then we don't constrain the next tool (it can be any tool)

letta/schemas/tool_rule.py
```diff
@@ -21,6 +21,17 @@ class ChildToolRule(BaseToolRule):
...
+    child_output_mapping: Dict[Union[bool, str, int], str] = Field(..., description="The output case to check for mapping")
```
sarahwooders, line 30: Could we do `Dict[Any, str]`? This is mapping the output of the tool to the child name, right?

letta/schemas/tool_rule.py
```diff
@@ -21,6 +21,17 @@ class ChildToolRule(BaseToolRule):
...
+    children: List[str] = Field(..., description="The child tool to call when output matches the case")
```
sarahwooders, line 31: I'm confused what this field is - isn't the child tool specified in `child_output_mapping`?

letta/schemas/tool_rule.py
```diff
@@ -21,6 +21,17 @@ class ChildToolRule(BaseToolRule):
...
+    throw_error: bool = Field(default=False, description="Whether to throw an error when output doesn't match any case")
```
sarahwooders, line 32: maybe rename this to `require_output_mapping` to make more clear what this means

mlong93: yeah this is repetitive. removed `children`

letta/schemas/tool_rule.py
mlong93: yeah this is repetitive. removed `children`

sarahwooders (approved):

PR #2281, opened 2024-12-18
fix: propagate error on tool failure

carenthomas:
reinstate prior functionality. following up with a more robust implementation that doesn't infer status based on the string (but this is fine for now since I am setting the string the same way at all callsites)

**How to test**
poetry run pytest -s -vv tests/test_client.py::test_function_always_error

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -10,6 +10,7 @@ from typing import List, Literal, Optional, Tuple, Union
 from letta.constants import (
     BASE_TOOLS,
     CLI_WARNING_PREFIX,
+    ERROR_MESSAGE_PREFIX,
     FIRST_MESSAGE_ATTEMPTS,
     FUNC_FAILED_HEARTBEAT_MESSAGE,
     IN_CONTEXT_MEMORY_KEYWORD,
@@ -863,7 +864,7 @@ class Agent(BaseAgent):
                 return messages, False, True  # force a heartbeat to allow agent to handle error
 
             # Step 4: check if function response is an error
-            if function_response_string.startswith("Error"):
+            if function_response_string.startswith(ERROR_MESSAGE_PREFIX):
                 function_response = package_function_response(False, function_response_string)
                 # TODO: truncate error message somehow
                 messages.append(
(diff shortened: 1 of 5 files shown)


letta/agent.py
```diff
@@ -844,8 +844,29 @@ def _handle_ai_response(
                 self.interface.function_message(f"Error: {error_msg}", msg_obj=messages[-1])
                 return messages, False, True  # force a heartbeat to allow agent to handle error
 
+            # Step 4: check if function response is an error
+            if function_response_string.startswith("Error"):
```
sarahwooders, line 848: where is this `"Error"` substring coming from? Could we maybe make this a constant (e.g. `SANDBOX_ERROR_MSG_PREFIX` so that we don't accidentally break this later on if we are modifying the function response elsewhere?

letta/server/rest_api/interface.py
```diff
@@ -238,7 +238,7 @@ def function_message(self, msg: str, msg_obj: Optional[Message] = None, include_
             new_message = {"function_return": msg, "status": "success"}
 
         elif msg.startswith("Error: "):
-            msg = msg.replace("Error: ", "")
+            msg = msg.replace("Error: ", "", count=1)
```
sarahwooders, line 241: what is `count`?

letta/agent.py
carenthomas: sure I'll pull it out into a constant and use it where I'm generating the error message

> following up with a more robust implementation that doesn't infer status based on the string (but this is fine for now since I am setting the string the same way at all callsites)

mentioned in the summary but I'll make this more robust in my follow up (tomorrow)!

letta/server/rest_api/interface.py
carenthomas: it's the Max amount of replaces that will be done. Without it there is an unintended error in prod where the error string passed in:

__**Error: **__Error executing function always_error: ZeroDivision__**Error: **__division by zero

gets stripped to 

Error executing function always_error: ZeroDivisiondivision by zero

PR #2290, opened 2024-12-19
feat: add error codes to composio errors

4shub:
Adds some error codes to composio


letta/server/rest_api/routers/v1/tools.py
```diff
@@ -243,6 +243,7 @@ def add_composio_tool(
         raise HTTPException(
             status_code=400,  # Bad Request
             detail={
+                "error_code": "EnumStringNotFound",
```
carenthomas, line 246: key name `code` instead of `error_code` to keep it consistent with the other callsites on front end and backend

carenthomas (changes requested):

PR #2306, opened 2024-12-21
fix: record the external memory summary inside of the context viewer

cpacker:


cpacker:
@mattzh72 we should (in a later PR) update the get context viewer route to not call `load_agent`

PR #2338, opened 2025-01-08
fix: update authorization header to use X-BARE-PASSWORD format

theocnrds:
Fixes a bug in Python SDK. Setting `token=<password>` when instantiating a RESTClient was not sufficient to authorize since `CheckPasswordMiddleware` was checking for `X-BARE-PASSWORD` instead of 'Authorization` header.

RESTClient is now setting `X-BARE-PASSWORD="password <password>` as a header.

**How to test**
```python
from letta.client.client import RESTClient

client = RESTClient(base_url="http://0.0.0.0:8283", token="<password>")
print(client.personas())
```

### DIFF
--- a/letta/client/client.py
+++ b/letta/client/client.py
@@ -428,7 +428,7 @@ class RESTClient(AbstractClient):
         super().__init__(debug=debug)
         self.base_url = base_url
         self.api_prefix = api_prefix
-        self.headers = {"accept": "application/json", "authorization": f"Bearer {token}"}
+        self.headers = {"accept": "application/json", "X-BARE-PASSWORD": f"password {token}"}
         if headers:
             self.headers.update(headers)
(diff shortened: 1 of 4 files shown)


letta/client/client.py
```diff
@@ -428,7 +428,7 @@ def __init__(
         super().__init__(debug=debug)
         self.base_url = base_url
         self.api_prefix = api_prefix
-        self.headers = {"accept": "application/json", "authorization": f"Bearer {token}"}
+        self.headers = {"accept": "application/json", "X-BARE-PASSWORD": f"password {token}"}
```
sarahwooders, line 431: So we actually need the `token` field for cloud users - would you be able to instead make `password` an additional field, where *either* password or token will be used?

sarahwooders (approved):

PR #2461, opened 2025-02-28
refactor(azure-embedding): Removed unnecessary and commented Azure embedding code

kk-src:
Removed the embedding calls in `azure_openai.py` which is no longer used. The code path exercised is the `get_text_embedding` in `class AzureOpenAIEmbedding`. Tested this by running directory upload with Azure embedding model.

Also removed commented out code from Azure code path as a general clean up.

**How to test**
- Configure Azure environment variables.
- Run upload directory command to run generating embedding.

**Have you tested this PR?**
Yes, by running upload directory and choosing Azure embedding model.


cpacker (approved):
Thank you!

PR #2548, opened 2025-04-09
Renaming AWS_ACCESS_KEY to AWS_ACCESS_KEY_ID as per aws naming convention

### LINKED ISSUE #2538: Bug in AWS support for Letta due to incorrect variable name for access key.
**Describe the bug**

Hello all. I noticed a bug in AWS support. Line 17 of aws_bedrock.py is looking for the wrong access key variable. You have AWS_ACCESS_KEY. The correct variable names are AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and optionally AWS_SESSION_TOKEN for short-term creds. If the AWS_ACCESS_KEY_ID starts with AS then it's short-term and needs the session token.

It's easy enough to fix the variable name so it will work with Letta, but you probably meant to have the default names so people could paste in creds without having to change variable names.

Also this is not going to work with instance profiles, task roles, or execution roles easily. You'd be better off just doing a try and except to see if you can set up a session with boto3, and let the SDK find the creds via the normal path (environment variables as well as execution roles or creds files.

In general only using the access key id and secret access key means you're using long-lived creds, which is not good for security best practices. Better to support execution roles on AWS so they can manage short-term creds for you.

Do you take pull requests for fixes?

Thanks for this great product.

**Please describe your setup**

I don't think you need the extra details given the nature of this issue, but this is in the main branch of letta that I just cloned.

punitsu:
**Please describe the purpose of this pull request.*
- Resolves #2538 

**How to test**
- I ran unit tests 

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**


kianjones9:
Sorry just to close the loop here we have since added more support for public cloud providers and this should be up-to-date. Officially closing this issue and related. Thank you for your contribution 🙏 .

PR #2568, opened 2025-04-16
fix: only trim in-context messages to cutoff

jfkirk:

There is a bug in `summarize_messages_inplace()` that causes the agent to throw out all of its messages every time it summarizes, instead of only those up to `[:cutoff]`. This was caused by a call to `trim_all_in_context_messages_except_system()` which blanks the message stream instead of `trim_older_in_context_messages(num=cutoff)` which would only remove the earliest messages in the stream.

This PR swaps the `trim_*()` function in summarization.

**Additional context**

The symptom of this is that the context window messages are totally emptied every time summarization runs.

image

### DIFF
--- a/letta/agent.py
+++ b/letta/agent.py
@@ -1103,7 +1103,7 @@ class Agent(BaseAgent):
         logger.info(f"Packaged into message: {summary_message}")
 
         prior_len = len(in_context_messages_openai)
-        self.agent_state = self.agent_manager.trim_all_in_context_messages_except_system(agent_id=self.agent_state.id, actor=self.user)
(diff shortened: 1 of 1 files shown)


sarahwooders:
@jfkirk could you please rebase on main? I'm not sure why but the tests are failing.

jfkirk:
@sarahwooders just rebased this branch off of c6e55e04d149f3ab3db41da1b06a2b80c576e76d (the 0.7.7 release)

sarahwooders (approved):
Thanks for the contribution!

PR #2689, opened 2025-06-19
chore(ci): Improve Docker Integration Tests

kianjones9:

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**

### DIFF
--- a/.github/workflows/docker-integration-tests.yaml
+++ b/.github/workflows/docker-integration-tests.yaml
@@ -56,8 +56,10 @@ jobs:
         PYTHONPATH: ${{ github.workspace }}:${{ env.PYTHONPATH }}
       run: |
         pipx install poetry==2.1.3
-        poetry install -E dev -E postgres
+        poetry install --all-extras
+        poetry run pytest -s tests/test_plugins.py
         poetry run pytest -s tests/test_*.py
+
     - name: Print docker logs if tests fail
       if: failure()
       run: |
--- a/.github/workflows/docker-integration-tests.yaml
+++ b/.github/workflows/docker-integration-tests.yaml
@@ -57,9 +57,7 @@ jobs:
       run: |
         pipx install poetry==2.1.3
         poetry install -E dev -E postgres
-        poetry run pytest -s tests/test_client.py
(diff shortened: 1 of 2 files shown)


letta/data_sources/redis_client.py
```diff
@@ -246,7 +246,13 @@ def _get_group_exclusion_key(group: str) -> str:
 class NoopAsyncRedisClient(AsyncRedisClient):
     # noinspection PyMissingConstructor
     def __init__(self):
-        pass
```
cliandy, line 249: these all should not be set since this is a noop instance

.github/workflows/docker-integration-tests.yaml
```diff
@@ -56,9 +57,9 @@ jobs:
         PYTHONPATH: ${{ github.workspace }}:${{ env.PYTHONPATH }}
       run: |
         pipx install poetry==2.1.3
-        poetry install -E dev -E postgres
-        poetry run pytest -s tests/test_client.py
-#        poetry run pytest -s tests/test_client_legacy.py
+        poetry install --all-extras
+        poetry run pytest -s tests/test_plugins.py
+        poetry run pytest -s tests/test_*.py
```
cliandy, line 62: is this just to increase the testing we do for oss, and doesnt the first test line get covered by the glob?

kianjones9: 1) yes we need to increase testing as sometimes we release an OSS docker image which may get used in ways which weren't tested in closed-source, and 2) yes I will remove it, I had only added it to test it first in case other tests failed first.

letta/data_sources/redis_client.py
kianjones9: well I think because we are defining this constructor, we implicitly override the parent class constructor thus we never set `_client` for example, and then when we run `test_plugins`, the test fails because 
```
FAILED tests/test_plugins.py::test_redis_flag - AttributeError: 'NoopAsyncRedisClient' object has no attribute '_client'. Did you mean: 'get_client'?
=================== 1 failed, 3 passed, 16 warnings in 0.69s ===================
```

PR #2717, opened 2025-07-10
Add vertex region env var

andrewrfitz:
To allow more customization of the vertex client -- for example, gemini 2.5 flash lite is only available in the global region

**How to test**

**Have you tested this PR?**

**Related issues or PRs**

**Additional context**

### DIFF
--- a/letta/llm_api/google_vertex_client.py
+++ b/letta/llm_api/google_vertex_client.py
@@ -34,10 +34,12 @@ class GoogleVertexClient(LLMClientBase):
 
     def _get_client(self):
         timeout_ms = int(settings.llm_request_timeout_seconds * 1000)
+        # Use vertex_ai_region if set, otherwise fall back to google_cloud_location
+        location = model_settings.vertex_ai_region or model_settings.google_cloud_location
         return genai.Client(
             vertexai=True,
             project=model_settings.google_cloud_project,
-            location=model_settings.google_cloud_location,
+            location=location,
             http_options=HttpOptions(api_version="v1", timeout=timeout_ms),
         )
 
--- a/letta/settings.py
+++ b/letta/settings.py
@@ -135,6 +135,7 @@ class ModelSettings(BaseSettings):
     # google vertex
     google_cloud_project: Optional[str] = None
     google_cloud_location: Optional[str] = None
+    vertex_ai_region: Optional[str] = None
 
     # together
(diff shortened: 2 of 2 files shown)


cliandy:
setting that config can make things a bit confusing as other paths in the code use the location setting

@carenthomas pointed me towards this function in `server.py` where you can set up custom configs on your version. Would this work?

```python
    def get_local_llm_configs(self):
        llm_models = []
        try:
            llm_configs_dir = os.path.expanduser("~/.letta/llm_configs")
            if os.path.exists(llm_configs_dir):
                for filename in os.listdir(llm_configs_dir):
                    if filename.endswith(".json"):
                        filepath = os.path.join(llm_configs_dir, filename)
                        try:
                            with open(filepath, "r") as f:
                                config_data = json.load(f)
                                llm_config = LLMConfig(**config_data)
                                llm_models.append(llm_config)
                        except (json.JSONDecodeError, ValueError) as e:
                            warnings.warn(f"Error parsing LLM config file {filename}: {e}")
        except Exception as e:
            warnings.warn(f"Error reading LLM configs directory: {e}")
        return llm_models
```

andrewrfitz:
> setting that config can make things a bit confusing as other paths in the code use the location setting
> 
> @carenthomas pointed me towards this function in `server.py` where you can set up custom configs on your version. Would this work?
> 
> ```python
>     def get_local_llm_configs(self):
>         llm_models = []
>         try:
>             llm_configs_dir = os.path.expanduser("~/.letta/llm_configs")
>             if os.path.exists(llm_configs_dir):
>                 for filename in os.listdir(llm_configs_dir):
>                     if filename.endswith(".json"):
>                         filepath = os.path.join(llm_configs_dir, filename)
>                         try:
>                             with open(filepath, "r") as f:
>                                 config_data = json.load(f)
>                                 llm_config = LLMConfig(**config_data)
>                                 llm_models.append(llm_config)
>                         except (json.JSONDecodeError, ValueError) as e:
>                             warnings.warn(f"Error parsing LLM config file {filename}: {e}")
>         except Exception as e:
>             warnings.warn(f"Error reading LLM configs directory: {e}")
>         return llm_models
> ```

So in this case I would drop a file on the disk of the machine? Also, this change wouldn't be affecting the LLMConfig of each agent -- it would just be a global setting for the vertex client

PR #2744, opened 2025-08-02
fix: all model types returned from ollama provider

antondevson:
Makes sure only completion models are returned from `list_llm_models_async` and only embeddings from `list_embedding_models_async`, instead of both from each. Also some minor refactoring, moving constants to [constants.py](../blob/main/letta/constants.py).

**How to test**
Run `/v1/models` and `/v1/models/embedding` and see that the models returned are completion models and embeddings respectively, currently in main both are returned together for each endpoint.

**Have you tested this PR?**
Yes, as above.

**Related issues or PRs**
Fixes the second issue discussed in https://github.com/letta-ai/letta/issues/2739.

One more optimization here would be to cache the calls to `/api/show` since they all have to be fetched for both functions, but this probably adds more complexity than it's worth at this stage.

### DIFF
--- a/letta/constants.py
+++ b/letta/constants.py
@@ -10,6 +10,7 @@ DEFAULT_TIMEZONE = "UTC"
 
 ADMIN_PREFIX = "/v1/admin"
 API_PREFIX = "/v1"
+OLLAMA_API_PREFIX = "/v1"
 OPENAI_API_PREFIX = "/openai"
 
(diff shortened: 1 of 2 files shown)


antondevson:
It seems [providers.py](../blob/main/letta/schemas/providers.py) has a lot of code duplication with the contents of `letta/schemas/providers/`, this file should probably be deleted. I tried building with it removed and it works but I haven't tested more than that. Should likely be a separate task.

sarahwooders (approved):

PR #2823, opened 2025-09-10
feat: hold reference to asyncio tasks in memory

carenthomas:
Based on the asyncio documentation, we should be holdering references to asyncio tasks in memory to prevent the garbage collector from cleaning them up: https://docs.python.org/3/library/asyncio-task.html#creating-tasks

This probably doesn't scale well in the long term, but eventually we should move to a better system for background task management rather than just depending on the event loop.

### DIFF
--- a/letta/utils.py
+++ b/letta/utils.py
@@ -1093,14 +1093,38 @@ def make_key(*args, **kwargs):
     return str((args, tuple(sorted(kwargs.items()))))
 
 
-def safe_create_task(coro, logger: Logger, label: str = "background task"):
+# Global set to keep strong references to background tasks
+_background_tasks: set = set()
+
+
+def safe_create_task(coro, label: str = "background task"):
     async def wrapper():
         try:
             await coro
         except Exception as e:
             logger.exception(f"{label} failed with {type(e).__name__}: {e}")
 
-    return asyncio.create_task(wrapper())
+    task = asyncio.create_task(wrapper())
+
+    # Add task to the set to maintain strong reference
+    _background_tasks.add(task)
+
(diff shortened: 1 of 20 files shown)


letta/utils.py
```diff
@@ -1093,14 +1093,38 @@ def make_key(*args, **kwargs):
...
+    # Add task to the set to maintain strong reference
+    _background_tasks.add(task)
+
+    # Remove task from set when done to prevent memory leaks
+    task.add_done_callback(_background_tasks.discard)
+
+    return task
+
+
+def get_background_task_count() -> int:
```
mattzh72, line 1118: are these used?

carenthomas: lol yes Claude was overzealous

mattzh72: can we add this to a trace / expose as API so we can use to debug on cloud (without affecting sdk)

letta/utils.py
```diff
@@ -1289,6 +1321,12 @@ def fire_and_forget(coro, task_name: Optional[str] = None, error_callback: Optio
 
     task = asyncio.create_task(coro)
 
+    # Add task to the set to maintain strong reference
+    _background_tasks.add(task)
+
+    # Remove task from set when done to prevent memory leaks
+    task.add_done_callback(_background_tasks.discard)
```
mattzh72, line 1328: jw can a task have multiple done callbacks? because some tasks might have callbacks already

carenthomas: yeah there can be multiple it seems

letta/utils.py
carenthomas: yep added

PR #2829, opened 2025-09-11
feat: remove user scoping from jobs

carenthomas:


carenthomas:
bad Claude. abandoning in favor of two part solution: 
- https://github.com/letta-ai/letta/pull/2831
- https://github.com/letta-ai/letta/pull/2832

PR #2878, opened 2025-09-14
feat: move telemetry endpoint into steps

carenthomas:
## This PR

**Implementation Details:**

Move the telemetry `get_provider_trace_for_step` endpoint into the steps routes since it uses the step ID path parameter.

**Notes:**

Using `trace` instead of `traces` which is technically an anti-pattern, but a step can only have one trace and it sounds better so maybe it's fine?

## This Project

**Background:**

The SDK Stabilization project is a focused technical initiative to finalize our API surface in preparation for the 1.0 release. This effort addresses inconsistencies, edge cases, and developer pain points to deliver a robust and reliable SDK.

Key objectives include:
1. Standardizing endpoint patterns and parameter structures for consistency
2. Resolving edge cases and unexpected behaviors in the current implementation
3. Strengthening error handling with predictable error codes and useful messages
4. Finalizing type definitions to ensure accurate developer tooling support
5. Documenting clear compatibility guarantees and deprecation policies

This work represents the critical transition from beta to production-ready status, establishing a solid foundation that developers can depend on with confidence. The 1.0 release will mark our commitment to API stability while creating a clean baseline for future enhancements.


carenthomas:
Note - re-adding back the deleted route so we maintain backwards compatibility: https://github.com/letta-ai/letta/pull/2879

PR #2925, opened 2025-09-15
docs: Add posthog analytics

cpfiffer:
Adds PostHog analytics configuration to Fern documentation:

- Configures PostHog analytics in fern/docs.yml with API key and endpoint
- Updates GitHub Actions workflows to include POSTHOG_API_KEY environment variable
- Enables tracking of documentation usage and user behavior

This is a clean re-creation after the main branch force push, containing only the analytics configuration changes.

### DIFF
--- a/.github/workflows/fern-docs-preview.yml
+++ b/.github/workflows/fern-docs-preview.yml
@@ -24,6 +24,7 @@ jobs:
         working-directory: fern
         env:
           FERN_TOKEN: ${{ secrets.FERN_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/fern-docs-preview.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         working-directory: fern
         env:
           FERN_TOKEN: ${{ secrets.FERN_TOKEN }}
+          POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_KEY }}
```
kianjones9, line 27: do we the analytics in the dev unpublished docs? might just decrease our SNR

PR #2935, opened 2025-09-16
feat: add `stop_reason` to runs

sarahwooders:


carenthomas:
looks like there are some test failures but lgtm

Issue #3057, opened 2025-11-01
OpenAI Tool Return Serialisation


hebenon:
**Describe the bug**
When a tool executes client-side (e.g., via letta-code) the resulting MessageRole.tool object persisted by the server contains populated tool_returns but an empty content list. The OpenAI Responses serializer in Message.to_openai_responses_dicts still assumes a single TextContent entry and asserts, causing every subsequent tool step to crash before the agent can continue.

**Please describe your setup**
- [X] How are you running Letta?
  - Docker
- [X] Describe your setup
  - What's your OS (Windows/MacOS/Linux)? Linux (Ubuntu 22.04)
  - What is your `docker run ...` command (if applicable)
    podman run -d -p 8283:8283 -v ~/.letta/.persist/pgdata:/var/lib/postgresql/data -e OPENAI_API_KEY=XXXXXXX

**Screenshots**

**Additional context**
- What model you are using: openai/gpt-5
- Example stack trace:
[...]

PR #3058, opened 2025-11-01
fix: OpenAI Tool Return Serialisation

hebenon:
This PR is a potential fix for #3057. When a tool executes client-side (e.g., via letta-code) the resulting MessageRole.tool object persisted by the server contains populated tool_returns but an empty content list. The OpenAI Responses serializer in Message.to_openai_responses_dicts still assumes a single TextContent entry and asserts, causing every subsequent tool step to crash before the agent can continue.

Changes in the fix include:
- Reworked Message.to_openai_dict and Message.to_openai_responses_dicts so tool messages prefer the tool_returns payload (mirroring the Anthropic serializer) with fallback to legacy single-TextContent handling.
- Updated create_parallel_tool_messages_from_llm_response to inject a TextContent placeholder when only tool_returns are produced, preserving older consumers.
- Added test_message_serialization.py to validate both behaviors (tool-return-first and legacy fallback).

**How to test**
- Configure an agent to use an OpenAI model, with upserted tools from letta-cli.
- Give the agent a request that relies on local tools to complete.

**Have you tested this PR?**
Post-fix, tools can run in letta-cli with OpenAI agents. E.g.:
image

**Related issues or PRs**
This PR addresses #3057. 

No.

**Additional context**
For transparency, this fix was vibecoded with GitHub Copilot.

### DIFF
--- a/letta/schemas/message.py
+++ b/letta/schemas/message.py
@@ -1181,11 +1181,24 @@ class Message(BaseMessage):
                         tool_call_dict["id"] = tool_call_dict["id"][:max_tool_id_length]
 
         elif self.role == "tool":
-            assert self.tool_call_id is not None, vars(self)
+            tool_call_id = self.tool_call_id
+            if text_content is None and self.tool_returns:
+                aggregated_outputs = [tr.func_response for tr in self.tool_returns if tr.func_response]
+                if aggregated_outputs:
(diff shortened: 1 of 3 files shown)


hebenon:
This PR is no longer required, as the underlying issue has been fixed with a different commit set.

PR #3070, opened 2025-11-19
fix: prevent false positive in Secret.get_plaintext() for plaintext values

### LINKED ISSUE #3069: Fix OpenAI API Key Configuration in .env File
### Problem
When setting up the Letta server via the installer, the `.env` file may be created with quoted values for API keys. This causes the server to fail when attempting to use OpenAI API keys, resulting in 500 errors when creating agents.

### Root Causes

1. **Quoted Environment Variables**
   - Docker's `--env-file` flag includes quotes as part of the environment variable value
   - If `.env` contains: `OPENAI_API_KEY="sk-proj-..."`
   - Docker sets: `OPENAI_API_KEY="sk-proj-..."` (with quotes in the value)
   - This causes OpenAI API to reject requests with "Invalid authorization header"

2. **Missing LETTA_ENCRYPTION_KEY**
   - Letta's Secret class requires `LETTA_ENCRYPTION_KEY` to be set, even if encryption is not actively used
   - Without it, the Secret class may incorrectly identify API keys as encrypted and fail to decrypt them
   - This results in `None` being returned instead of the actual API key

### Required Fixes

1. **Ensure .env file format does not include quotes around values:**
   ```bash
   # ✅ Correct
   OPENAI_API_KEY=sk-proj-abc123...
   
   # ❌ Incorrect
   OPENAI_API_KEY="sk-proj-abc123..."
   ```

2. **Ensure LETTA_ENCRYPTION_KEY is set in .env:**
   ```bash
   LETTA_ENCRYPTION_KEY=dummy-encryption-key-32-chars-long!!
   ```
   Note: This is required even if encryption is not being used.

### Impact
- **Severity:** High
- **User Impact:** Server cannot authenticate with OpenAI API, agent creation fails
- **Workaround:** Manual editing of .env file after installation

### References
See detailed incident report for full technical details and investigation.

actuallyrizzn:
## Problem

When a Secret is created from plaintext (via `from_plaintext()` with `was_encrypted=False`), the `is_encrypted()` heuristic can incorrectly identify long API keys (e.g., OpenAI API keys) as encrypted. This causes `get_plaintext()` to return `None` when no encryption key is available, even though the value was explicitly stored as plaintext.

## Root Cause

The `CryptoUtils.is_encrypted()` method uses a heuristic that checks if a value is base64-decodable and meets minimum encrypted size requirements. Long API keys (164+ characters) can trigger false positives, causing the code to attempt decryption and fail.

## Solution

Check the `was_encrypted` flag before trusting the `is_encrypted()` heuristic. If `was_encrypted=False`, we know the value was explicitly created as plaintext, so we should trust the cached value.

## Changes

- Added early return in `get_plaintext()` when `was_encrypted=False` and cache exists
- This prevents false positives from clearing the cache unnecessarily
- Preserves existing behavior for actually encrypted values

## Testing

This fix resolves the issue where OpenAI API keys stored as plaintext would return `None` from `get_plaintext()` when `LETTA_ENCRYPTION_KEY` was not set.

Fixes #3069

### DIFF
--- a/tests/test_secret.py
+++ b/tests/test_secret.py
@@ -278,3 +278,104 @@ class TestSecret:
                 assert mock_decrypt.call_count == 1
         finally:
             settings.encryption_key = original_key
+
+    def test_get_plaintext_with_false_positive_encryption_check(self):
+        """Test that get_plaintext() works correctly when is_encrypted() returns false positive.
+        
+        This test covers the fix for issue #3069 where long API keys (e.g., OpenAI keys)
+        can trigger false positives in the is_encrypted() heuristic, causing get_plaintext()
+        to return None even when the value was explicitly created as plaintext.
+        """
+        from letta.settings import settings
+
+        # Clear encryption key to simulate the problematic scenario
+        original_key = settings.encryption_key
+        settings.encryption_key = None
+
+        try:
+            # Use a long API key that would trigger false positive in is_encrypted()
+            # OpenAI API keys are typically 164+ characters and base64-decodable
+            long_api_key = "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"
+            
+            # Verify this would trigger false positive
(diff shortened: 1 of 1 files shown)


actuallyrizzn:
Hi! I see there are some failing CI checks. The Alembic migration issue looks like it's from the branch being out of sync with main, and there's a linting issue in 	ools.py (unrelated to my changes). Happy to fix these if you'd like, or I can leave them for you to handle. The actual code changes in this PR are just the Secret fix and tests.

PR #3083, opened 2025-11-27
Updated readme with actual argument

Godofnothing:
This PR fixes `README.md` with the correct argument `api_key` instead of legacy `token` in the Python CLI demo.


cpacker (approved):
Thank you! Sorry I think the entire README actually needs to be patched for the V1 SDK release, but this is one of the more obvious / worse errors so thanks for the patch

PR #3097, opened 2025-12-17
Correct provider name for openai-proxy in LLMConfig

SootyOwl:
This pull request includes a targeted fix to the LLM configuration logic in `letta_agent_v3.py`. The change addresses a compatibility issue with Pydantic validation by normalizing the provider name when using the `openai-proxy` provider.

LLM configuration fix:

* Updated the `_build_summarizer_llm_config` function to convert the provider name from `openai-proxy` to `openai` when parsing the model string, ensuring compatibility with Pydantic LLMConfig validation.

Fixes summarization for OpenAI compatible endpoints.


gergesh:
This is needed for using letta code with a self hosted letta server and openai-compatible endpoints. Thanks for the PR and a release would be amazing!

PR #3173, opened 2026-02-01
feat(azure): support reasoning models via Responses API routing

### LINKED ISSUE #3169: Azure client fails with GPT-5/reasoning models - Responses API not supported
## Bug Description

Azure client throws `TypeError: Missing required arguments; Expected either ('messages' and 'model')` when using GPT-5.x or other reasoning models (o1, o3, o4).

## Root Cause

1. `use_responses_api()` returns `True` for models starting with `gpt-5`, `o1`, `o3`, `o4`
2. `build_request_data()` creates Responses API format (with `input` field instead of `messages`)
3. Azure client always calls `chat.completions.create()` which expects `messages` field
4. OpenAI client has routing logic for Responses API, Azure client does not

### Option 1: Minimal fix - Add Responses API routing

Add routing in `request_async` to call `responses.create()` when appropriate:

```python
# Azure Responses API requires standard OpenAI client with /openai/v1/ URL
if use_responses_api(llm_config):
    client = AsyncOpenAI(api_key=api_key, base_url=f"{base_url}/openai/v1/")
    resp = await client.responses.create(**request_data)
else:
    client = AsyncAzureOpenAI(api_key=api_key, azure_endpoint=base_url, api_version=api_version)
    response = await client.chat.completions.create(**request_data)
```

**Caveat**: Requires two different client types due to Azure URL structure differences.

### Option 2: Unified client using /openai/v1/ for both APIs

Azure's `/openai/v1/` endpoint supports both Chat Completions and Responses API:

```python
default_query = {"api-version": api_version} if api_version else None
client = AsyncOpenAI(api_key=api_key, base_url=f"{base_url}/openai/v1/", default_query=default_query)

if use_responses_api(llm_config):
    resp = await client.responses.create(**request_data)
else:
    response = await client.chat.completions.create(**request_data)
```

**Caveat**: Needs testing to confirm Chat Completions works via `/openai/v1/` path for all deployments.

### Option 3: Use user-provided base URL directly (Recommended)

Instead of hardcoding `/openai/v1/`, rely on the user to configure the correct base URL:

```python
# User configures base_url as: https://resource.openai.azure.com/openai/v1/
default_query = {"api-version": api_version} if api_version else None
client = AsyncOpenAI(api_key=api_key, base_url=base_url, default_query=default_query)

if use_responses_api(llm_config):
    resp = await client.responses.create(**request_data)
else:
    response = await client.chat.completions.create(**request_data)
```

**Caveat**: Breaking change - requires docs update and users to change `AZURE_BASE_URL` to include `/openai/v1/` suffix.

## References

- [Azure Responses API docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/responses) - shows `/openai/v1/` URL requirement
- [OpenAI Python SDK azure.py](https://github.com/openai/openai-python/blob/main/src/openai/lib/azure.py) - `_deployments_endpoints` doesn't include `/responses`

## Environment

- Letta Docker (latest)
- Azure OpenAI with GPT-5.2 deployment
- Python openai SDK

yttrium400:
Fixes #3169. This PR implements dynamic routing for Azure OpenAI 'Responses API' requests (used by reasoning models like o1/o3). When `input` is present in the request payload,  now utilizes the standard `OpenAI` (or `AsyncOpenAI`) client with a modified base URL (`/openai/v1`) as required by Azure.

### DIFF
--- a/letta/llm_api/azure_client.py
+++ b/letta/llm_api/azure_client.py
@@ -1,7 +1,7 @@
 import os
 from typing import List, Optional, Tuple
 
-from openai import AsyncAzureOpenAI, AzureOpenAI
+from openai import AsyncAzureOpenAI, AsyncOpenAI, AzureOpenAI, OpenAI
 from openai.types.chat.chat_completion import ChatCompletion
 
 from letta.llm_api.openai_client import OpenAIClient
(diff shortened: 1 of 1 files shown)


themavik (commented):
request() now assigns api_version from env twice back-to-back in the BYOK branch (duplicate line) — looks accidental and should be one assignment. If a payload could ever include both input and messages you would take the chat path; fine if impossible, otherwise worth an explicit guard or comment.

PR #3183, opened 2026-02-13
fix: normalize_mcp_schema adds missing 'properties' for object schemas

### LINKED ISSUE #3145: normalize_mcp_schema should add missing 'properties' field for object schemas
## Description

When importing MCP tools, schemas with `{"type": "object"}` but no `properties` field fail validation. The `normalize_mcp_schema` function should add `properties: {}` when missing.

## Current Behavior

MCP servers may return tool schemas like:
```json
{"type": "object"}
```

This causes the error:
```
Failed to auto-sync tools from MCP server mcpjungle-remote: {'type': 'object', 'additionalProperties': False}. 
Server was created successfully but tools were not persisted.
```

## Expected Behavior

`normalize_mcp_schema` should normalize this to:
```json
{"type": "object", "properties": {}, "additionalProperties": false}
```

## Root Cause

The issue originates from a bug in [mcp-go library](https://github.com/mark3labs/mcp-go/issues/694) where empty `properties` gets dropped during JSON serialization. However, Letta's normalization could easily handle this case.

## Suggested Fix

In `letta/functions/schema_generator.py`, add to `normalize_mcp_schema`:

```python
def normalize_mcp_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
    # Add empty properties if missing for object types
    if schema.get("type") == "object" and "properties" not in schema:
        schema["properties"] = {}
    
    # ... existing normalization logic
```

## Impact

This affects any MCP server that returns tools with no parameters, causing them to fail import even though they're valid tools.

avantikau:
## Summary
- `normalize_mcp_schema` now adds `"properties": {}` to any object schema that is missing it, fixing import failures for zero-argument MCP tools
- Moved the `properties` assertion in `generate_tool_schema_for_mcp` to after normalization so the fix takes effect before validation
- Handles both top-level and nested object schemas via the recursive `normalize_object_schema` helper

## Problem
MCP servers (particularly those using the [mcp-go library](https://github.com/mark3labs/mcp-go/issues/694)) may return tool schemas like `{"type": "object"}` without a `properties` field. This caused validation failures:
```
Failed to auto-sync tools from MCP server: {'type': 'object', 'additionalProperties': False}.
Server was created successfully but tools were not persisted.
```

## Root Cause
The mcp-go library drops empty `properties` maps during JSON serialization, producing `{"type": "object"}` instead of `{"type": "object", "properties": {}}`. Letta's `normalize_mcp_schema` already handled adding `additionalProperties` but did not add the missing `properties` field.

## Changes
**`letta/functions/schema_generator.py`**:
1. In `normalize_object_schema`: Added `properties: {}` when missing for object types (before the existing `additionalProperties` fix)
2. In `generate_tool_schema_for_mcp`: Moved `assert "properties" in parameters_schema` to after `normalize_mcp_schema()` is called, so normalization can fix the schema before validation

## Test plan
- [ ] Verify that a schema like `{"type": "object"}` is normalized to `{"type": "object", "properties": {}, "additionalProperties": false}`
- [ ] Verify that schemas that already have `properties` are unaffected
- [ ] Verify that nested object schemas without `properties` are also fixed
- [ ] Run existing MCP schema tests to ensure no regressions

Fixes #3145

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/letta/functions/schema_generator.py
+++ b/letta/functions/schema_generator.py
@@ -586,9 +586,10 @@ def generate_schema_from_args_schema_v2(
 def normalize_mcp_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
     """
     Normalize an MCP JSON schema to fix common issues:
-    1. Add explicit 'additionalProperties': false to all object types
-    2. Add explicit 'type' field to properties using $ref
-    3. Process $defs recursively
+    1. Add empty 'properties': {} to object types missing it
+    2. Add explicit 'additionalProperties': false to all object types
+    3. Add explicit 'type' field to properties using $ref
+    4. Process $defs recursively
 
     Args:
         schema: The JSON schema to normalize (will be modified in-place)
@@ -604,8 +605,10 @@ def normalize_mcp_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
     def normalize_object_schema(obj_schema: Dict[str, Any], defs: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
         """Recursively normalize an object schema."""
 
-        # If this is an object type, add additionalProperties if missing
+        # If this is an object type, add additionalProperties and properties if missing
         if obj_schema.get("type") == "object":
+            if "properties" not in obj_schema:
+                obj_schema["properties"] = {}
             if "additionalProperties" not in obj_schema:
                 obj_schema["additionalProperties"] = False
 
@@ -705,12 +708,14 @@ def generate_tool_schema_for_mcp(
     description = mcp_tool.description
 
     assert "type" in parameters_schema, parameters_schema
-    assert "properties" in parameters_schema, parameters_schema
     # assert "required" in parameters_schema, parameters_schema
 
     # Normalize the schema to fix common issues with MCP schemas
-    # This adds additionalProperties: false and explicit types for $ref properties
+    # This adds additionalProperties: false, explicit types for $ref properties,
+    # and ensures object schemas have a 'properties' field
     parameters_schema = normalize_mcp_schema(parameters_schema)
+
+    assert "properties" in parameters_schema, parameters_schema
 
     # Zero-arg tools often omit "required" because nothing is required.
     # Normalise so downstream code can treat it consistently.


biswajeetdev (commented):
Clean fix. MCP tools with no parameters emit schemas like `{"type": "object"}` with no `properties` key, which was hitting the assert before normalization had a chance to repair it. Moving the assert to after `normalize_mcp_schema` and having normalization inject `"properties": {}` is the right layering.

One edge case to consider: some MCP schemas use `"type": "object"` with only `$defs` and `$ref` entries (no direct `properties`), relying on JSON Schema `$ref` resolution to define the shape. After this change those schemas also get `"properties": {}` injected. Depending on how downstream code consumes the schema, an empty `"properties"` alongside `$ref` could cause a validator to require properties to be empty while the `$ref` says otherwise. Worth checking whether the `$ref`-flattening logic that runs after normalization handles this case, or whether the `"properties": {}` injection should be skipped when `$ref` or `$defs` keys are present.

ErenAta16 (commented):
The thing a reviewer will worry about here is whether injecting `"properties": {}` alongside the existing `"additionalProperties": false` tightens any schema, since together they read as "no keys permitted". It does not. Validation is unchanged:

```
document     {"type":"object","additionalProperties":false}   ...same + "properties":{}
{}           accept                                           accept
{"x": 1}     reject                                           reject
```

`additionalProperties` governs every property not matched by `properties` or `patternProperties`, so with `properties` absent all keys are already additional and `false` already forbids them. Adding the empty object is a no-op for any validator. Worth putting that in the description, because "add `properties: {}` to every object schema" looks like a semantic change and reviewing it as one wastes time.

So this is purely about shape: giving downstream consumers, and the assertion below, a key they can rely on existing. That is the right fix for zero-argument MCP tools, where a server legitimately sends an object schema with nothing in it.

**Moving the assertion after `normalize_mcp_schema` is the actual fix**, and I would lead with that rather than with the normalization:

```python
-    assert "properties" in parameters_schema, parameters_schema
-    # This adds additionalProperties: false and explicit types for $ref properties
+    # ... and ensures object schemas have a 'properties' field
+    assert "properties" in parameters_schema, parameters_schema
```

Before, a zero-argument tool failed at import on an assertion, before the normalizer that would have supplied the missing key ever ran. The normalizer change alone would not have helped while the assert stayed above it.

One thing worth deciding: an `assert` here means the check disappears under `python -O`. If this is a guard against malformed server input rather than an internal invariant, and MCP schemas come from third-party servers, it wants to be a raised exception with a message naming the tool. If it is genuinely an internal invariant that the normalizer now guarantees, the assert is fine and the docstring change already documents why.

Numbering the docstring steps 1-4 with the new step first matches the execution order, which is a small thing that keeps the comment honest as more normalization gets added.

PR #3204, opened 2026-02-25
fix: preserve HTTPS scheme in trailing-slash redirects

### LINKED ISSUE #3189: Trailing-slash redirects use HTTP scheme behind HTTPS proxies, breaking Authorization forwarding
## Bug Description

When Letta is deployed behind an HTTPS-terminating reverse proxy (Cloudflare Tunnels, nginx, Traefik, etc.), API endpoints that trigger trailing-slash redirects generate `Location: http://...` URLs instead of `Location: https://...`. This causes HTTP clients that follow redirects (including Node.js `fetch()`) to strip the `Authorization` header per the [Fetch specification](https://fetch.spec.whatwg.org/#http-redirect-fetch), resulting in `401 Unauthorized` errors.

## Root Cause

In [`letta/server/rest_api/app.py`](https://github.com/letta-ai/letta/blob/main/letta/server/rest_api/app.py), `uvicorn.run()` is called without `proxy_headers=True`, so Starlette doesn't respect `X-Forwarded-Proto` when generating redirect URLs. There is even a commented-out middleware for this:

```python
# Add reverse proxy middleware to handle X-Forwarded-* headers
# app.add_middleware(ReverseProxyMiddleware, base_path=settings.server_base_path)
```

## Reproduction

**Setup:** Letta v0.16.4 behind Cloudflare Tunnel (TLS terminated at edge, forwarded as HTTP to Letta)

```bash
# This triggers a 307 redirect with http:// Location header
$ curl -v -X POST "https://letta.example.com/v1/conversations?agent_id=agent-xxx" \
  -H "Authorization: Bearer $LETTA_API_KEY"

# Response:
# < HTTP/2 307
# < location: http://letta.example.com/v1/conversations/?agent_id=agent-xxx
#              ^^^^
#              Should be https://

# Adding trailing slash manually bypasses the redirect and works:
$ curl -X POST "https://letta.example.com/v1/conversations/?agent_id=agent-xxx" \
  -H "Authorization: Bearer $LETTA_API_KEY"
# => 200 OK
```

## Impact

Per the Fetch spec, when a request is redirected from HTTPS to HTTP (a security downgrade), the `Authorization` header is stripped. This means:

- **Node.js `fetch()`** and **browser `fetch()`** silently drop auth on redirect → 401
- **Any HTTPS-terminated proxy deployment** is affected (Cloudflare Tunnels, nginx, AWS ALB, GCP LB, etc.)
- Confirmed affected: `POST /v1/conversations` — any endpoint that adds a trailing-slash redirect will have the same issue
- This specifically breaks the [claude-subconscious](https://github.com/letta-ai/claude-subconscious) plugin's `SessionStart` hook, which uses Node.js `fetch()` to create conversations

## Suggested Fix

Add proxy header support to the `uvicorn.run()` call in `app.py`:

```python
uvicorn.run(
    "letta.server.rest_api.app:app",
    host=host or "localhost",
    port=port or REST_DEFAULT_PORT,
    workers=settings.uvicorn_workers,
    reload=reload or settings.uvicorn_reload,
    timeout_keep_alive=settings.uvicorn_timeout_keep_alive,
    access_log=False,
    proxy_headers=True,
    forwarded_allow_ips="*",
)
```

This makes uvicorn respect `X-Forwarded-Proto: https` from reverse proxies when generating redirect Location headers.

## Environment

- Letta version: 0.16.4
- Deployment: Docker behind Cloudflare Tunnel
- Client: Node.js 22+ (native `fetch()`)

Nicknamess96:
## Summary

Enable uvicorn's proxy header support so that `X-Forwarded-Proto` from reverse proxies is respected when constructing trailing-slash redirect `Location` headers, preventing unintended HTTPS-to-HTTP protocol downgrades.

## Problem

When Letta is deployed behind an HTTPS-terminating reverse proxy (Cloudflare Tunnels, nginx, Traefik, AWS ALB, etc.), endpoints that trigger trailing-slash redirects generate `Location: http://...` URLs instead of `Location: https://...`. Per the [Fetch specification](https://fetch.spec.whatwg.org/#http-redirect-fetch), HTTP clients strip `Authorization` headers on HTTPS-to-HTTP redirects, resulting in `401 Unauthorized` errors.

Root cause: `uvicorn.run()` in `start_server()` is called without `proxy_headers=True`, so Starlette doesn't read `X-Forwarded-Proto` when building redirect URLs.

## Changes

- Add `proxy_headers=True` and `forwarded_allow_ips="*"` to both `uvicorn.run()` call sites (HTTP and `LOCAL_HTTPS` modes) in `letta/server/rest_api/app.py`
- Add tests verifying redirect scheme preservation with uvicorn's `ProxyHeadersMiddleware`
- Add tests verifying `uvicorn.run()` receives the correct proxy parameters

## Testing

- 7 new tests in `tests/test_proxy_headers.py`, all passing
- Formatted with `ruff format` (line length 140)
- Linted with `ruff check` (all checks passed)

Fixes #3189

### DIFF
--- a/letta/server/rest_api/app.py
+++ b/letta/server/rest_api/app.py
@@ -926,6 +926,8 @@ def start_server(
                 ssl_keyfile="certs/localhost-key.pem",
                 ssl_certfile="certs/localhost.pem",
                 access_log=False,
+                proxy_headers=True,
+                forwarded_allow_ips="*",
             )
 
     else:
@@ -965,4 +967,6 @@ def start_server(
                 reload=reload or settings.uvicorn_reload,
                 timeout_keep_alive=settings.uvicorn_timeout_keep_alive,
                 access_log=False,
+                proxy_headers=True,
+                forwarded_allow_ips="*",
             )
--- a/tests/test_proxy_headers.py
+++ b/tests/test_proxy_headers.py
@@ -1,0 +1,170 @@
+"""Tests for HTTPS proxy header handling in trailing-slash redirects.
+
+Verifies that when Letta is deployed behind an HTTPS-terminating reverse proxy
+(e.g., Cloudflare Tunnels, nginx, Traefik), trailing-slash redirects preserve
+the original HTTPS scheme from the X-Forwarded-Proto header instead of
+downgrading to HTTP.
+
+See: https://github.com/letta-ai/letta/issues/3189
+"""
+
+from unittest.mock import patch
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
+
+
+def _build_app_with_proxy_middleware() -> FastAPI:
+    """Build a minimal FastAPI app with uvicorn's ProxyHeadersMiddleware.
+
+    In production, uvicorn injects ``ProxyHeadersMiddleware`` when
+    ``proxy_headers=True`` is passed to ``uvicorn.run()``.  We replicate
+    that here so the test exercises the same ASGI middleware chain that
+    runs in a real deployment behind a reverse proxy.
+    """
+    app = FastAPI(redirect_slashes=True)
+
+    @app.get("/v1/health/")
+    async def health():
+        return {"status": "ok"}
+
+    @app.post("/v1/conversations/")
+    async def create_conversation():
+        return {"id": "conv-123"}
+
+    return app
+
+
+def _wrap_with_proxy_middleware(app: FastAPI) -> ProxyHeadersMiddleware:
+    """Wrap a FastAPI app with uvicorn's ProxyHeadersMiddleware.
+
+    This simulates the middleware that uvicorn adds when
+    ``proxy_headers=True`` and ``forwarded_allow_ips="*"`` are passed
+    to ``uvicorn.run()``.
+    """
+    return ProxyHeadersMiddleware(app, trusted_hosts="*")
+
+
+@pytest.fixture
+def app():
+    return _build_app_with_proxy_middleware()
+
+
+@pytest.fixture
+def proxy_client(app):
+    """Client with ProxyHeadersMiddleware (simulates uvicorn proxy_headers=True)."""
+    wrapped = _wrap_with_proxy_middleware(app)
+    return TestClient(wrapped, raise_server_exceptions=False)
+
+
+@pytest.fixture
+def plain_client(app):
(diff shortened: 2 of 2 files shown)


Nicknamess96:
Friendly ping — this is a one-line fix to preserve the HTTPS scheme when the trailing-slash redirect middleware rewrites URLs. Happy to address any feedback.

alvinttang (commented):
## Code Review

### Security: `forwarded_allow_ips="*"` trusts all proxy headers unconditionally

**File: `letta/server/rest_api/app.py`, both `uvicorn.run()` call sites**

```python
proxy_headers=True,
forwarded_allow_ips="*",
```

Setting `forwarded_allow_ips="*"` means **any** client can set `X-Forwarded-Proto`, `X-Forwarded-For`, etc., and uvicorn will trust them. This is a well-known security anti-pattern:

1. **IP spoofing**: An attacker can set `X-Forwarded-For: 127.0.0.1` to bypass IP-based access controls or rate limiting that relies on the client IP from the ASGI scope.
2. **Scheme spoofing**: An attacker can set `X-Forwarded-Proto: https` to trick the application into thinking it's behind HTTPS when it isn't, potentially bypassing secure-cookie checks.

For deployments behind a known reverse proxy, `forwarded_allow_ips` should be set to the proxy's IP address(es) or subnet, not `"*"`. If the deployment topology varies, this should at minimum be configurable via an environment variable with a secure default:

```python
forwarded_allow_ips=os.environ.get("LETTA_FORWARDED_ALLOW_IPS", "127.0.0.1"),
```

This way the default is secure (only trust localhost as a proxy), and operators behind Cloudflare/nginx can explicitly opt in to broader trust.

### Minor: Tests don't test the actual `app.py` middleware chain

The tests in `test_proxy_headers.py` build a separate minimal FastAPI app and manually wrap it with `ProxyHeadersMiddleware`. This validates that uvicorn's middleware works correctly (which is uvicorn's responsibility), but doesn't test that the actual Letta application's middleware stack, route definitions, and startup sequence correctly integrate the proxy headers configuration. Consider adding an integration test that imports the real `create_app()` and verifies the middleware is present.

themavik (commented):
Enabling forwarded_allow_ips="*" means any client that can reach the socket can forge X-Forwarded-Proto/Host; fine if Letta only listens on loopback or behind a locked-down proxy hop, but worth calling out in docs or tightening to known proxy CIDRs for public binds.

